PR #2, opened 2022-05-19
setup.cfg: add protobuf dependency and update package_dir

stefanst3000:
Add protobuf to required dependencies, since is needed when using the
android-emulator transport. Also, update package_dir with the explicit
path to bumble/transport.

Signed-off-by: Stefan Stanacar <stefanst@google.com>


setup.cfg
```diff
@@ -27,6 +27,7 @@ package_dir =
     bumble = bumble
     bumble.apps = apps
     bumble.apps.link_relay = apps/link_relay
+    bumble.transport = bumble/transport
```
barbibulle, line 30: I don't think this is needed, because `transport` is already a subdir of `bumble`. I realize that this also means that line 29 above is also likely not needed (but line 28 is needed, because `apps` isn't a subdir)

barbibulle (approved):

setup.cfg
stefanst3000: Indeed is not needed, I'll remove it (in the internal version line 26 was missing `bumble.transport` in `package`s directive and I just carried the explicit package_dir over).

stefanst3000: Removed.

barbibulle (approved):

PR #3, opened 2022-06-02
gbg/usb serial number

barbibulle:
Add support for specifying a serial number for USB devices.
Also includes a small refactoring of the documentation.

### DIFF
--- a/bumble/transport/usb.py
+++ b/bumble/transport/usb.py
@@ -37,14 +37,17 @@ async def open_usb_transport(spec):
     '''
     Open a USB transport.
     The parameter string has this syntax:
-    either <index> or <vendor>:<product>
+    either <index> or <vendor>:<product>[/<serial-number>]
     With <index> as the 0-based index to select amongst all the devices that appear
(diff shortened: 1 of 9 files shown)


bumble/transport/usb.py
```diff
@@ -268,22 +271,32 @@ async def close(self):
...
+                product_id, serial_number = product_id.split('/')
+                for device in context.getDeviceIterator(skip_on_error=True):
+                    if (
+                        device.getVendorID() == int(vendor_id, 16) and
+                        device.getProductID() == int(product_id, 16) and
+                        device.getSerialNumber() == serial_number
+                    ):
+                        found = device
+                        break
+                device.close()
```
mogenson, line 284: You want to close the `device` from the iterator right? I think this line should be indented one more level to be inside the `for` loop.

barbibulle: yes, indeed.

barbibulle: now fixed in latest commit

mogenson (approved):

PR #4, opened 2022-06-11
classic pairing io

barbibulle:
Extend the notion of `PairingConfig` and `PairingDelegate` to Classic pairing, so that clients can specify their own non-default options and perform the actual I/O operations, just like for LE pairing.
The `pair.py` tool is also enhanced to support Classic pairing.
NOTE: there are no self-tests for this, because the in-process virtual controller implementation doesn't yet support Classic (this feature only works with non-builtin controllers)


bumble/device.py
```diff
@@ -1141,17 +1157,121 @@ async def on_inquiry_complete(self):
...
+        authentication_requirements = (
+            (
+                HCI_MITM_NOT_REQUIRED_NO_BONDING_AUTHENTICATION_REQUIREMENTS,
+                HCI_MITM_REQUIRED_NO_BONDING_AUTHENTICATION_REQUIREMENTS
+            ),
+            (
+                HCI_MITM_NOT_REQUIRED_GENERAL_BONDING_AUTHENTICATION_REQUIREMENTS,
+                HCI_MITM_REQUIRED_GENERAL_BONDING_AUTHENTICATION_REQUIREMENTS
+            )
+        )[1 if pairing_config.bonding else 0][1 if pairing_config.mitm else 0]
```
DeltaEvo, line 1199: I don't like a lot `[1 if pairing_config.bonding else 0][1 if pairing_config.mitm else 0]`

```python
        # Compute the authentication requirements
        authentication_requirements = (
            (
                HCI_MITM_REQUIRED_GENERAL_BONDING_AUTHENTICATION_REQUIREMENTS
                if pairing_config.mitm
                else HCI_MITM_REQUIRED_GENERAL_BONDING_AUTHENTICATION_REQUIREMENTS
            )
            if pairing_config.bonding
            else (
                HCI_MITM_REQUIRED_NO_BONDING_AUTHENTICATION_REQUIREMENTS
                if pairing_config.mitm
                else HCI_MITM_NOT_REQUIRED_NO_BONDING_AUTHENTICATION_REQUIREMENTS
            )
        )
```
Maybe using the conditional expression is a bit better, but I don't seems to find a way to format it in a way I really like it.

What do you think ?

bumble/device.py
```diff
@@ -1141,17 +1157,121 @@ async def on_inquiry_complete(self):
...
+
+        # Respond
+        self.host.send_command_sync(
+            HCI_IO_Capability_Request_Reply_Command(
+                bd_addr                     = connection.peer_address,
+                io_capability               = io_capability,
+                oob_data_present            = 0x00,
```
DeltaEvo, line 1206: nitpick: Add a comment to precise that `0x00` is `Not Present` ?

bumble/device.py
```diff
@@ -1141,17 +1157,121 @@ async def on_inquiry_complete(self):
...
         connection.emit('connection_authentication_failure', error)
 
+    # [Classic only]
+    @host_event_handler
+    @with_connection_from_address
+    def on_authentication_io_capability_request(self, connection):
+        # Ask what the pairing config should be for this connection
+        pairing_config = self.pairing_config_factory(connection)
+
+        # Map the SMP IO capability to a Classic IO capability
```
8lank, line 1178: Shouldn’t we have an additional abstraction on top of pair/authenticate?

* We have a `pair` method for BLE only (it may be worth adding a comment on it btw) and `authenticate` for classic only, which is a bit confusing from a user perspective.
* The current structure uses SMP global variables within classic pairing methods (for IO caps), which should be fine for most users (since it is abstracted in `PairingDelegate`), but a bit confusing for a Bluetooth informed developer (since SMP is supposed to handle BLE only).
* It may simplify implementing Cross Transport Key Derivation (CTKD) if we are able to share keys as well. 

Ideally, I would like to define IO caps for my device (not for a connection), call `pair` (or `authenticate`) on a specific connection (with BR/EDR or BLE transport), and, if there is another connection with the same peer using another transport, be able to use CTKD (if enabled) to avoid having to redo the pairing process.

apps/pair.py
```diff
@@ -134,7 +139,7 @@ async def display_number(self, number):
         # Display a PIN code
         print(color('###-----------------------------------', 'yellow'))
         print(color(f'### Pairing with {self.peer_name}', 'yellow'))
-        print(color(f'### PIN: {number:06}', 'yellow'))
+        print(color(f'### PIN: {number:0{digits}}', 'yellow'))
```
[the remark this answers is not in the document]
barbibulle: My understanding is that the verification code is usually shown as a 6-digit 0-padded value.
At least, that's what I observe on iOS and Android.

apps/pair.py
```diff
@@ -278,10 +302,14 @@ async def pair(sc, mitm, bond, io, prompt, request, print_keys, keystore_file, d
 
             if not request:
                 try:
-                    await connection.pair()
+                    if mode == 'le':
+                        await connection.pair()
+                    else:
+                        await connection.authenticate()
+                    return
```
8lank, line 309: As I mentioned in another comment, it would be great to hide the difference to the user

bumble/device.py
barbibulle: I had also started with an if/else expression, but reverted to the table form because the formatting was sort of odd. Maybe just adding comments in the table form would make it more obvious. I'll try something in the next commit and we can discuss how that looks.

bumble/device.py
barbibulle: Yes, we should definitely try to unify the API for Classic `authenticate` and BLE `pair`. I was leaving that as a future enhancement, once the underlying transport-specific APIs are well implemented.
The general idea is that most developers are going to be interested in either BLE or Classic, and only a small subset will be interested in both at the same time. The goal is to offer common abstractions when they come at no cost (or close to no cost) to the transport-specific users, so that we don't burden single-transport users with things that aren't relevant to them.
About key sharing, that should already be supported, since the notion of key store is the same regardless of the transport. But dual-connections haven't really been tested yet, and thus CTKD is still on the TODO list.

I also agree that it would be nice to refactor the code a bit in the future to move the PairingDelegate and PairingConfig in a transport-independent place, and update the SMP implementation to use that transport-independent class with mappings back to the BLE-specific values. That can be done in a future PR (the fact that the smp classes are used internally here isn't exposed to the users of the `Device` class, so that sort of refactoring should be fairly easy, just an import statement change.

apps/pair.py
barbibulle: I agree. This started as two separate methods, mostly because initially only BLE pairing was implemented, then basic Classic authentication was added. I propose that we start with the two APIs as they are for now, and in a future PR unify them in a single API.

bumble/device.py
8lank: Sounds good, let's keep that as a future enhancement 👍

apps/pair.py
8lank: Sounds good!

apps/pair.py
```diff
@@ -62,12 +63,16 @@ async def update_peer_name(self):
             # We already asked the peer
             return
 
-        # Try to get the peer's name
-        if self.peer:
-            peer_name = await get_peer_name(self.peer)
-            self.peer_name = f'{peer_name or ""} [{self.peer.connection.peer_address}]'
+        # For classic, just use the address
```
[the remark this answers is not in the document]
barbibulle: Fixed in latest commit

bumble/device.py
```diff
@@ -292,7 +292,18 @@ def wrapper(self, connection_handle, *args, **kwargs):
...
+
+
+# Decorator that converts the first argument from a bluetooth address to a connection
+def with_connection_from_address(function):
+    @functools.wraps(function)
+    def wrapper(self, address, *args, **kwargs):
+        for connection in self.connections.values():
+            if connection.peer_address == address:
+                return function(self, connection, *args, **kwargs)
+        logger.warn(f'no connection found for address {address}')
```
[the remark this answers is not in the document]
barbibulle: Fixed in latest commit

apps/pair.py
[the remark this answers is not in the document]
DeltaEvo: 👍

DeltaEvo (approved):

PR #6, opened 2022-06-21
Added release tasks

rramirezmata:
Added two tasks to tasks.py, release and release_tests.

Applied black formatter

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -24,34 +24,48 @@ ROOT_DIR = os.path.dirname(os.path.realpath(__file__))
 ns = Collection()
 
 build_tasks = Collection()
-ns.add_collection(build_tasks, name='build')
+ns.add_collection(build_tasks, name="build")
 
 
 @task
 def build(ctx):
-    ctx.run('python -m build')
+    ctx.run("python -m build")
 
-build_tasks.add_task(build, default=True, name='build')
 
-test_tasks = Collection()
-ns.add_collection(test_tasks, name='test')
+build_tasks.add_task(build, default=True, name="build")
+
 
 @task
-def test(ctx, filter=None, junit=False):
+def release_build(ctx):
+    ctx.run('python -m pip install .[build]')
+    build(ctx)
+
+
+build_tasks.add_task(release_build, name="release")
+
+
+@task
+def mkdocs(ctx):
+    ctx.run("mkdocs build -f docs/mkdocs/mkdocs.yml")
+
+
+build_tasks.add_task(mkdocs, name="mkdocs")
+
+test_tasks = Collection()
+ns.add_collection(test_tasks, name="test")
+
+
+@task
+def test(ctx, filter=None, junit=False, install=False):
+    # Install the package before running the tests
+    if install:
+        ctx.run("python -m pip install .[test]")
     args = ""
     if junit:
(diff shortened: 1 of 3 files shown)


mogenson:
FYI invoke tasks named "release" and "release_tests" are required for internal packaging / distribution. These tasks shouldn't affect external users at all.

stefanst3000 (approved):

tavip (approved):

tasks.py
```diff
@@ -24,17 +24,28 @@
...
+    ctx.run("python -m build")
 
-build_tasks.add_task(build, default=True, name='build')
+
+build_tasks.add_task(build, default=True, name="build")
+
+
+@task
+def release_build(ctx):
+    ctx.run('pip install "$(grep build setup.cfg | cut -f2 -d-)"')
```
barbibulle, line 40: pip should be invoked as `python -m pip` rather than just `pip`, which is considered safer (even if, on a properly configured host, it should be the same)

tasks.py
```diff
@@ -43,15 +54,24 @@ def test(ctx, filter=None, junit=False):
...
+
+
+@task
+def release_test(ctx):
+    ctx.run("pip install .[test]")
+    test(ctx, junit=True)
+
 
-test_tasks.add_task(test, name='test', default=True)
+test_tasks.add_task(release_test, name="release")
```
barbibulle, line 69: The name `release` for this task is somewhat misleading, as it doesn’t really perform a “release” of the package, it merely installs the package and runs the tests. While we don’t yet have a proper release script for the package, we will soon, and reserving this task name now will be an issue/conflict when we do.

tasks.py
```diff
@@ -43,15 +54,24 @@ def test(ctx, filter=None, junit=False):
...
-            .format(os.path.join(ROOT_DIR, "tests"), args))
+    ctx.run("python -m pytest {} {}".format(os.path.join(ROOT_DIR, "tests"), args))
+
+
+test_tasks.add_task(test, name="test", default=True)
+
+
+@task
+def release_test(ctx):
+    ctx.run("pip install .[test]")
```
barbibulle, line 65: Consider using `python -m pip` (see other comment)

tasks.py
rramirezmata: Do you have a suggestion for the task name?

barbibulle: I would suggest adding a Boolean `install` option to the `test` task instead of adding a new task. That option, when set, would perform the install step before running the tests.

tasks.py
```diff
@@ -37,18 +37,29 @@ def build(ctx):
 
 @task
 def release_build(ctx):
-    ctx.run('pip install "$(grep build setup.cfg | cut -f2 -d-)"')
+    ctx.run('python -m pip install "$(grep build setup.cfg | cut -f2 -d-)"')
```
barbibulle, line 40: `python -m pip install .[development]` would be a cleaner way to install the `build` package (the `build` module is listed as a dependency in the `development` group in `setup.cfg`. Maybe even put `build >= 0.7` in a separate `build` category in `setup.cfg` (instead of `development`) and use `python -m pip install .[build]` to install the build-time modules.

barbibulle (approved):

PR #9, opened 2022-07-14
Predefined services: Device Information Service and Battery Service

Arrowbox:
Added some initial support for pre-defined services and two example services. I'm open to changes this was just what made sense in an initial approach.


Arrowbox:
Renamed `spec` to `profiles` and separated implementations by service rather than keep all services in one file and all characteristics in another.

Arrowbox:
Closing in favor of #13

PR #10, opened 2022-07-15
Make pairing and link mode configurable

zxzxwu:
These are critical pairing modes. SC(Secure Connection) and SSP(Secure Simple Pairing) determine the procedure and algorithem in authentication.
Besides, LE_Host_Support perameter is disabled in controllers by default(Vol. 4, Part E, 7.3.39), so a write command is necessary to make sure LE is enabled or not.


bumble/device.py
```diff
@@ -268,6 +273,10 @@ def load_from_dict(self, config):
         self.advertising_interval_min = config.get('advertising_interval', self.advertising_interval_min)
         self.advertising_interval_max = self.advertising_interval_min
         self.keystore = config.get('keystore')
+        self.le_enabled = config.get('le_enabled', self.le_enabled)
```
barbibulle, line 276: nit: align '=' characters to keep the assignments in a column form

zxzxwu: Done

barbibulle (approved):

PR #12, opened 2022-07-20
more hci types

barbibulle:
This adds a few types and constants that were still missing (should update to the core spec 5.3).
Also included is a small app that shows the controller info (version, supported features, ...)


mogenson (approved):
👍

PR #15, opened 2022-07-25
add support for the heart rate service

barbibulle:
Simple implementation of the standard HR profile, with client and server example apps.
Tested with virtual and physical HR sensors.

### DIFF
--- a/examples/heart_rate_server.py
+++ b/examples/heart_rate_server.py
@@ -1,0 +1,95 @@
+# Copyright 2021-2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


mogenson (approved):

examples/heart_rate_server.py
```diff
@@ -0,0 +1,95 @@
...
+            serial_number     = '7654321',
+            hardware_revision = '1.1.3',
+            software_revision = '2.5.6',
+            system_id         = (0x123456, 0x8877665544)
+        )
+
+        heart_rate_service = HeartRateService(
+            read_heart_rate_measurement = lambda _: HeartRateService.HeartRateMeasurement(
+                heart_rate              = 100 + int(50 * math.sin(time.time() * math.pi / 60)),
+                sensor_contact_detected = random.choice((True, False, None)),
```
Arrowbox, line 64: Not to be too pedantic, but sensor contact support shouldn't change once connected.

Arrowbox: `HeartRateService` could treat `sensor_contact_detected` as indicating support if it is not None.

Arrowbox (approved):

examples/heart_rate_server.py
barbibulle: That's the way the class is supposed to handle `sensor_contact_detected`, unless I made a mistake: The value is either `True`, or `False`, or `None`. When `True` or `False`, the 'supported' bit is set, and the value is included. If `None, the 'supported' bit isn't set and there's no value.

You're right, the supported/not-supported bit shouldn't change once connected. I've put it this way here more as a way to test different combinations easily.

PR #16, opened 2022-07-26
Adding in Device.connected_to context manager and Peer.sustain

Arrowbox:
* Device.connected_to connects to a peer as a context manager.
* The peer is disconnected when leaving the context
* Will perform discovery automatically on connecting, if requested
* Added HCI_StatusError that can be raised when a HCI Command Status
  event is received that doesn't show "PENDING" as status
* Added Connection.sustain to wait for a timeout or disconnect
* Peer.sustain also maps to Connectin.sustain
* Updated battery_client.py to use .connected_to and .sustain

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -18,6 +18,7 @@
 import json
 import asyncio
 import logging
+from  contextlib import asynccontextmanager, AsyncExitStack
 
 from .hci import *
 from .host import Host
@@ -148,9 +149,23 @@ class Peer:
             await service.discover_characteristics()
             return self.create_service_proxy(proxy_class)
 
+    async def sustain(self, timeout=None):
+        await self.connection.sustain(timeout)
+
     # [Classic only]
     async def request_name(self):
         return await self.connection.request_remote_name()
+
+    async def __aenter__(self):
+        await self.discover_services()
+        for service in self.services:
+            await self.discover_characteristics()
+
+        return self
+
+    async def __aexit__(self, exc_type, exc_value, traceback):
+        pass
+
 
     def __str__(self):
         return f'{self.connection.peer_address} as {self.connection.role_name}'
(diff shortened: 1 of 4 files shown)


bumble/device.py
```diff
@@ -868,6 +887,26 @@ async def disconnect(self, connection, reason):
             connection.remove_listener('disconnection_failure', pending_disconnection.set_exception)
             self.disconnecting = False
 
+    @asynccontextmanager
```
barbibulle, line 890: I like the idea of a context manager for connections. Maybe we should have the `Connection` object itself implement the context management protocol, so that you would simply use: `with await device.connect(xyz) as connection`. (similar to how the File object implements the context management protocol, so you can choose to call `open(xyz)` or `with open(xyz) as input`).
This would also work regardless of whether you're intending to use GATT or not (Peer is mostly an abstraction for the GATT client functions, not something that's currently useful for Classic connections, and `auto_discovery` also only applies to GATT, not to all connections).
The `auto_discovery` functionality could then be moved to the `Peer` constructor, something like `peer = Peer(connection, auto_discovery=True)`.

Arrowbox: Adding it to the `Connection` object directly makes a lot of sense. While `Peer` is only applicable to GATT, it is the more common use-case in the examples. I think there is still value in a context manager that can do the initial setup. For GATT clients, returning a `Peer` and performing discovery is a common flow.

Would you be amenable to adding the context manager to `Connection`, add a context manager to `Peer` for performing discovery and adding a `client=None` to the `.connect` method that takes a class that can be instantiated with a connection and is a context manager. So you can do `async with device.connect(addr, client=Peer) as peer` or `async with device.connect(addr) as conn`.

barbibulle: That would work. But we'd lose the `auto_discovery` argument, which is useful. Maybe we can offer the simple `device.connect(addr)` that returns a `Connection` for the generic use case (whether or not the caller then decides to instantiate a `Peer`), plus a convenience `device.connect_for_gatt(addr, auto_discovery=False)` that would return a context-managed `Peer` with optional auto-discovery (like you have now)

barbibulle (approved):

PR #17, opened 2022-07-29
Implement 'write' command for console.py

mogenson:
I noticed that the 'write' command in `console.py` is unimplemented.
Here's an attempt at adding a `do_write()` handler.

Syntax is `write <attribute> <value>`. Supports a value of type string,
hexadecimal string, or integer.

Ex:
- `write 180D.2A38 hello`
- `write 180D.2A38 0xbeef`
- `write 180D.2A38 123`

Write with response method is used if supported by characteristic,
otherwise write without response.

Tested with run_gatt_server.py example to verify sent data.

### DIFF
--- a/apps/console.py
+++ b/apps/console.py
@@ -32,6 +32,7 @@ from bumble.core import UUID, AdvertisingData
 from bumble.device import Device, Connection, Peer
 from bumble.utils import AsyncRunner
 from bumble.transport import open_transport_or_link
+from bumble.gatt import Characteristic
 
 from prompt_toolkit import Application
 from prompt_toolkit.history import FileHistory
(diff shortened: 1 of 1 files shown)


apps/console.py
```diff
@@ -462,6 +463,43 @@ async def do_read(self, params):
...
+                                characteristic,
+                                value,
+                                with_response=(
+                                    characteristic.properties & Characteristic.WRITE
+                                ),  # use write with response, if supported
+                            )
+                            return
+            self.show_error('no such characteristic')
+        else:
+            self.show_error('no such characteristic')
```
barbibulle, line 501: It would be nice to support specifying the characteristic by its handle, as is done for `read`. Both `read_value` and `write_value` should support passing in either an attribute handle or a characteristic. The code to parse the command and retrieve the handle or characteristic could be factored out and shared between the `read` and `write` commands.

mogenson: Sure thing. Added a `find_attribute()` method to look up a characteristic or attribute handle for `do_read()` and `do_write()`.

barbibulle (approved):

PR #20, opened 2022-08-03
add long read self test

barbibulle:
Add a test to check that we can read characteristics of all sizes, regardless of MTU size.
Also includes a small fix for characteristic adapters

### DIFF
--- a/tests/self_test.py
+++ b/tests/self_test.py
@@ -164,6 +164,44 @@ async def test_self_gatt():
 
 
 # -----------------------------------------------------------------------------
+@pytest.mark.asyncio
+async def test_self_gatt_long_read():
+    # Create two devices, each with a controller, attached to the same link
+    two_devices = TwoDevices()
+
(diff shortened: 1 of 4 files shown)


tests/self_test.py
```diff
@@ -163,6 +163,44 @@ async def test_self_gatt():
...
+
+    # Add some GATT characteristics to device 1
+    characteristics = [
+        Characteristic(
+            f'3A143AD7-D4A7-436B-97D6-5B62C315{i:04X}',
+            Characteristic.READ,
+            Characteristic.READABLE,
+            bytes([x & 255 for x in range(i)])
+        )
+        for i in range(0, 513)
```
Arrowbox, line 180: Why 513?

Arrowbox (approved):
LGTM

tests/self_test.py
barbibulle: That's because the spec allows GATT characteristic values up to 512 bytes

PR #22, opened 2022-08-11
Add subscribe and unsubscribe commands to console.py

mogenson:
Subscribe command will enable notify or indicate events from the
characteristic, depending on supported characteristic properties, and
print recevied values to the output window.

Unsubscribe will stop notify or indicate events.

Rename find_attribute() to find_characteristic() and return a
characteristic for a set of UUIDS, a characteristic for an attribute
handle, or None.

Print read and received values has a hex string.

Add an unsubscribe implementation to gatt_client.py. Reset the CCCD bits
to 0x0000 and remove all subscribers for a characteristic, since no more
notify or indicates events will be comming.

### DIFF
--- a/bumble/gatt_client.py
+++ b/bumble/gatt_client.py
@@ -110,6 +110,9 @@ class CharacteristicProxy(AttributeProxy):
     async def subscribe(self, subscriber=None):
         return await self.client.subscribe(self, subscriber)
 
+    async def unsubscribe(self, subscriber=None):
+        return await self.client.unsubscribe(self, subscriber)
+
     def __str__(self):
         return f'Characteristic(handle=0x{self.handle:04X}, uuid={self.uuid}, properties={Characteristic.properties_as_string(self.properties)})'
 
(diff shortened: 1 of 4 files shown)


bumble/gatt_client.py
```diff
@@ -548,6 +551,23 @@ async def subscribe(self, characteristic, subscriber=None):
...
+
+        # Look for the CCCD descriptor
+        cccd = characteristic.get_descriptor(GATT_CLIENT_CHARACTERISTIC_CONFIGURATION_DESCRIPTOR)
+        if not cccd:
+            logger.warning('unsubscribing to characteristic with no CCCD descriptor')
+            return
+
+        # This removes all subscribers for this attribute from the sets!
+        self.notification_subscribers.pop(characteristic.handle, None)
+        self.indication_subscribers.pop(characteristic.handle, None)
```
mogenson, line 567: It seems useful to have an unsubscribe method.

However, I'm not sure how to handle removing subscribers. Should we remove a specific subscriber that the same caller previously added with subscribe()? How can we figure out which subscriber that is? It's a common pattern to use a lambda function for the subscriber, which means the caller can't go back and get a reference to it.

Since we're setting the CCCD bits to 0, none of the subscribers are going to get any updates for this characteristic. So for now, I'm removing all of them.

bumble/gatt_client.py
```diff
@@ -548,6 +551,23 @@ async def subscribe(self, characteristic, subscriber=None):
...
 
+    async def unsubscribe(self, characteristic):
+        # If we haven't already discovered the descriptors for this characteristic, do it now
+        if not characteristic.descriptors_discovered:
+            await self.discover_descriptors(characteristic)
+
+        # Look for the CCCD descriptor
+        cccd = characteristic.get_descriptor(GATT_CLIENT_CHARACTERISTIC_CONFIGURATION_DESCRIPTOR)
+        if not cccd:
+            logger.warning('unsubscribing to characteristic with no CCCD descriptor')
```
barbibulle, line 562: nit: unsubscribing from

bumble/gatt_client.py
barbibulle: I'd suggest adding a `subscriber` parameter to `unsubscribe`, which would default to `None`. If not `None`, this would just remove that specific subscriber. If `None` it would remove all subscribers. That seems pretty simple. It doesn't cover all cases, because you couldn't selectively remove the lambda subscriber that gets added automatically (without removing the explicit subscribers as well), but I don't think that's a problem.

barbibulle (approved):

bumble/gatt_client.py
mogenson: Sounds good. Removing a matching subscriber if provided. Otherwise removing all subscribers (and lambdas) for attribute.

Added some unsubscribe asserts to tests/gatt_test.py to check logic.

PR #31, opened 2022-08-22
Update generate_company_id_list.py

nirek13:
**Made** code cleaner and shorter


barbibulle (commented):
While this commit makes the code shorter, it makes it more difficult to understand, because it loses the information about row[1] being the company ID and row[2] being the company name.
Since this script is a tool that's only run once in a while to generate/update a file (it isn't runtime code for the library), its performance isn't important. Clarity and maintainability are more important, so I feel that this PR would go against that principle.

PR #33, opened 2022-08-25
add usb_probe tool and improve compatibility

barbibulle:
This PR adds a simple tool to list USB devices and print the strings that may be used as a Bumble transport name to refer to them.
It also improves compatibility with older devices that may not be 100% compliant with the spec w/r/t how event masks are set.
Finally, this fixes a bug when interacting with controllers that report (0,0) for LE-specific ACL buffer size and counts.

NOTE: some USB dongles are very flaky, and can easily get into a non-working or stale state. There doesn't yet seem to be an easy way to catch this situation and reset it automatically using the Python USB API. When that happens, re-opening the device a second or third time seems to work. This needs to be investigated further.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 Bluetooth Stack for Apps, Emulation, Test and Experimentation
 =============================================================
 
-<img src="docs/mkdocs/src/images/logo_framed.png" alt="drawing" width="200" height="200"/>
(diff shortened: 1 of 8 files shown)


mogenson (approved):

turon (commented):
Thanks for adding support for usb_probe! :+1:

PR #40, opened 2022-10-05
maintain the att mtu only at the connection level

barbibulle:
The ATT MTU must be the same for the GATT client and server on the same device for a given connection.
With this PR, the ATT MTU is only maintained in one place (the connection) instead of locally and separately on the client and server.
Added a quick test.


Arrowbox (approved):
LGTM

PR #42, opened 2022-10-09
Refactor and improve the doc for Bumble on Linux

barbibulle:
Small refactor and edits.


docs/mkdocs/src/platforms/linux.md
```diff
@@ -1,48 +1,80 @@
...
+The HCI device referenced by an `hci-socket` transport (`hci<X>`, where `<X>` is an integer, with `hci0` being the first controller device, and so on) must be in the `DOWN` state before it can be opened as a transport.
+You can bring a HCI controller `UP` or `DOWN` with `hciconfig hci<X> up` and `hciconfig hci<X> up`.
+
+!!! tip "HCI Socket Permissions"
+    By default, when running as a regular user, you won't have the permission to use
+    an HCI socket to a Bluetooth controller (you may see an exception like `PermissionError: [Errno 1] Operation not permitted`).
+
+    If you want to run without using `sudo`, you can use the `setcap` command to grant your python runtime the permissions required for raw HCI sockets:
+    ```
+    $ sudo setcap 'cap_net_raw,cap_net_admin=eip' <path/to/python3>'
```
tavip, line 74: This is potentially dangerous since it allows any user to gain net admin privileges. I think a better solution would be to add the current user to  /etc/security/capabilities.conf and just set the inheritable bit on the python runtime.

tavip (approved):

docs/mkdocs/src/platforms/linux.md
barbibulle: Good point. I've pushed a commit with a better/safer recommendation.

PR #43, opened 2022-10-10
support with_response on adapters

barbibulle:
When calling `write_value(..., with_response=True)` on a proxy with a type adapter, the `with_response` parameter has to be supported.

### DIFF
--- a/bumble/gatt.py
+++ b/bumble/gatt.py
@@ -346,8 +346,11 @@ class CharacteristicAdapter:
     async def read_decoded_value(self):
         return self.decode_value(await self.wrapped_characteristic.read_value())
 
-    async def write_decoded_value(self, value):
-        return await self.wrapped_characteristic.write_value(self.encode_value(value))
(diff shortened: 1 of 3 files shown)


bumble/gatt.py
```diff
@@ -346,8 +346,11 @@ def write_encoded_value(self, connection, value):
     async def read_decoded_value(self):
```
Minhaul, line 346: For `write_encoded_value()` above, are the same changes not necessary?

barbibulle: The other `write_encoded_value` in this adapter class is for the GATT server side (when you want to get a callback for a received WRITE), which doesn't have a `with_response` parameter, since that's not passed to the callback handler (the stack will send a response or not, depending on whether the operation is a WRITE request or command, but the handler doesn't know that).

barbibulle: (the `CharacteristicAdapter` class is a general purpose class that can be used both on the client and server side)

Minhaul (approved):

PR #46, opened 2022-10-17
Add runtime switch for filtering by address.

AlanRosenthal:
New commands:
         * scan on filter=address:pattern
         * filter address <filter pattern>

### DIFF
--- a/apps/console.py
+++ b/apps/console.py
@@ -27,6 +27,7 @@ import logging
 import click
 from collections import OrderedDict
 import colors
+import re
 
 from bumble.core import UUID, AdvertisingData, TimeoutError, BT_LE_TRANSPORT
 from bumble.device import ConnectionParametersPreferences, Device, Connection, Peer
@@ -165,6 +166,9 @@ class ConsoleApp:
                     'services': None,
                     'attributes': None,
                     'log': None
+                },
+                'filter': {
+                    'address': None,
                 },
                 'connect': LiveCompleter(self.known_addresses),
                 'update-parameters': None,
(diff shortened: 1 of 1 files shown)


mogenson:
I talked to @AlanRosenthal in person and suggested passing an address value as a parameter to the `scan` command, instead of a startup command line flag.

The `do_scan` function is here: https://github.com/google/bumble/blob/main/apps/console.py#L367

When handling the `scan on` command, you could check if there's a `params[1]` and if it's an address.

barbibulle:
+1 on setting the filter via the console commands rather than CLI parameters.
I would suggest using a syntax like `address=<xyz>` as a scan on/off argument rather than just a positional argument, so that if we want to add other filters/qualifiers later on (like filter by RSSI, or advertisement type, or PHY) we can do that in a non-ambiguous way.

apps/console.py
```diff
@@ -536,6 +546,15 @@ async def do_exit(self, params):
     async def do_quit(self, params):
         self.ui.exit()
 
+    async def do_filter(self, params):
+        if params[0] == "address":
+            if len(params) != 2:
+                self.show_error('invalid syntax', 'expected filter address <pattern>')
+                return
+            self.device.listener.address_filter = params[1]
+        if params[0] == "disable":
```
barbibulle, line 555: since the `filter address <pattern>` takes a regexp, maybe `disable address` isn't necessary, you could just `filter address .*` ?

apps/console.py
```diff
@@ -372,6 +379,9 @@ async def do_scan(self, params):
             else:
                 await self.device.start_scanning()
         elif params[0] == 'on':
+            if len(params) == 2:
```
barbibulle, line 382: consider: parsing extra parameters as a sequence of `<name>=<value>`, with `filter=<filter>` for filtering, and `<filter>` would be `address:<pattern>`. This would make adding other arguments to `scan` easier later on.
ex: `scan on filter=address:01:02:.*`
(maybe overkill?)

AlanRosenthal: added!

apps/console.py
AlanRosenthal: removed disable address

apps/console.py
```diff
@@ -455,6 +459,15 @@ async def do_scan(self, params):
...
+                if not params[1].startswith("filter="):
+                    self.show_error('invalid syntax', 'expected address filter=key1:value1,key2,value. available filters: address')
```
barbibulle, line 464: filter=key1:value1,key2,value --> filter=key1:value1,key2:value

barbibulle (approved):

mogenson (approved):

apps/console.py
AlanRosenthal: :doh: fixed

PR #47, opened 2022-10-17
gbg/extended scanning

barbibulle:
This adds support for using extended scanning (scanning for extended advertisements). Only the scanning part of extended advertising is implemented. Support for the advertising part will come in a separate PR.
Also included is support for monitoring and changing PHYs.


mogenson (approved):
Used console.py from this PR throughout the day yesterday and it performed as expected.

uael (approved):
Everything sounds correct to me

PR #48, opened 2022-10-19
classic: update `Device.connect` to allow parallels connection creation

uael:
According to the specification nothing prevent the Host from creating multiple connections at the same time. This commit add this mechanisme by matching the `connection` and `connection_failure` events against the peer address.

### DIFF
--- a/bumble/core.py
+++ b/bumble/core.py
@@ -100,6 +100,11 @@ class ConnectionError(BaseError):
     FAILURE            = 0x01
     CONNECTION_REFUSED = 0x02
 
+    def __init__(self, error_code, transport, peer_address, error_namespace='', error_name='', details=''):
+        self.transport = transport
+        self.peer_address = peer_address
+        super().__init__(error_code, error_namespace, error_name, details)
+
 
 # -----------------------------------------------------------------------------
 # UUID
--- a/bumble/core.py
+++ b/bumble/core.py
@@ -101,9 +101,9 @@ class ConnectionError(BaseError):
     CONNECTION_REFUSED = 0x02
 
     def __init__(self, error_code, transport, peer_address, error_namespace='', error_name='', details=''):
+        super().__init__(error_code, error_namespace, error_name, details)
         self.transport = transport
         self.peer_address = peer_address
-        super().__init__(error_code, error_namespace, error_name, details)
 
 
 # -----------------------------------------------------------------------------
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -1718,7 +1718,7 @@ class Device(CompositeEventEmitter):
             asyncio.create_task(new_connection())
 
     @host_event_handler
-    def on_connection_failure(self, transport, connection_handle, peer_address, error_code):
+    def on_connection_failure(self, transport, peer_address, error_code):
         logger.debug(f'*** Connection failed: {HCI_Constant.error_name(error_code)}')
 
         # For directed advertising, this means a timeout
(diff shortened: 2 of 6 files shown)


uael:
I'm not sure how to add a proper testing for that logic, can you point me out somewhere I could start looking ?

bumble/device.py
```diff
@@ -1704,11 +1718,11 @@ async def new_connection():
             asyncio.create_task(new_connection())
 
     @host_event_handler
-    def on_connection_failure(self, connection_handle, error_code):
+    def on_connection_failure(self, transport, connection_handle, peer_address, error_code):
```
barbibulle, line 1721: nit: switch the order of the `transport` and `connection_handle` arguments so that they are in the same order as for the `on_connection` method.

bumble/device.py
```diff
@@ -1717,6 +1731,9 @@ def on_connection_failure(self, connection_handle, error_code):
             'hci',
             HCI_Constant.error_name(error_code)
         )
+        error.transport = transport
+        error.connection_handle = connection_handle # FIXME: Connection handle sounds to be a dummy value here
```
barbibulle, line 1735: you're right, the `connection_handle` here isn't useful. There's a `connection_handle` in the underlying HCI event, but in the case of a connection failure, it means nothing. So I suggest actually removing this parameter from the `on_connection_failure` method altogether.

bumble/device.py
```diff
@@ -1717,6 +1731,9 @@ def on_connection_failure(self, connection_handle, error_code):
             'hci',
             HCI_Constant.error_name(error_code)
         )
+        error.transport = transport
```
barbibulle, line 1734: I suggest making `transport` a constructor parameter for `ConnectionError`

bumble/device.py
```diff
@@ -1717,6 +1731,9 @@ def on_connection_failure(self, connection_handle, error_code):
...
+        error.connection_handle = connection_handle # FIXME: Connection handle sounds to be a dummy value here
+        error.peer_address = peer_address
```
barbibulle, line 1736: I suggest making `peer_address` a constructor parameter for `ConnectionError` (with a default to None)

barbibulle (approved):

bumble/device.py
uael: I was doing the change but ConnectionError is also used here https://github.com/google/bumble/blob/main/bumble/rfcomm.py#L637
Do I still proceed ?

uael: I did the change, lmk what you think :+1:

bumble/device.py
```diff
@@ -1704,16 +1718,18 @@ async def new_connection():
             asyncio.create_task(new_connection())
 
     @host_event_handler
-    def on_connection_failure(self, connection_handle, error_code):
+    def on_connection_failure(self, transport, connection_handle, peer_address, error_code):
```
barbibulle, line 1721: With the modified `emit` from line 386 of `host.py`, the `connection_handle` argument is not longer present (which makes sense), so this won't work, you should remove `connection_handle` from here too.

bumble/core.py
```diff
@@ -100,6 +100,11 @@ class ConnectionError(BaseError):
     FAILURE            = 0x01
     CONNECTION_REFUSED = 0x02
 
+    def __init__(self, error_code, transport, peer_address, error_namespace='', error_name='', details=''):
+        self.transport = transport
+        self.peer_address = peer_address
+        super().__init__(error_code, error_namespace, error_name, details)
```
barbibulle, line 106: Maybe move this to the first line of the constructor

barbibulle (approved):

uael:
@barbibulle 
Ok I think we are done now, let me know if there is anything to add/modify.
I let you merge when we feel it's ready :+1:

PR #50, opened 2022-10-20
device: raise a `CommandTimeoutError` error on command timeout

### LINKED ISSUE #49: `Device.send_command` procedure catch `TimeoutError`
In [device.py#L785](https://github.com/google/bumble/blob/main/bumble/device.py#L785) the `send_command` procedure actually catch the `TimeoutError` which does not make it to the caller.
I was running some test and it break at [device.py#L1247](https://github.com/google/bumble/blob/main/bumble/device.py#L1247) with result set to `None`:
```
  File "./bumble/examples/run_classic_connect.py", line 48, in connect
    connection = await device.connect(target_address, transport=BT_BR_EDR_TRANSPORT)
  File "./bumble/bumble/device.py", line 1259, in connect
    if result.status != HCI_Command_Status_Event.PENDING:
AttributeError: 'NoneType' object has no attribute 'status'
```
I may be wrong here but I'm not sure this is the desired behavior.

A command timeout usually mean that controller may be dead and may need to be reboot.
I suggest adding a new `CommandTimeoutError` error type, which can be catches at user level to properly act against the error.

What do you think ?

uael:
Fixes #49 

We may need to add some tests ?

### DIFF
--- a/bumble/core.py
+++ b/bumble/core.py
@@ -91,6 +91,10 @@ class TimeoutError(Exception):
     """ Timeout Error """
 
 
+class CommandTimeoutError(Exception):
+    """ Command Timeout Error """
+
+
 class InvalidStateError(Exception):
     """ Invalid State Error """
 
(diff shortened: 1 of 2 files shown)


barbibulle (approved):

uael:
@barbibulle I'm not sure if this change is complete. Please merge if it sounds OK to you

PR #51, opened 2022-10-21
Use random address in console.py if device config is not provided

mogenson:
If a device configuration is not provided on startup, generate a random BT address instead of using a default static value of "F0:F1:F2:F3:F4:F5". This is helpful to avoid colisions when there are two instances of console.py running nearby.

Testing:
Started console.py and began advertising a few times. Scanned from a second instance of console.py and observed that the advertising address changed with each restart.

### DIFF
--- a/apps/console.py
+++ b/apps/console.py
@@ -20,14 +20,14 @@
 # Imports
 # -----------------------------------------------------------------------------
 import asyncio
-from bumble.hci import HCI_Constant
-import os
-import os.path
 import logging
-import click
-from collections import OrderedDict
-import colors
+import os
+import random
 import re
+from collections import OrderedDict
+
+import click
+import colors
 
 from bumble.core import UUID, AdvertisingData, TimeoutError, BT_LE_TRANSPORT
 from bumble.device import ConnectionParametersPreferences, Device, Connection, Peer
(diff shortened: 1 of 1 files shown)


apps/console.py
```diff
@@ -20,21 +20,22 @@
...
-import os
-import os.path
 import logging
-import click
+import os
+import random
+import re
 from collections import OrderedDict
+
+import click
```
mogenson, line 29: I ran `isort` on the first chunk of imports here.

apps/console.py
```diff
@@ -271,7 +272,8 @@ async def run_async(self, device_config, transport):
             if device_config:
                 self.device = Device.from_config_file_with_hci(device_config, hci_source, hci_sink)
             else:
-                self.device = Device.with_hci('Bumble', 'F0:F1:F2:F3:F4:F5', hci_source, hci_sink)
+                random_address = ':'.join(map(lambda c: f"{c:02X}", random.sample(range(255), 6)))
```
mogenson, line 275: Is it worth logging the generated address?

barbibulle: Would be nice indeed to be able to show which address was picked. The log panel may not be the best output, because it may scroll past that line already when the other log lines are printed at startup. Maybe we could add a ‘device’ tab (Which we could access with `show device`) where we could show that, and maybe other controller info.

barbibulle, line 275: The two ms bits should be 0b11 for static random addresses.

barbibulle (approved):

apps/console.py
mogenson: Added a log. I like the idea of a "device" tab. I'll split that into a different PR.

mogenson: Woops! Fixed.

barbibulle (approved):

PR #53, opened 2022-10-21
Add a show device tab

mogenson:
Show configuration data about the Bumble device. Make this the default tab on startup.

### DIFF
--- a/apps/console.py
+++ b/apps/console.py
@@ -127,7 +127,7 @@ class ConsoleApp:
         self.known_attributes = []
         self.device           = None
         self.connected_peer   = None
-        self.top_tab          = 'scan'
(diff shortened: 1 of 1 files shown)


apps/console.py
```diff
@@ -356,7 +363,7 @@ def show_services(self, services):
         self.services_text.text = lines
         self.ui.invalidate()
 
-    async def show_attributes(self, attributes):
+    def show_attributes(self, attributes):
```
mogenson, line 366: Drive by fix: this was the only show_* method that was async, but it doesn't need to be.

mogenson:
This is what the Device tab looks like.

bumble_show_device

barbibulle (approved):

PR #65, opened 2022-11-09
fix classic connection event filtering

barbibulle:
A previous commit broke `await device.connect()` for Classic connections when an address was passed that wasn't correctly set to PUBLIC. Since BR/EDR addresses can't be random or private, this PR forces the address type used in the comparison when filtering for connections in the event handler.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -1212,15 +1212,16 @@ class Device(CompositeEventEmitter):
         if type(peer_address) is str:
             try:
                 peer_address = Address(peer_address)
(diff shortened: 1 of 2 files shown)


uael:
Isn't it better to raise a value error in this case?

uael:
And add a way to convert an address before calling connect ?

uael (approved):

barbibulle:
> Isn't it better to raise a value error in this case?

Good point. I modified the PR to only set the type explicitly when the address is passed as a string.

uael (approved):

PR #66, opened 2022-11-09
Add `bumble-console --device-config` support for gatt services

AlanRosenthal:
This PR adds support for bumble-console to be preloaded with gatt services via `--device-config`. This PR also adds some type annotations

### DIFF
--- a/.github/workflows/python-build-test.yml
+++ b/.github/workflows/python-build-test.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Get history and tags for SCM versioning to work
       run: |
         git fetch --prune --unshallow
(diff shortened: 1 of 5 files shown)


bumble/att.py
```diff
@@ -684,7 +684,10 @@ def __init__(self, attribute_type, permissions, value = b''):
         EventEmitter.__init__(self)
         self.handle           = 0
         self.end_group_handle = 0
-        self.permissions      = permissions
+        if type(permissions) is str:
```
barbibulle, line 687: in which case would you need to pass the permissions as an integer string?

barbibulle (approved):

PR #68, opened 2022-11-09
Pairing improvements

uael:


bumble/device.py
```diff
@@ -2067,13 +2067,13 @@ def on_authentication_user_confirmation_request(self, connection, code):
...
 
-        can_confirm = pairing_config.delegate.io_capability not in {
+        can_compare = pairing_config.delegate.io_capability not in {
             smp.SMP_NO_INPUT_NO_OUTPUT_IO_CAPABILITY,
             smp.SMP_DISPLAY_ONLY_IO_CAPABILITY
         }
 
         # Respond
-        if can_confirm and pairing_config.delegate:
+        if can_compare and pairing_config.delegate:
```
barbibulle, line 2076: I think the `and pairing_config.delegate` here is a leftover from a previous version and isn't needed (it can't be None anymore, and in fact we access `pairing_config.delegate.io_capability` just above). The same applies to line 2116.

bumble/smp.py
```diff
@@ -715,6 +718,21 @@ def check_expected_value(self, expected, received, error):
             return False
         return True
 
+    def prompt_user_for_confirm(self, next_steps):
```
barbibulle, line 721: should probably be called `prompt_user_for_confirmation`.

barbibulle (approved):

barbibulle:
the change makes sense, but it looks like the tests need to be fixed/changed as well.

bumble/device.py
uael: Done :+1:

bumble/smp.py
uael: Done :+1:

bumble/device.py
```diff
@@ -731,8 +737,7 @@ def __init__(self, name = None, address = None, config = None, host = None, gene
             self.random_address = address
 
         # Setup SMP
-        # TODO: allow using a public address
```
uael, line 734: The following change try to implement that `TODO`

uael:
> the change makes sense, but it looks like the tests need to be fixed/changed as well.

Yes I just fixed :)

I had to add a new logic for this reason:
https://github.com/google/bumble/pull/68/commits/eb46f60c87be98e5819367c19a9d0eb1e750668b#r1018585319

I think the change is "good-enough" for now, but it's kind of a hack, you may have a better solution

barbibulle:
> I think the change is "good-enough" for now, but it's kind of a hack, you may have a better solution

Yes, this works well enough for now. We can revisit it later when we implement extended advertising.

barbibulle (approved):

Issue #67, opened 2022-11-09
Subscribing to characteristic with both notify and indicate returns ATT error


mogenson:
Some BT stacks do not handle setting the notify and indicate CCCD bits simultaneously when subscribing to a characteristic.

For example, Cordio returns an ATT error of 0x80. In the Cordio stack this is defined as `ATT_ERR_VALUE_RANGE`, and in the Bumble stack this is defined as `WRITE_REQUIRES_AUTHORIZATION`. 

https://github.com/packetcraft-inc/stacks/blob/3656312d6b73e2a2c1c8b33ee0385bc199dd97e6/ble-host/sources/stack/att/atts_ccc.c#L231

Should the GATT client `subscribe()` method prefer notify over indicate or indicate over notify? Should we let the caller choose with a separate subscribe with notify and subscribe with indicate method like the `notify_subscribers()`/`indicate_subscribers()` methods? Maybe we can have a keyword argument to `subscribe` that's an enum of values `notify`, `indicate`, or `either`. 

Also, is it invalid to have a characteristic that supports both notify and indicate properties?

https://github.com/google/bumble/blob/main/bumble/gatt_client.py#L562

mogenson:
@AlanRosenthal @Arrowbox FYI

barbibulle:
The 0x80 error from Cordio shouldn't be seen in an ATT error returned over the air. From the code:
```
/** \name ATT Application Error Codes
 * These codes may be sent to application but are not present in any ATT PDU.
 */
/**@{*/
#define ATT_ERR_VALUE_RANGE           0x80      /*!< \brief Value out of range */
```

It is legal for a characteristic to support both indications and notifications, and it is up to the client to decide what flags to write to the CCCD.  It would be odd, however, for a client to ask for both.
The [iOS API](https://developer.apple.com/documentation/corebluetooth/cbperipheral/1518949-setnotifyvalue?language=objc) says:
"...If the specified characteristic’s configuration allows both notifications and indications, calling this method enables notifications only...", so on that platform the client doesn't have a choice.

I think it would make sense for the `subscribe()` method to take an optional preference, which would default to 'prefer notifications'. It should be a preference rather than a hard choice, so that the subscriber would end up subscribing regardless of the server config. A client that really wants just one mode and not the other can always inspect the properties and decide.

mogenson:
Huh strange, I was seeing error 0x80 in the write response:
```
[...]

PR #71, opened 2022-11-11
Add prefer_notify option to gatt_client subscribe()

mogenson:
If characteristic supports Notify and Indicate, the prefer_notify option will subscribe with Notify if True or Indicate if False.

If characteristic only supports one property, Notify or Indicate, that mode will be selected, regardless of the prefer_notify setting.

Tested with a characteristic that supports both Notify and Indicate and verified that prefer_notify sets the desired mode.

### DIFF
--- a/bumble/gatt_client.py
+++ b/bumble/gatt_client.py
@@ -26,20 +26,17 @@
 import asyncio
 import logging
 import struct
+
 from colors import color
 
-from .core import ProtocolError, TimeoutError
-from .hci import *
 from .att import *
-from .gatt import (
-    GATT_CLIENT_CHARACTERISTIC_CONFIGURATION_DESCRIPTOR,
-    GATT_REQUEST_TIMEOUT,
-    GATT_PRIMARY_SERVICE_ATTRIBUTE_TYPE,
-    GATT_SECONDARY_SERVICE_ATTRIBUTE_TYPE,
-    GATT_CHARACTERISTIC_ATTRIBUTE_TYPE,
-    Characteristic,
-    ClientCharacteristicConfigurationBits
-)
+from .core import InvalidStateError, ProtocolError, TimeoutError
+from .gatt import (GATT_CHARACTERISTIC_ATTRIBUTE_TYPE,
+                   GATT_CLIENT_CHARACTERISTIC_CONFIGURATION_DESCRIPTOR,
+                   GATT_PRIMARY_SERVICE_ATTRIBUTE_TYPE, GATT_REQUEST_TIMEOUT,
+                   GATT_SECONDARY_SERVICE_ATTRIBUTE_TYPE, Characteristic,
+                   ClientCharacteristicConfigurationBits)
+from .hci import *
 
 # -----------------------------------------------------------------------------
 # Logging
@@ -115,7 +112,7 @@ class CharacteristicProxy(AttributeProxy):
     async def discover_descriptors(self):
         return await self.client.discover_descriptors(self)
 
-    async def subscribe(self, subscriber=None):
+    async def subscribe(self, subscriber=None, prefer_notify=True):
         if subscriber is not None:
             if subscriber in self.subscribers:
                 # We already have a proxy subscriber
@@ -129,7 +126,7 @@ class CharacteristicProxy(AttributeProxy):
                 self.subscribers[subscriber] = on_change
                 subscriber = on_change
 
-        return await self.client.subscribe(self, subscriber)
+        return await self.client.subscribe(self, subscriber, prefer_notify)
 
     async def unsubscribe(self, subscriber=None):
         if subscriber in self.subscribers:
@@ -547,7 +544,7 @@ class Client:
 
         return attributes
 
-    async def subscribe(self, characteristic, subscriber=None):
+    async def subscribe(self, characteristic, subscriber=None, prefer_notify=True):
         # If we haven't already discovered the descriptors for this characteristic, do it now
         if not characteristic.descriptors_discovered:
             await self.discover_descriptors(characteristic)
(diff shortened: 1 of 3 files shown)


mogenson:
Fix for #67

bumble/gatt_client.py
```diff
@@ -558,23 +555,31 @@ async def subscribe(self, characteristic, subscriber=None):
...
-        bits = ClientCharacteristicConfigurationBits.DEFAULT
-        subscriber_sets = []
-        if characteristic.properties & Characteristic.NOTIFY:
-            bits |= ClientCharacteristicConfigurationBits.NOTIFICATION
-            subscriber_sets.append(self.notification_subscribers.setdefault(characteristic.handle, set()))
-        if characteristic.properties & Characteristic.INDICATE:
-            bits |= ClientCharacteristicConfigurationBits.INDICATION
-            subscriber_sets.append(self.indication_subscribers.setdefault(characteristic.handle, set()))
+        def characteristic_is_notify(characteristic, prefer_notify):
+            if (
```
barbibulle, line 559: suggest: more readable as: `if characteristic.properties & Characteristic.NOTIFY and characteristic.properties & Characteristic.INDICATE`

bumble/gatt_client.py
```diff
@@ -558,23 +555,31 @@ async def subscribe(self, characteristic, subscriber=None):
...
+                characteristic.properties & (Characteristic.NOTIFY | Characteristic.INDICATE)
+            ) == (Characteristic.NOTIFY | Characteristic.INDICATE):
+                return True if prefer_notify else False
```
barbibulle, line 562: `return prefer_notify` would be sufficient

barbibulle (approved):

AlanRosenthal (approved):
lgtm

bumble/gatt_client.py
AlanRosenthal: parens aren't needed

bumble/gatt_client.py
```diff
@@ -558,23 +555,32 @@ async def subscribe(self, characteristic, subscriber=None):
...
 
         # Add subscribers to the sets
-        for subscriber_set in subscriber_sets:
-            if subscriber is not None:
-                subscriber_set.add(subscriber)
-            # Add the characteristic as a subscriber, which will result in the characteristic
-            # emitting an 'update' event when a notification or indication is received
-            subscriber_set.add(characteristic)
+        subscriber_set = subscribers.setdefault(characteristic.handle, set())
+        if subscriber is not None:
```
AlanRosenthal, line 579: `if subscriber:` is sufficient

bumble/gatt_client.py
```diff
@@ -558,23 +555,32 @@ async def subscribe(self, characteristic, subscriber=None):
             logger.warning('subscribing to characteristic with no CCCD descriptor')
```
AlanRosenthal, line 555: OOS, but this warning is wrong. we don't know if theres no CCCD as we just discovered descriptors and haven't checked if CCCD exists

bumble/gatt_client.py
```diff
@@ -558,23 +555,32 @@ async def subscribe(self, characteristic, subscriber=None):
...
+            elif characteristic.properties & Characteristic.NOTIFY:
+                return True
+            elif characteristic.properties & Characteristic.INDICATE:
+                return False
+            else:
+                raise InvalidStateError('characteristic is not notify or indicate')
+
+        bits, subscribers = (
+            (ClientCharacteristicConfigurationBits.NOTIFICATION, self.notification_subscribers)
+            if characteristic_is_notify(characteristic, prefer_notify)
```
AlanRosenthal, line 573: i found this block (571-575) a bit hard to read. it's a tuple + if expression.
consider converting it to a if statement

bumble/gatt_client.py
mogenson: The line above gets the CCCD only if it exists:
`cccd = characteristic.get_descriptor(GATT_CLIENT_CHARACTERISTIC_CONFIGURATION_DESCRIPTOR)`

bumble/gatt_client.py
mogenson: This is the formatting that black chose, so I'll stick with it.

bumble/gatt_client.py
mogenson: This may be a matter of opinion. I prefer the more succinct if-expression to the following if-statement.

```
bits = None
subscribers = None
if characteristic_is_notify(characteristic, prefer_notify):
    bits = ClientCharacteristicConfigurationBits.Notification
    subscribers = self.notification_subscribers
else:
    bits = ClientCharacteristicConfigurationBits.INDICATION
    subscribers = self.indication_subscribers
````

Since they both do exactly the same thing, I'll leave it as-is.

Comment (barbibulle) [reply bumble/gatt_client.py:573]: You could also put the `bits` and `subscribers` assignments in the if/else statement used for `characteristic_is_notify` (and thus get rid of the local function). 
Also, since there's always at least one branch taken, you don't need to initialize `bits` and `subscribers` to `None`, they'll always be set  in one of the branches.

File: bumble/gatt_client.py [hunk 6]
```diff
@@ -558,23 +555,32 @@ async def subscribe(self, characteristic, subscriber=None):
...
 
         # Add subscribers to the sets
-        for subscriber_set in subscriber_sets:
-            if subscriber is not None:
-                subscriber_set.add(subscriber)
-            # Add the characteristic as a subscriber, which will result in the characteristic
-            # emitting an 'update' event when a notification or indication is received
-            subscriber_set.add(characteristic)
+        subscriber_set = subscribers.setdefault(characteristic.handle, set())
+        if subscriber:
```
barbibulle, line 579: comparing to `None` should be explicit here (`if subscriber is not None:`). Even though unlikely, a not-None argument could still be "falsy", and thus would be ignored here.

barbibulle (approved):

bumble/gatt_client.py
mogenson: Ok, I'll change it back.

PR #73, opened 2022-11-15
lower the number of test cases for l2cap in order to speed up the test

barbibulle:
The test was using unnecessarily many combinations, causing the test to take sometimes more than a minute. 
This more limited set should be enough.
(also, fix the assert syntax, parentheses are not needed)


uael (approved):
Look good to me :+1:

PR #77, opened 2022-11-21
Fix for 'Host' object has no attribute 'add_listener'

mogenson:
Pyee's add_listener() method was not added until release 9.0.0. Bumble's setup.cfg specifies a minimum pyee version of 8.2.2.

Remove the call to add_listener() in l2cap.py. If the add_listener() API is prefered over the on(), another solution would be to bump the pyee version requirement.

### DIFF
--- a/apps/console.py
+++ b/apps/console.py
@@ -443,7 +443,10 @@ class ConsoleApp:
         # Discover all services, characteristics and descriptors
         self.append_to_output('discovering services...')
         await self.connected_peer.discover_services()
-        self.append_to_output(f'found {len(self.connected_peer.services)} services, discovering charateristics...')
+        self.append_to_output(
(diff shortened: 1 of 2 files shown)


apps/console.py
```diff
@@ -443,7 +443,10 @@ async def discover_services(self):
         # Discover all services, characteristics and descriptors
         self.append_to_output('discovering services...')
         await self.connected_peer.discover_services()
-        self.append_to_output(f'found {len(self.connected_peer.services)} services, discovering charateristics...')
+        self.append_to_output(
+            f'found {len(self.connected_peer.services)} services,'
+            ' discovering characteristics...'
```
mogenson, line 448: drive by typo fix: charateristics -> characteristics

barbibulle (approved):

PR #80, opened 2022-11-22
Added class CharacteristicDeclaration, gatt_server getters

AlanRosenthal:
* Converted CharacteristicDeclaration implementation to class
* Added ability to get a gatt_server attribute by service UUID, characteristics UUID, descriptor UUID

### DIFF
--- a/bumble/gatt.py
+++ b/bumble/gatt.py
@@ -272,6 +272,23 @@ class Characteristic(Attribute):
 
 
 # -----------------------------------------------------------------------------
+class CharacteristicDeclaration(Attribute):
+    '''
+    See Vol 3, Part G - 3.3.1 CHARACTERISTIC DECLARATION
+    '''
+    def __init__(self, characteristic, value_handle):
+        declaration_bytes = struct.pack(
+            '<BH',
+            characteristic.properties,
+            value_handle
+        ) + characteristic.uuid.to_pdu_bytes()
+        super().__init__(GATT_CHARACTERISTIC_ATTRIBUTE_TYPE, Attribute.READABLE, declaration_bytes)
+        self.characteristic = characteristic
+
+    def __str__(self):
+        return f'CharacteristicDeclaration(handle=0x{self.handle:04X}, value_handle=0x{self.value_handle:04X}, uuid={self.uuid}, properties={Characteristic.properties_as_string(self.properties)})'
+
+# -----------------------------------------------------------------------------
 class CharacteristicValue:
     '''
     Characteristic value where reading and/or writing is delegated to functions
--- a/bumble/gatt_server.py
+++ b/bumble/gatt_server.py
@@ -26,6 +26,7 @@
 import asyncio
 import logging
 from collections import defaultdict
+from typing import Tuple, Optional
 from pyee import EventEmitter
 from colors import color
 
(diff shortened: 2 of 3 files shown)


bumble/gatt_server.py
```diff
@@ -79,6 +80,45 @@ def get_attribute(self, handle):
                 return attribute
         return None
 
+    def get_service_attribute(self, service_uuid: UUID) -> Service:
+        for attribute in self.attributes:
+            if attribute.type == GATT_PRIMARY_SERVICE_ATTRIBUTE_TYPE:
+                if attribute.uuid == service_uuid:
+                    return attribute
+        raise NotFoundError()
```
barbibulle, line 88: would be better to return `None` when not found, to avoid forcing the caller to use a `try`.

bumble/gatt_server.py
```diff
@@ -79,6 +80,45 @@ def get_attribute(self, handle):
                 return attribute
         return None
 
+    def get_service_attribute(self, service_uuid: UUID) -> Service:
+        for attribute in self.attributes:
```
barbibulle, line 84: more concise with an iterator: `return next((attribute in self.attribute if attribute.type == GATT_PRIMARY_SERVICE_ATTRIBUTE_TYPE and attribute.uuid == service_uuid), None)`

bumble/gatt_server.py
```diff
@@ -79,6 +80,45 @@ def get_attribute(self, handle):
...
+        service_handle = self.get_service_attribute(service_uuid)
+
+        for attribute in map(
+            self.get_attribute,
+            range(service_handle.handle, service_handle.end_group_handle + 1),
+        ):
+            if attribute.type == GATT_CHARACTERISTIC_ATTRIBUTE_TYPE:
+                if attribute.uuid == characteristic_uuid:
+                    return (attribute, self.get_attribute(attribute.value_handle))
+        raise NotFoundError()
```
barbibulle, line 102: same comment as above about `None` return.

bumble/gatt_server.py
```diff
@@ -79,6 +80,45 @@ def get_attribute(self, handle):
...
+                if attribute.uuid == service_uuid:
+                    return attribute
+        raise NotFoundError()
+
+    def get_characteristic_attribute(
+        self, service_uuid: UUID, characteristic_uuid: UUID
+    ) -> Tuple[CharacteristicDeclaration, Characteristic]:
+        service_handle = self.get_service_attribute(service_uuid)
+
+        for attribute in map(
```
barbibulle, line 95: consider using an iterator (see comment above)

bumble/gatt.py
```diff
@@ -271,6 +271,26 @@ def __str__(self):
         return f'Characteristic(handle=0x{self.handle:04X}, end=0x{self.end_group_handle:04X}, uuid={self.uuid}, properties={Characteristic.properties_as_string(self.properties)})'
 
 
+# -----------------------------------------------------------------------------
+class CharacteristicDeclaration(Attribute):
```
barbibulle, line 275: this class doesn't seem to be used here. Is the intent that a client would use that class directly for its own purposes?

bumble/gatt.py
```diff
@@ -271,6 +271,26 @@ def __str__(self):
...
+    '''
+    See Vol 3, Part G - 3.3.1 CHARACTERISTIC DECLARATION
+    '''
+    def __init__(self, uuid, properties, value_handle):
```
barbibulle, line 279: Since this declaration is for a `Characteristic`, would it make sense to pass a `Characteristic` object to this constructor, instead of `uuid`, `properties` and `value_handle` separately, and keep a reference as `self.characteristic`?

AlanRosenthal: great suggestion, will do

bumble/gatt.py
AlanRosenthal: sorry, forgot to git-add that section of the PR. see newest revision

bumble/gatt_server.py
AlanRosenthal: updated, let me know what you think

bumble/core.py
```diff
@@ -98,6 +98,8 @@ class CommandTimeoutError(Exception):
 class InvalidStateError(Exception):
     """ Invalid State Error """
 
+class NotFoundError(Exception):
```
uael, line 101: Sounds like not used anymore

bumble/gatt_server.py
```diff
@@ -79,6 +80,65 @@ def get_attribute(self, handle):
...
+            (
+                attribute
+                for attribute in self.attributes
+                if attribute.type == GATT_PRIMARY_SERVICE_ATTRIBUTE_TYPE
+                and attribute.uuid == service_uuid
+            ),
+            None,
+        )
+
+        # return None
```
uael, line 94: Nit: to remove

uael (changes requested):

uael (approved):

barbibulle (approved):

PR #85, opened 2022-11-28
Add `bumble-console --device-config` support for gatt services

AlanRosenthal:
This PR adds support for bumble-console to be preloaded with gatt services via `--device-config`.

This PR also adds some type annotations.

Formally #66

### DIFF
--- a/bumble/gatt.py
+++ b/bumble/gatt.py
@@ -22,6 +22,7 @@
 # -----------------------------------------------------------------------------
 # Imports
 # -----------------------------------------------------------------------------
+from __future__ import annotations
 import asyncio
 import enum
 import types
@@ -187,7 +188,7 @@ class Service(Attribute):
     See Vol 3, Part G - 3.1 SERVICE DEFINITION
     '''
 
-    def __init__(self, uuid, characteristics, primary=True):
+    def __init__(self, uuid, characteristics: list[Characteristic], primary=True):
         # Convert the uuid to a UUID object if it isn't already
         if type(uuid) is str:
             uuid = UUID(uuid)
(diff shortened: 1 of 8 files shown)


tests/gatt_test.py
```diff
@@ -705,6 +705,57 @@ def on_d3_connection(connection):
     assert d2_connection.att_mtu == 50
 
 
+# -----------------------------------------------------------------------------
+@pytest.mark.asyncio
+async def test_char_property_to_string():
```
mogenson, line 710: Do these tests need to be async?

bumble/gatt.py
```diff
@@ -22,6 +22,7 @@
 # -----------------------------------------------------------------------------
 # Imports
 # -----------------------------------------------------------------------------
+from __future__ import annotations
```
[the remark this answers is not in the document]
AlanRosenthal: yeah! the issue I ran into was I wanted to use a type hint that hadn't been defined yet. see https://stackoverflow.com/a/61544901

tests/gatt_test.py
AlanRosenthal: good call out, let me redo them

bumble/gatt.py
```diff
@@ -256,10 +257,17 @@ def properties_as_string(properties):
...
+        return functools.reduce(lambda x,y: x|get_dict_key_by_value(Characteristic.PROPERTY_NAMES, y), properties_str.split(','), 0)
+
+    def __init__(self, uuid, properties, permissions, value = b'', descriptors: list[Descriptor] = []):
         super().__init__(uuid, permissions, value)
         self.uuid        = self.type
-        self.properties  = properties
+        if type(properties) is str:
+            self.properties = Characteristic.string_to_properties(properties)
+        else:
+            self.properties  = properties
```
barbibulle, line 270: nit: extra space

bumble/gatt.py
```diff
@@ -256,10 +257,17 @@ def properties_as_string(properties):
             if properties & p
         ])
 
-    def __init__(self, uuid, properties, permissions, value = b'', descriptors = []):
+    @staticmethod
+    def string_to_properties(properties_str: str):
+        return functools.reduce(lambda x,y: x|get_dict_key_by_value(Characteristic.PROPERTY_NAMES, y), properties_str.split(','), 0)
```
barbibulle, line 262: style: break into multiple lines and spaces after `,` and around `|`  (I suggest `flake8` to detect code style warnings, even though we're not 100% strict about that, it helps) :
```
return functools.reduce(
    lambda x, y: x | get_dict_key_by_value(Characteristic.PROPERTY_NAMES, y),
    properties_str.split(','),
    0
)
```

barbibulle (approved):

bumble/gatt.py
AlanRosenthal: @barbibulle do you think I should verify `get_dict_key_by_value` returns non-None?

barbibulle: I think it's best to let this throw if you pass a property name that's not supported.

PR #87, opened 2022-11-28
add ASHA GATT UUID

yuyangh:
add ASHA related UUIDs


bumble/gatt.py
```diff
@@ -151,6 +151,14 @@
 # Battery Service
 GATT_BATTERY_LEVEL_CHARACTERISTIC = UUID.from_16_bits(0x2A19, 'Battery Level')
 
+# ASHA Service
+ASHA_SERVICE                             = UUID.from_16_bits(0xFDF0, 'Audio Streaming for Hearing Aid')
```
barbibulle, line 155: those should be prefixed with `GATT_` like the other contants in this file.

yuyangh: fixed

PR #88, opened 2022-11-29
Host: spawn each asynchronous task with the right aliveness

uael:


uael:
Failure in CI is unrelated

bumble/device.py
```diff
@@ -479,8 +478,8 @@ def request_pairing(self):
     async def authenticate(self):
         return await self.device.authenticate(self)
 
-    async def encrypt(self):
-        return await self.device.encrypt(self)
+    async def encrypt(self, enable=0x01):
```
barbibulle, line 481: suggest: make `enable` a boolean, and only map it to the constant 0x01 at the lower level (HCI)

bumble/device.py
```diff
@@ -1809,7 +1809,7 @@ def on_encryption_failure(error_code):
                 result = await self.send_command(
                     HCI_Set_Connection_Encryption_Command(
                         connection_handle = connection.handle,
-                        encryption_enable = 0x01
+                        encryption_enable = enable
```
barbibulle, line 1812: with suggestion above, this would be: `0x01 if enable else 0x00`

bumble/device.py
```diff
@@ -2257,6 +2256,9 @@ def on_remote_name_failure(self, connection, address, error):
     def on_connection_encryption_change(self, connection, encryption):
         logger.debug(f'*** Connection Encryption Change: [0x{connection.handle:04X}] {connection.peer_address} as {connection.role_name}, encryption={encryption}')
         connection.encryption = encryption
+        if not connection.authenticated and encryption == 0x02:  # AES
```
barbibulle, line 2259: suggest: make `0x02` a named constant somewhere for `AES`.

bumble/host.py
```diff
@@ -92,7 +92,19 @@ def __init__(self, controller_source = None, controller_sink = None):
         if controller_sink:
             self.set_packet_sink(controller_sink)
 
+    async def flush(self):
```
barbibulle, line 95: is there a need for `flush` other than in a reset? If not, maybe the event should be called `reset`, and move that logic inside the `reset()` method.

bumble/utils.py
```diff
@@ -21,7 +21,7 @@
 import collections
 from functools import wraps
 from colors import color
-from pyee import EventEmitter
+from pyee import EventEmitter as BaseEventEmitter
```
barbibulle, line 24: It might be safer to leave `EventEmitter` as-is, and name the subclass something else (maybe `AbortableEventEmitter`. This will be make less error prone for importers, as they will explicitly decide to import one of the other, not based on where the import comes from (from `pyee` directly or from `utils.py`).

bumble/host.py
```diff
@@ -26,6 +25,7 @@
 from .gatt import *
 from .smp import *
 from .core import ConnectionParameters
+from .utils import EventEmitter
```
barbibulle, line 28: see comment below about naming the `EventEmitter` subclass.

bumble/utils.py
```diff
@@ -59,6 +59,32 @@ def deregister(self, emitter):
     return cls
 
 
+# -----------------------------------------------------------------------------
+class EventEmitter(BaseEventEmitter):
+
+    def abort_on(self, event: str, coro_or_future):
```
barbibulle, line 65: suggest: the 'usual' term for a "coroutine or future" is an "[awaitable](https://docs.python.org/3/library/asyncio-task.html#awaitables)" (and `class typing.Awaitable` for type annotations.

bumble/utils.py
```diff
@@ -59,6 +59,32 @@ def deregister(self, emitter):
     return cls
 
 
+# -----------------------------------------------------------------------------
+class EventEmitter(BaseEventEmitter):
```
barbibulle, line 63: I like how this class simplifies a lot of the multi-event waiting we need to do in the `Device` class. Since async tasks, awaitables and cancellation can be tricky sometimes, it would be great to add some tests to make sure that this class behaves as expected in different situations, and avoid difficult-to-trace regressions if we make modifications to it later.

bumble/host.py
uael: Yes we need `flush` outside of the `reset`

bumble/utils.py
uael: I added a "very" basic test, this might not be enough

barbibulle (approved):

PR #90, opened 2022-11-29
GATT improvements

uael:


bumble/gatt_client.py
```diff
@@ -712,7 +712,7 @@ async def read_characteristics_by_uuid(self, uuid, service):
                     logger.warning(f'bogus handle value: {attribute_handle}')
                     return []
 
-                characteristics_values.append(attribute_value)
+                characteristics_values.append((attribute_handle, attribute_value))
```
barbibulle, line 715: I think we should add a comment /doc on the method to explain that what's returned is a list of (handle, value) pairs now.

barbibulle (approved):

bumble/gatt_client.py
AlanRosenthal: also add a type to function definition so mypy can catch any mismatched usage

barbibulle:
closing as obsolete.

PR #92, opened 2022-11-29
add ASHA profile

yuyangh:


bumble/profiles/asha_service.py
```diff
@@ -0,0 +1,138 @@
...
+# -----------------------------------------------------------------------------
+class AshaService(TemplateService):
+  UUID = GATT_ASHA_SERVICE
+
+  def __init__(self,device:Device):
+    self.device=device
+
+    # Handler for volume control
+    def on_volume_write(connection, value):
+      print('--- VOLUME Write:', value[0])
```
barbibulle, line 47: use `logger.debug()` instead of directly `print()` (same for other uses of `print()` in this file). See other files in the project for examples of how to instantiate a `logger` object with the right namespace.

bumble/profiles/asha_service.py
```diff
@@ -0,0 +1,138 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# -----------------------------------------------------------------------------
+# Imports
+# -----------------------------------------------------------------------------
+import struct
+from ..device import (
+  Device, AdvertisingType
```
barbibulle, line 21: style: use 4 spaces for indentation (through the file).

bumble/profiles/asha_service.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Handler for volume control
+    def on_volume_write(connection, value):
+      print('--- VOLUME Write:', value[0])
+
+    # Handler for audio control commands
+    def on_audio_control_point_write(connection, value):
+      print('--- AUDIO CONTROL POINT Write:', value.hex())
+      opcode = value[0]
+      if opcode == 1:
```
barbibulle, line 53: suggest: define named constants for the opcodes

bumble/profiles/asha_service.py
```diff
@@ -0,0 +1,138 @@
...
+            (AdvertisingData.INCOMPLETE_LIST_OF_16_BIT_SERVICE_CLASS_UUIDS, bytes(GATT_ASHA_SERVICE)),
+            (AdvertisingData.SERVICE_DATA_16_BIT_UUID, bytes(GATT_ASHA_SERVICE) + bytes([
+                0x01,  # Protocol Version
+                capability,  # Capability
+            ]) + bytes(truncated_hisyncid))
+        ])
+    )
+
+    # TODO enable more advertising_type and own_address_type
+    advertising_type = AdvertisingType.UNDIRECTED_CONNECTABLE_SCANNABLE
```
barbibulle, line 134: this is probably not needed, since this is the default advertising type already.

bumble/profiles/asha_service.py
```diff
@@ -0,0 +1,138 @@
...
+  PackedCharacteristicAdapter
+)
+
+
+# -----------------------------------------------------------------------------
+class AshaService(TemplateService):
+  UUID = GATT_ASHA_SERVICE
+
+  def __init__(self,device:Device):
+    self.device=device
```
barbibulle, line 43: it's probably best to define a class that's not bound to a specific device. See comment below on `start_advertising`

bumble/profiles/asha_service.py
```diff
@@ -0,0 +1,138 @@
...
+
+    characteristics = [self.read_only_properties_characteristic,
+                       self.audio_control_point_characteristic,
+                       self.audio_status_characteristic,
+                       self.volume_characteristic,
+                       self.le_psm_out_characteristic]
+
+    super().__init__(characteristics)
+
+  async def start_advertising(self, capability: int,
```
barbibulle, line 118: rather than duplicating the `start_advertising` method of `Device`, I would recommend just creating a factory method that can construct and return the advertising data, and let the client then combine it with its own data (like name, maybe other services) and then start advertising. Only the `SERVICE_DATA_16_BIT_UUID` is really needed in fact, since the `INCOMPLETE_LIST_OF_16_BIT_SERVICE_CLASS_UUIDS` would also be something the client may want to add to.

bumble/profiles/asha_service.py
yuyangh: ack

bumble/profiles/asha_service.py
yuyangh: ack

bumble/profiles/asha_service.py
yuyangh: ack

bumble/profiles/asha_service.py
yuyangh: ack

bumble/gatt.py
```diff
@@ -202,6 +210,13 @@ def __init__(self, uuid, characteristics, primary=True):
         self.characteristics   = characteristics[:]
         self.primary           = primary
 
+    def get_service_advertising_data(self):
```
uael, line 213: `get_data` ?

yuyangh: I am willing to make the change, but just feel `get_data` is a little bit ambiguous as there are many data in GATT, and this advertising_data is generated from service, not directly from GATT stored data. 
@barbibulle Do you have a better naming idea for this? :)

barbibulle: I would recommend a compromise: call this `get_advertising_data` (omit the term `service` since this is a method on the `Service` class already, but make it clear that this returns the `Service Data` for advertising.

bumble/gatt.py
```diff
@@ -202,6 +210,13 @@ def __init__(self, uuid, characteristics, primary=True):
...
+        """
+        Get Service specific advertising data
+        Defined by each Service, default value is empty
```
barbibulle, line 216: I would make the default `None`, because you could have a service that has service-specific data that's empty (that's legal as per the spec I believe), which is different from a servie that has no service-specific data.

bumble/device.py
```diff
@@ -940,6 +940,11 @@ async def start_advertising(
         if self.advertising:
             await self.stop_advertising()
 
+        # add each Service's advertising data
+        for attribute in self.gatt_server.attributes:
```
barbibulle, line 944: it isn't safe to mutate `self.advertisting_data` here. First, this would keep appending every time you call `start_advertisting`. Second, an app might not want to advertise the service data for all its services. That should be a choice of the code that sets `.advertising_data`. So maybe a better way it to create a helper method on the GATT server that can return a list of all the advertising data for all the registered services. The caller would then be able to call something like: `device.advertising_data = some_advertising_data + b''.join(device.get_advertising_service_data())` where (`get_advertising_service_data` would return a (possibly empty) list of advertising data elements, one per service).

bumble/gatt.py
yuyangh: ack

bumble/gatt.py
yuyangh: ack

bumble/device.py
yuyangh: ack

bumble/gatt_server.py
```diff
@@ -66,6 +66,15 @@ def send_gatt_pdu(self, connection_handle, pdu):
     def next_handle(self):
         return 1 + len(self.attributes)
 
+    def get_advertising_service_data(self):
+        service_advertising_data_map = dict[Service, bytes]()
+        for attribute in self.attributes:
+            if isinstance(attribute, Service):
+                if not attribute.get_advertising_data():
```
barbibulle, line 73: this is sub-optimal, as it will call `get_advertising_data()` twice when not None.
Consider instead using the 'walrus' operator and using a comprehension to construct the dictionary, with something like:
```
return {
    attribute: data for attribute in self.attributes 
    if isinstance(attribute, Service) and (data := attribute.get_advertising_data())
}

File: bumble/profiles/asha_service.py [hunk 11]
```diff
@@ -0,0 +1,144 @@
...
+    SUPPORTED_CODEDC_ID = [0x02, 0x01]  # Codec IDs [G.722 at 16 kHz]
+
+    def __init__(self, capability: int, hisyncid: [int]):
+        self.hisyncid = hisyncid
+        self.capability = capability  # Device Capabilities [Left, Monaural]
+        self.render_delay = [00, 00]
+        self.reserved_for_future_use = [00, 00]
+
+        # Four least significant bytes of the HiSyncId.
+        self.truncated_hisyncid = self.hisyncid[:4]
```
barbibulle, line 59: it might not be necessary to store this as a class property, since it is only used in one method below and not expensive to compute as needed.

bumble/profiles/asha_service.py
```diff
@@ -0,0 +1,144 @@
...
+    PROTOCOL_VERSION = 0x01
+    RESERVED_FOR_FUTURE_USE = [00, 00]
+    FEATURE_MAP = [0x01]  # [LE CoC audio output streaming supported]
+    SUPPORTED_CODEDC_ID = [0x02, 0x01]  # Codec IDs [G.722 at 16 kHz]
+
+    def __init__(self, capability: int, hisyncid: [int]):
+        self.hisyncid = hisyncid
+        self.capability = capability  # Device Capabilities [Left, Monaural]
+        self.render_delay = [00, 00]
+        self.reserved_for_future_use = [00, 00]
```
barbibulle, line 56: not needed, there's already a `RESERVED_FOR_FUTURE_USE` class constant.

bumble/profiles/asha_service.py
```diff
@@ -0,0 +1,144 @@
...
+    OPCODE_STATUS = 3
+    PROTOCOL_VERSION = 0x01
+    RESERVED_FOR_FUTURE_USE = [00, 00]
+    FEATURE_MAP = [0x01]  # [LE CoC audio output streaming supported]
+    SUPPORTED_CODEDC_ID = [0x02, 0x01]  # Codec IDs [G.722 at 16 kHz]
+
+    def __init__(self, capability: int, hisyncid: [int]):
+        self.hisyncid = hisyncid
+        self.capability = capability  # Device Capabilities [Left, Monaural]
+        self.render_delay = [00, 00]
```
barbibulle, line 55: why is this a per-instance property, even though it is currently a constant like FEATURE_MAP or SUPPORTED_CODEC_ID?

bumble/profiles/asha_service.py
```diff
@@ -0,0 +1,144 @@
...
+# -----------------------------------------------------------------------------
+class AshaService(TemplateService):
+    UUID = GATT_ASHA_SERVICE
+    OPCODE_START = 1
+    OPCODE_STOP = 2
+    OPCODE_STATUS = 3
+    PROTOCOL_VERSION = 0x01
+    RESERVED_FOR_FUTURE_USE = [00, 00]
+    FEATURE_MAP = [0x01]  # [LE CoC audio output streaming supported]
+    SUPPORTED_CODEDC_ID = [0x02, 0x01]  # Codec IDs [G.722 at 16 kHz]
```
barbibulle, line 50: typo: `CODEDC` --> `CODEC`

yuyangh: ack

bumble/profiles/asha_service.py
yuyangh: ack

bumble/profiles/asha_service.py
yuyangh: ack

bumble/profiles/asha_service.py
yuyangh: ack

bumble/gatt_server.py
yuyangh: ack

PR #93, opened 2022-11-30
Add Device::add_default_services()

AlanRosenthal:
This will allow a test to:
a: add services to a device
b: reset services via `Server()`
c: add the default services back

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -797,9 +797,7 @@ class Device(CompositeEventEmitter):
         # Register the SDP server with the L2CAP Channel Manager
         self.sdp_server.register(self.l2cap_channel_manager)
 
-        # Add a GAP Service if requested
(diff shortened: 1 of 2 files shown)


bumble/device.py
```diff
@@ -808,6 +806,11 @@ def __init__(self, name = None, address = None, config = None, host = None, gene
         # Set the initial host
         self.host = host
 
+    def add_default_services(self, generic_access_service):
```
barbibulle, line 809: Should move this down to just after the `def add_services` method definition.

barbibulle, line 809: suggest adding a default `True` value for `generic_access_service`, to mirror the default used for the `Device` constructor.

barbibulle (approved):

barbibulle (approved):

PR #95, opened 2022-12-09
Fix `show attributes`

AlanRosenthal:
`show attributes` wasn't being populated since `show_attributes()` was never called.

Also updated `show attributes` to match the color and indentation of `show services`

### DIFF
--- a/apps/console.py
+++ b/apps/console.py
@@ -57,7 +57,7 @@ from bumble.core import UUID, AdvertisingData, BT_LE_TRANSPORT
 from bumble.device import ConnectionParametersPreferences, Device, Connection, Peer
 from bumble.utils import AsyncRunner
 from bumble.transport import open_transport_or_link
-from bumble.gatt import Characteristic
+from bumble.gatt import Characteristic, Service, CharacteristicDeclaration, Descriptor
 from bumble.hci import (
     HCI_Constant,
     HCI_LE_1M_PHY,
@@ -154,10 +154,10 @@ class ConsoleApp:
                     'rssi': {'on': None, 'off': None},
                     'show': {
                         'scan': None,
-                        'services': None,
-                        'attributes': None,
                         'log': None,
                         'device': None,
+                        'local-services': None,
+                        'remote-services': None,
                     },
                     'filter': {
                         'address': None,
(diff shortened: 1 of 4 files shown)


apps/console.py
```diff
@@ -370,9 +371,13 @@ def show_services(self, services):
 
     def show_attributes(self, attributes):
         lines = []
-
         for attribute in attributes:
-            lines.append(('ansicyan', f'{attribute}\n'))
+            if type(attribute) is Service:
```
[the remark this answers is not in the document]
AlanRosenthal: what do you think about doing something like:
`show local-service`
`show remote-service`
or 
`show service <local|remote>`

apps/console.py
```diff
@@ -355,24 +356,30 @@ def show_services(self, services):
         lines = []
         del self.known_attributes[:]
         for service in services:
-            lines.append(('ansicyan', str(service) + '\n'))
+            lines.append(('ansicyan', f'{service}\n'))
 
             for characteristic in service.characteristics:
-                lines.append(('ansimagenta', '  ' + str(characteristic) + '\n'))
+                lines.append(('ansimagenta', '  {characteristic}\n'))
```
mogenson, line 362: Need an `f` here, eg `f' {characteristic}\n'`.

apps/console.py
```diff
@@ -355,24 +356,30 @@ def show_services(self, services):
...
         for attribute in attributes:
-            lines.append(('ansicyan', f'{attribute}\n'))
+            if type(attribute) is Service:
+                lines.append(('ansicyan', f'{attribute}\n'))
+            elif type(attribute) in (Characteristic, CharacteristicDeclaration):
+                lines.append(('ansimagenta', f'  {attribute}\n'))
+            elif type(attribute) is Descriptor:
+                lines.append(('ansigreen', f'    {attribute}\n'))
+            else:
+                lines.append(('ansiyellow', f"{attribute}"))
```
mogenson, line 382: nit: mixed single and double quotes in this section

AlanRosenthal: applied black to this section

apps/console.py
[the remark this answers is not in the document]
barbibulle: Yes, that's a good way to handle it.

barbibulle (approved):

barbibulle (approved):

PR #96, opened 2022-12-10
format with Black

barbibulle:
This PR reformats everything with 'black -S'. (the -S option is to keep quotes unchanged)
One exception: most files that start with a "Constants" section have a locally-fenced (fmt: off/on) section to keep the constants as a table.
A subsequent PR will add automation and settings for Black and pylint, but this PR is only for the initial (re)formatting as a single commit, so that it can then be ignored by a `git blame`.


mogenson:
Do you have a strong opinion on single quotes vs double quotes? Although you can use `black -S` to leave quotation unchanged, it would be nice to auto-format code one way or the other instead of relying on developers to match a preferred style.

`black` is an opinionated code formatter and does not provide an option to format strings to single quotes. If you feel strongly about single quotes, maybe one of the `black` forks (like [axblack](https://pypi.org/project/axblack/) or [oitnb](https://pypi.org/project/oitnb/)) would be useful?

mogenson (approved):

AlanRosenthal (approved):

uael (approved):

barbibulle:
> Do you have a strong opinion on single quotes vs double quotes? Although you can use `black -S` to leave quotation unchanged, it would be nice to auto-format code one way or the other instead of relying on developers to match a preferred style.

I do prefer single quotes a lot, but that's a personal preference. The reason I'm punting on it for now, and still leaving quotes unchanged, is because I want to leave room for the opportunity for `black` to eventually support single quotes (as do the people who created the `blue` fork of `black`: "...Instead, our hope is that eventually we’ll be able to work with the black maintainers to add just a little bit of configuration and merge back into the black project. We’d be ecstatic if blue eventually were retired...".
If that doesn't happen in a reasonable time, then maybe we can decide to just go with single quotes. But let's wait a bit... just in case.
NOTE: I may very well use `blue` for my own setup, so that I can keep my quotes consistent.

PR #99, opened 2022-12-12
Add a check in CI for formatting, and `invoke format`

AlanRosenthal:


barbibulle:
Definitely something we want to have. But I’ve starting a PR that incorporates this task as well as a few others for linting, so maybe better to wait for that before merging this one, because in the larger PR the task will be grouped with others under a common task group. I should have it up for review shortly.

PR #101, opened 2022-12-15
formatting linting automation

barbibulle:
This adds tasks for formatting (or checking the format) with `Black` and linting with `pylint`.
There's a `pylint` config with reasonable settings for now. Some linters are disabled, because `pylint` just doesn't work with decorators that change the method or class signatures, and we do use those for all the protocol classes. I hope that `pylint` eventually gets better support for that, but so far, even the basic support for marking some functions as 'mutating decorators' doesn't really work.

There's some new documentation that explains the formatter and linting choices.

Also included are GitHub tasks that will check the format and linting errors (errors only at that point).

Finally, the PR includes changes that should remove all Error and Warning linter errors, as well as most 'Convention' errors (other than `invalid-name` because we still have some short names like `sc` and other names directly ported from the spec, which we want to keep as-is to make it easier to map the code to the spec).

The changes should be only for linting, no functional change.


AlanRosenthal (approved):

uael:
That's pretty cool :+1:

.gitignore
```diff
@@ -8,3 +8,4 @@ test-results.xml
 __pycache__
 # generated by setuptools_scm
 bumble/_version.py
+.vscode/launch.json
```
uael, line 11: Maybe `.vscode/` without file ?

uael (approved):

.gitignore
barbibulle: we do want `.vscode/setttings` for the workspace settings where some basic linter settings are.

PR #103, opened 2022-12-15
Misc device fixes

uael:


bumble/device.py
```diff
@@ -2230,45 +2230,43 @@ def on_connection(
...
+                peer_address,
+                peer_resolvable_address,
+                role,
+                connection_parameters,
+                ConnectionPHY(HCI_LE_1M_PHY, HCI_LE_1M_PHY),
+            )
+            self.connections[connection_handle] = connection
+
+            # Emit an event to notify listeners of the new connection
+            self.emit('connection', connection)
```
barbibulle, line 2256: the problem here is that we're going to emit an initial connection with possibly the wrong PHY, then update its PHY, but the event has already been emitted, so the listener won't seen the change. Even if we emitted an update, it would look like a PHY transition, which isn't really right, since the initial PHY may not be 1M.

uael: I can delay the emit after the read if supported, and just emit otherwise ? That was my first version

uael: Done :+1:

bumble/device.py
```diff
@@ -2230,45 +2230,45 @@ def on_connection(
...
+                self_address,
+                peer_address,
+                peer_resolvable_address,
+                role,
+                connection_parameters,
+                ConnectionPHY(HCI_LE_1M_PHY, HCI_LE_1M_PHY),
+            )
+            self.connections[connection_handle] = connection
+
+            if not self.host.supports_command(HCI_LE_READ_PHY_COMMAND):
```
barbibulle, line 2255: nit: it may be more natural to switch the order of the `if` here (`if x: ... else: ...` instead of `if not x: ... else: ...`)

barbibulle (approved):

PR #104, opened 2022-12-19
fix libusb loading on Windows

barbibulle:
A recent change to use `libusb_package` for loading `libusb` broke support for Windows, because `get_library_path` returns a `WindowsPath`, not an `str`, which isn't iterable.
This PR converts the path to an `str` before invoking `ctypes.CDLL` or `ctypes.WinDLL`.


mogenson (approved):
Thanks

PR #109, opened 2023-01-10
transport: make libusb_package optional

qiaoccolato:


bumble/transport/usb.py
```diff
@@ -45,11 +46,18 @@ def load_libusb():
...
+        sys.modules['libusb_package'] = libusb_package
+        spec.loader.exec_module(libusb_package)
+
+        if libusb_path := libusb_package.get_library_path():
+                logger.debug(f'loading libusb library at {libusb_path}')
+                dll_loader = ctypes.WinDLL if platform.system() == 'Windows' else ctypes.CDLL
+                libusb_dll = dll_loader(str(libusb_path), use_errno=True, use_last_error=True)
+                usb1.loadLibrary(libusb_dll)
+        else:
+            print ("NO LIBUSB PACKAGE MODULE!!!!!!!!!!!!")
```
uael, line 60: I think this can fail silently (see above comment)

bumble/transport/usb.py
```diff
@@ -45,11 +46,18 @@ def load_libusb():
...
-        dll_loader = ctypes.WinDLL if platform.system() == 'Windows' else ctypes.CDLL
-        libusb_dll = dll_loader(str(libusb_path), use_errno=True, use_last_error=True)
-        usb1.loadLibrary(libusb_dll)
+    if (spec := importlib.util.find_spec('libusb_package')) is not None:
+        libusb_package = importlib.util.module_from_spec(spec)
+        sys.modules['libusb_package'] = libusb_package
+        spec.loader.exec_module(libusb_package)
+
+        if libusb_path := libusb_package.get_library_path():
+                logger.debug(f'loading libusb library at {libusb_path}')
```
uael, line 55: nit: bad indentation

bumble/transport/pyusb.py
```diff
@@ -205,16 +206,23 @@ async def close(self):
             await self.sink.stop()
             usb.util.release_interface(self.device, 0)
 
+    device_finder =  usb.core.find
```
uael, line 209: `usb_find` ?

uael (changes requested):

bumble/transport/usb.py
qiaoccolato: according to commit/80fe2ea4224de10728760ff49db08ef2a5c59cde . If the library doesn't exist, `load_libusb` just do nothing silently. I don't see this as a fail.

bumble/transport/pyusb.py
qiaoccolato: I wanted to share the "device_finder" between "usb.core.find" and "libusb_package.find".

bumble/transport/usb.py
qiaoccolato: the bad ind and the log are removed.

bumble/transport/pyusb.py
uael: no I mean rename `device_finder` to `usb_find`

uael (approved):

uael:
LGTM, please rebase

bumble/transport/pyusb.py
```diff
@@ -205,16 +206,23 @@ async def close(self):
...
+    if (spec := importlib.util.find_spec('libusb_package')) is not None:
```
barbibulle, line 210: wouldn't `try: import libusb_package` be simpler here?

qiaoccolato: just a python code style, it's not recommended to use `import` in between functions, only at the top of the script. The importlib is doing the same job to load modules. I don't have strong preference here. I don't need to add the module to `sys`, I can do only a one-time call on the module.

barbibulle: The code style recommendation is mainly to avoid local imports of modules when it's not actually needed. But in this case, it really is about trying to import the module locally on purpose, so using `import` is justified. And this way, we're sure that we get the exact standard behavior of `import`, no risk to deviate from it subtly in a way that may be hard to troubleshoot.

bumble/transport/pyusb.py
```diff
@@ -205,16 +204,24 @@ async def close(self):
             await self.sink.stop()
             usb.util.release_interface(self.device, 0)
 
+    usb_find =  usb.core.find
+    try:
+        import libusb_package
+    except:
```
barbibulle, line 210: `except ImportError` would be better (avoid broad exception catching)

bumble/transport/usb.py
```diff
@@ -45,11 +44,16 @@ def load_libusb():
...
     when usb1.USBContext is created.
     '''
-    if libusb_path := libusb_package.get_library_path():
-        logger.debug(f'loading libusb library at {libusb_path}')
-        dll_loader = ctypes.WinDLL if platform.system() == 'Windows' else ctypes.CDLL
-        libusb_dll = dll_loader(str(libusb_path), use_errno=True, use_last_error=True)
-        usb1.loadLibrary(libusb_dll)
+    try:
+        import libusb_package
+    except:
```
barbibulle, line 49: same comment as above.

barbibulle (approved):

PR #111, opened 2023-01-13
don't set a random address when it is 00:00:00:00:00:00

barbibulle:
Some controllers don't like it when invoking HCI_LE_Set_Random_Address_Command with an all-0s address.
So, when using a device with a config where no random address is set, don't send such an address.

### DIFF
--- a/bumble/hci.py
+++ b/bumble/hci.py
@@ -1638,8 +1638,8 @@ class HCI_Object:
             # Map the value if needed
             if value_mappers:
                 value_mapper = value_mappers.get(key, value_mapper)
-                if value_mapper is not None:
-                    value = value_mapper(value)
+            if value_mapper is not None:
+                value = value_mapper(value)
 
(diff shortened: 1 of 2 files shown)


bumble/hci.py
```diff
@@ -1638,8 +1638,8 @@ def format_fields(hci_object, keys, indentation='', value_mappers=None):
             # Map the value if needed
             if value_mappers:
                 value_mapper = value_mappers.get(key, value_mapper)
-                if value_mapper is not None:
-                    value = value_mapper(value)
+            if value_mapper is not None:
```
barbibulle, line 1641: this fixes a regression from a previous PR.

uael:
I suggest using a `HCI_LE_Rand` command to create a random address before send instead

uael:
Perfect!

uael (approved):

PR #112, opened 2023-01-19
Add timer to auto update local-services screen

AlanRosenthal:
The local services screen might update via gatt write.

If you call `show local-services`, it will start a timer to update the screen once a second

### DIFF
--- a/apps/console.py
+++ b/apps/console.py
@@ -281,7 +281,7 @@ class ConsoleApp:
             self.device.listener = DeviceListener(self)
             await self.device.power_on()
             self.show_device(self.device)
-            self.show_local_services(self.device.gatt_server.attributes)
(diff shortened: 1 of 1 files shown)


apps/console.py
```diff
@@ -660,6 +662,13 @@ async def do_advertise(self, params):
         else:
             self.show_error('unsupported arguments for advertise command')
 
+    async def update_timer(self):
+        if self.top_tab == 'local-services':
+            self.update_local_services(self.device.gatt_server.attributes)
+            self.ui.invalidate()
+            await asyncio.sleep(1)
+            await self.update_timer()
```
uael, line 670: I suggest using a loop instead, only for readability purpose.
Is tail call elimination supported for asynchronous function ?

PR #113, opened 2023-01-20
add basic support for mypy type checking

barbibulle:
Only the basic support for type checking here, including pre-commit checks.
Actual type annotations can be added later on, gradually, as needed.


uael (approved):
I'm going to rebase on this right now.
However I propose using `dict`, `list` and `tuple` variants instead of the ones from typing. If it's Ok for you I'm gonna change this + add types for `core`, `utils`, `hci`, `host` and `device` + an almost finished mypy plugin for `command` like decorators

barbibulle:
> I'm going to rebase on this right now.
> However I propose using `dict`, `list` and `tuple` variants instead of the ones from typing. If it's Ok for you I'm gonna change this + add types for `core`, `utils`, `hci`, `host` and `device` + an almost finished mypy plugin for `command` like decorators

Unfortunately using the builtin dict and tuple types is only possible starting with Python 3.9, but we still support 3.8. At some point in the future when we deprecate support for 3.8 we can update to the new types, but for now we still need to stick to the uppercase ones.

PR #116, opened 2023-01-25
fix libusb-package dependency

barbibulle:
The previous version was missing `tomli` as a dependency, thus breaking on some platforms.


barbibulle:
force-merging this hot fix as it should unblock users who need to install on some platforms.

PR #117, opened 2023-01-27
host: fixed `.latency` attribute error

uael:
Please double check

### DIFF
--- a/bumble/host.py
+++ b/bumble/host.py
@@ -660,7 +660,7 @@ class Host(AbortableEventEmitter):
                 connection_handle=event.connection_handle,
                 interval_min=event.interval_min,
                 interval_max=event.interval_max,
-                latency=event.latency,
+                max_latency=event.max_latency,
                 timeout=event.timeout,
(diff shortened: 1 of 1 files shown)


bumble/host.py
```diff
@@ -660,7 +660,7 @@ def on_hci_le_remote_connection_parameter_request_event(self, event):
                 connection_handle=event.connection_handle,
                 interval_min=event.interval_min,
                 interval_max=event.interval_max,
-                latency=event.latency,
+                max_latency=event.max_latency,
```
barbibulle, line 663: Yes! Somehow this fell through the cracks when I renamed `latency` to match the wording in the spec.

barbibulle (approved):

bumble/host.py
AlanRosenthal: what kind of bug would this be?

uael: `latency` was renamed with the `max_` prefix in both the event and command

PR #120, opened 2023-02-04
minor cleanup of the internals of the usb transport implementation

barbibulle:
Move the event processing thread from the `UsbSource` class  to the `UsbTransport` class (it doesn't logically belong to just the source, since the sink also needs callbacks).

Remove unused `context` property in `UsbPacketSource`

Minor re-naming of some properties.


mogenson (approved):
lgtm

Issue #124, opened 2023-02-08
JUST_WORKS bonding


cbout22:
Hi,

I need help understanding how pairing works from `apps/pair.py`. While using this app, I got an error message when I try to connect with an iPhone: (except for first connection)
```
Failed to Connect: Peer removed pairing information
```

Furthermore, when I disconnect and reconnect an Android, I have to redo pairing.

I configure it as follows :
- mode : "le"
- request = True
- device1.json

So my questions are: How to store and use pairing information with Bumble for a Just_Works pairing method ? And how can we use or are we supposed to use PairingDelegate  to create a device with a just works bonding?

I really appreciate any help you can provide.

uael:
You may need a persistent storage, do your config file passed to the app include a `keystore` entry?

cbout22:
I passed device1.json form examples files as config file. It contains the following field :
```json
"keystore":"JsonKeyStore"
```
Which refers to 'bumble.keys.JsonKeyStore', if I understand it well. But no `@@@ Pairing keys:` appears when I start 'pair.py'.

I've try to pass a file with the following field :
```json
"keystore":"JsonKeyStore:<absolutePath>/foo.json"
```
 but the following error appends (I think I don't understand syntax for keystore  field😄 ) :
```bash
File "/home/cbout/Documents/bumble/keys.py", line 220, in from_device_config
    filename = params[1]

IndexError: list index out of range
```

Furthermore, I passed foo.json as parameter using `--keystore-file <absolutePath>/foo.json` and pairing keys are saved into this file with peer_adress, irk and ltk.

I tried to toggle/untoggle sc and mtim  fields but nothing seems changed

barbibulle:
Hi.
LE pairing with Android should work.
The error you're seeing when specifying a keystore file name in a device config is a bug :-( (I'll send a PR shortly for that).
If you just use `"keystore": "JsonKeyStore"` you'll get a keystore file in a default location. The debug logs will show you where that it. It is platform-dependent. For exampe, on a mac, that's `~/Library/Application Support/Bumble/Pairing/<xxx>.json`.
As you found out, you can also use `--keystore-file` and pass a path to your keystore file.

To find out more where things fail for you, could you run the app with the env var `BUMBLE_LOGLEVEL` set to `debug`.
Ex:
`BUMBLE_LOGLEVEL=debug python apps/pair.py examples/device1.json usb:0 XX:XX:XX:XX:XX:XX`

cbout22:
Hi,

[...]

PR #125, opened 2023-02-08
fix getting the filename from the keystore option.

barbibulle:
As reported in #124 specifying a filename in a device config when using the `JsonKeyStore` keystore would fail.
Ex: `"keystore": "JsonKeyStore:mykeys.json"`


uael (approved):
lgtm

PR #126, opened 2023-02-09
smp: fix use on renamed keyword arguments

uael:
`digits` was renamed to `_digits` which removed the ability to use `PairingDelegate` directly


uael:
@barbibulle just updated with a proper fix :+1:

barbibulle:
See #129 for an alternative fix.

uael:
Closing as duplicate of #129

PR #128, opened 2023-02-10
Small patches for device and SMP

uael:
I added this feature myself, but it actually make absolutely no sense at all.. 😅


uael:
Sorry I have to fixes the tests!

PR #129, opened 2023-02-13
improve smp compatibility with other OS flows

barbibulle:
The timing when pairing with iOS/macOS with passkey wasn't correct (we would not send the peripheral public key soon enough).
Also included in this PR:
* remove dependency on aioconsole for the pair.py app
* fix the PairingDelegate base class to not use `_xxx` parameter names
* fix for pairing being stuck when pairing without secure connections

### DIFF
--- a/apps/pair.py
+++ b/apps/pair.py
@@ -19,8 +19,8 @@ import asyncio
 import os
 import logging
 import click
-import aioconsole
 from colors import color
+from prompt_toolkit.shortcuts import PromptSession
 
 from bumble.device import Device, Peer
 from bumble.transport import open_transport_or_link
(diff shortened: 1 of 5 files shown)


apps/pair.py
```diff
@@ -58,7 +72,19 @@ def __init__(self, mode, connection, capability_string, prompt):
...
+
+    def print(self, message):
+        print(color(message, 'yellow'))
+
+    async def prompt(self, message):
+        # Wait a bit to allow some of the log lines to print before we prompt
+        await asyncio.sleep(1)
+
+        session = PromptSession(message)
+        # with patch_stdout.patch_stdout(raw=True):
```
uael, line 85: to remove ?

bumble/smp.py
```diff
@@ -518,13 +518,14 @@ async def accept(self) -> bool:
...
-    async def compare_numbers(self, _number: int, _digits: int = 6) -> bool:
+    # pylint: disable-next=unused-argument
+    async def compare_numbers(self, number: int, digits: int) -> bool:
         return True
 
     async def get_number(self) -> int:
         return 0
 
-    async def display_number(self, _number: int, _digits: int = 6) -> None:
+    async def display_number(self, number: int, digits: int) -> None:
```
uael, line 528: add `# pylint: disable-next=unused-argument` before

uael (commented):
lgtm :+1:
The changes to SMP all make senses

apps/pair.py
barbibulle: yes, thanks.

bumble/smp.py
barbibulle: done

uael (approved):

PR #131, opened 2023-03-01
Add Btsnoop log support

zxzxwu:


barbibulle (commented):
This method works but may be somewhat too hard-coded. I will propose an alternative shortly and we can review it.

PR #132, opened 2023-03-02
Fix UUID byte-order in serialization

rahularya50:


bumble/core.py
```diff
@@ -144,9 +144,12 @@ def __init__(
 class UUID:
     '''
     See Bluetooth spec Vol 3, Part B - 2.5.1 UUID
+
+    Note that this class expects and works in little-endian byte-order throughout.
+    The exception is when interacting with strings, which are in big-endian byte-order.
     '''
 
-    BASE_UUID = bytes.fromhex('00001000800000805F9B34FB')
+    BASE_UUID = bytes.fromhex('00001000800000805F9B34FB')[::-1] # little-endian
```
barbibulle, line 152: I suggest putting it directly in little endian rather than converting: "BF43B9F50800000800010000"

bumble/core.py
```diff
@@ -209,13 +212,16 @@ def parse_uuid_2(cls, uuid_as_bytes, offset):
...
+        Serialize UUID in little-endian byte-order
+        '''
+        if not force_128:
             return self.uuid_bytes
 
-        if len(self.uuid_bytes) == 4:
-            return self.uuid_bytes + UUID.BASE_UUID
-
-        return self.uuid_bytes + bytes([0, 0]) + UUID.BASE_UUID
+        match len(self.uuid_bytes):
```
barbibulle, line 221: Unfortunately, `match` is a >= 3.10 feature, but we still need to support >= 3.8.

barbibulle (commented):
Thanks for the PR. I can't believe we have not run into this bug before!

bumble/core.py
uael: I would prefer the current solution, since it's the same has in the spec it's easier to understand

barbibulle: Makes sense. Let's keep it like this then.

uael (approved):

barbibulle (approved):

rahularya50:
I can't submit this, since I don't have write access. Would one of you mind doing so? Thanks

PR #133, opened 2023-03-02
Black format all files

zxzxwu:


barbibulle:
The project still uses `black == 22.10` which produces a different output. I need to check why this PR isn't breaking theGH action checks. If you run `inv project.pre-commit` in a python env with the project's dependencies installed, you should be able to check against the right version of the formatter.

PR #134, opened 2023-03-02
Add Btsnoop log support

zxzxwu:
Change branch.


barbibulle:
See #135 for a refactored implementation that offers a less-tight coupling, and also avoids non-portable filesystem access like "/tmp".

PR #137, opened 2023-03-03
Add device config and use random identity address during key distribution

bidsharmag:
Add two more device configs: `classic_enabled` and `random_identity_address`.
Use `random_identity_address` during key distribution if the device is configured with one.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -729,11 +729,13 @@ class DeviceConfiguration:
         # Setup defaults
         self.name = DEVICE_DEFAULT_NAME
         self.address = Address(DEVICE_DEFAULT_ADDRESS)
+        self.random_identity_address = None
         self.class_of_device = DEVICE_DEFAULT_CLASS_OF_DEVICE
         self.scan_response_data = DEVICE_DEFAULT_SCAN_RESPONSE_DATA
         self.advertising_interval_min = DEVICE_DEFAULT_ADVERTISING_INTERVAL
         self.advertising_interval_max = DEVICE_DEFAULT_ADVERTISING_INTERVAL
         self.le_enabled = True
+        self.classic_enabled = False
         # LE host enable 2nd parameter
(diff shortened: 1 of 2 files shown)


uael:
I may be wrong on this, but I think doing this would do the trick (before `power_on` or before `reset` in avatar):
```
device.address = Address(b'...', Address.RANDOM_IDENTITY_ADDRESS)
```

bidsharmag:
Yes it can work only for case where I just want address to be `Identity Address`. For the test I am writing, I want the device to advertise with random address but return `Identity Address` during key distribution phase of SMP.

barbibulle:
> Yes it can work only for case where I just want address to be `Identity Address`. For the test I am writing, I want the device to advertise with random address but return `Identity Address` during key distribution phase of SMP.

I think I understand what you need here. Adding a `random_identity_address` property to store that address in the device object is probably not the right model. I do have a branch on which I have worked on adding support for RPA, which does take care of advertising with a resolvable private address. I'll try to share that as a PR very soon, I think it would give you what you want.

bumble/device.py
```diff
@@ -968,6 +975,7 @@ def __init__(
             config = DeviceConfiguration()
         self.name = config.name
         self.random_address = config.address
+        self.random_identity_address = config.random_identity_address
```
AlanRosenthal, line 978: Please add a type for this

bumble/device.py
```diff
@@ -976,6 +984,7 @@ def __init__(
         self.keystore = KeyStore.create_for_device(config)
         self.irk = config.irk
         self.le_enabled = config.le_enabled
+        self.classic_enabled = config.classic_enabled
```
AlanRosenthal, line 987: might as well add a type for this too, even though it's just a bool

zxzxwu:
Resolved in #231

PR #139, opened 2023-03-06
two small hotfixes

barbibulle:
one listener method was missing from the base class, and an HCI command was not expecting the right result type.


uael (approved):
lgtm :+1:

PR #142, opened 2023-03-14
Fix small bug with services set via --device-config

AlanRosenthal:
before:
```
  File "/home/alanrosenthal/code/fitbit/bumble/bumble/gatt.py", line 572, in __str__
    f'Descriptor(handle=0x{self.handle:04X}, '
  File "/home/alanrosenthal/code/fitbit/bumble/bumble/att.py", line 756, in read_value
    self.permissions & self.READ_REQUIRES_ENCRYPTION
TypeError: unsupported operand type(s) for &: 'str' and 'int'
```

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -999,7 +999,7 @@ class Device(CompositeEventEmitter):
                 new_characteristic = Characteristic(
                     uuid=characteristic["uuid"],
                     properties=characteristic["properties"],
-                    permissions=int(characteristic["permissions"], 0),
+                    permissions=characteristic["permissions"],
                     descriptors=descriptors,
                 )
                 characteristics.append(new_characteristic)
--- a/bumble/att.py
+++ b/bumble/att.py
@@ -23,11 +23,12 @@
 # Imports
 # -----------------------------------------------------------------------------
 from __future__ import annotations
+import functools
 import struct
 from pyee import EventEmitter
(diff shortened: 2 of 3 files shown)


bumble/device.py
```diff
@@ -993,7 +993,7 @@ def __init__(
                 for descriptor in characteristic.get("descriptors", []):
                     new_descriptor = Descriptor(
                         attribute_type=descriptor["descriptor_type"],
-                        permissions=descriptor["permission"],
+                        permissions=int(descriptor["permission"], 0),
```
barbibulle, line 996: Shouldn't the JSON property be an int? Allowing a string where an int is expected is probably setting the wrong expectation. If we want to allow permissions to be a string, we could allow passing an array of constant names and parse the strings into ints.

AlanRosenthal: did a quick experiment:
```
>>> type(json.loads('{"x":9}')["x"])
<type 'int'>
>>> type(json.loads('{"x":"9"}')["x"])
<type 'unicode'>
>>> type(json.loads('{"x":"0x10"}')["x"])
<type 'unicode'>
```

Overall that makes sense. I was setting permission to "0x01" as a way to set the characteristics as readable, which ironically isn't very readable. 

proper fix incoming

barbibulle (approved):

PR #144, opened 2023-03-20
Add Classic Bluetooth link support

zxzxwu:
I want to test some features which can only work on Classic Bluetooth, such as role switching(in the following commits). However, we cannot rely on RootCanal because it's not mature as well.

The first commit centralize Local Linked Devices usage to a shared util module so that the following change won't change a lot. Then, I rename the original `LocalLink` class to `LocalLeLink`, and add a `LocalClassicLink` class to handle classic transport.

For the Remote Link path, I am not sure what we should do for it.

### DIFF
--- a/bumble/controller.py
+++ b/bumble/controller.py
@@ -21,7 +21,12 @@ import itertools
 import random
 import struct
 from bumble.colors import color
-from bumble.core import BT_CENTRAL_ROLE, BT_PERIPHERAL_ROLE
+from bumble.core import (
+    BT_CENTRAL_ROLE,
+    BT_PERIPHERAL_ROLE,
+    BT_LE_TRANSPORT,
+    BT_BR_EDR_TRANSPORT,
+)
 
 from bumble.hci import (
     HCI_ACL_DATA_PACKET,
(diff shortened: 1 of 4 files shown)


bumble/controller.py
```diff
@@ -432,16 +481,19 @@ def on_link_peripheral_disconnected(self, peer_address):
...
             )
 
-    def on_link_acl_data(self, sender_address, data):
+    def on_link_acl_data(self, sender_address, transport, data):
         # Look for the connection to which this data belongs
-        connection = self.find_connection_by_address(sender_address)
+        if transport == BT_LE_TRANSPORT:
+            connection = self.find_le_connection_by_address(sender_address)
+        else:
+            connection = self.find_classic_connection_by_handle(sender_address)
```
barbibulle, line 496: this looks odd, calling `find_classic_connection_by_handle` with an address instead of a handle.

zxzxwu: Done. Making a2dp test on BR/EDR to verify this.

bumble/controller.py
```diff
@@ -82,23 +94,33 @@ def on_hci_acl_data_packet(self, packet):
...
         self.hci_sink = None
         self.link = link
 
-        self.central_connections = (
-            {}
-        )  # Connections where this controller is the central
-        self.peripheral_connections = (
-            {}
-        )  # Connections where this controller is the peripheral
+        self.central_connections: dict[
```
uael, line 115: nit: Use `Dict` instead of `dict`

uael (approved):

bumble/controller.py
zxzxwu: Done

barbibulle (approved):

uael (approved):

PR #146, opened 2023-03-20
add auto-snooping for transports

barbibulle:
When the BUMBLE_SNOOPER env variable is set, a Snooper instance will be automatically created when a transport is instantiated.
A file-based scheme is implemented, where a filename pattern is used to construct the path of the snoop log file.
ex:
export BUMBLE_SNOOPER=btsnoop:file:/tmp/bumble_{now:%Y-%m-%d-%H:%M:%S}_{pid}.log

--> when a transport is created, a BtSnooper instance will be instantiated, logging to the file: /tmp/bumble_2023-03-20-11:15:24_17011.log

### DIFF
--- a/bumble/snoop.py
+++ b/bumble/snoop.py
@@ -144,7 +144,7 @@ def create_snooper(spec: str) -> Generator[Snooper, None, None]:
 
     if snooper_type == 'btsnoop':
         if ':' not in snooper_args:
-            raise ValueError('I/O type for btsnoop type missing')
+            raise ValueError('I/O type for btsnoop snooper type missing')
 
         io_type, io_name = snooper_args.split(':', maxsplit=1)
         if io_type == 'file':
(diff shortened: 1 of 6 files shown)


bumble/snoop.py
```diff
@@ -79,15 +89,82 @@ def snoop(self, hci_packet: HCI_Packet, direction: Snooper.Direction) -> None:
...
+
+            # Open the file
+            logger.debug(f'Snoop file: {file_path}')
+            with open(file_path, 'wb') as snoop_file:
+                _SNOOPER_INSTANCE_COUNT += 1
+                yield BtSnooper(snoop_file)
+                _SNOOPER_INSTANCE_COUNT -= 1
+                return
+
+        raise ValueError('I/O type not supported')
```
uael, line 168: nit: add `{io_type}` to error message

uael (approved):
Cool one ! I'm gonna add support into avatar asap

zxzxwu (approved):

PR #147, opened 2023-03-21
add benchmark tool and doc

barbibulle:
This is a "swiss army knife" benchmarking tool that can be used to test the throughput and/or latency between two devices. 
Supported modes:
* GATT
* L2CAP CoC
* RFComm

Supported roles:
* Server (sends data as fast as possible)
* Receiver (receives data from a Server role and sends back a final ACK when the server is done sending)
* Ping (sends periodic PINGs)
* Ping (responds to PINGs)
* 
One device runs as the Central (with the ability to specify a custom Connection Interval and PHY), the other as the Peripheral.
The modes are independent of the roles and of the central/peripheral choice. So, for example, you can have a Central be a Sender as a GATT client (default), and the Peripheral be the Receiver as a GATT server. But you could also have the Central be a Receiver as a GATT client, or a Sender as a GATT server, etc.

### DIFF
--- a/apps/bench.py
+++ b/apps/bench.py
@@ -323,7 +323,7 @@ class Receiver:
         self.expected_packet_index = packet_index + 1
 
         if packet_flags & PACKET_FLAG_LAST:
-            AsyncRunner.run(
+            AsyncRunner.spawn(
                 self.packet_io.send_packet(
                     struct.pack('>bbI', PacketType.ACK, packet_flags, packet_index)
                 )
@@ -423,7 +423,7 @@ class Ping:
             self.done.set()
             return
 
-        AsyncRunner.run(self.send_next_ping())
+        AsyncRunner.spawn(self.send_next_ping())
 
 
 # -----------------------------------------------------------------------------
@@ -472,7 +472,7 @@ class Pong:
 
         self.expected_packet_index = packet_index + 1
 
-        AsyncRunner.run(
+        AsyncRunner.spawn(
             self.packet_io.send_packet(
                 struct.pack('>bbI', PacketType.ACK, packet_flags, packet_index)
             )
--- a/apps/bench.py
+++ b/apps/bench.py
@@ -1,0 +1,1206 @@
+# Copyright 2021-2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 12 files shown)


apps/bench.py
```diff
@@ -0,0 +1,1206 @@
...
+                f'Speed: instant={instant_rx_speed:.4f}, average={average_rx_speed:.4f}',
+                'yellow',
+            )
+        )
+
+        self.last_timestamp = now
+        self.expected_packet_index = packet_index + 1
+
+        if packet_flags & PACKET_FLAG_LAST:
+            asyncio.create_task(
```
uael, line 325: Because of https://github.com/python/cpython/issues/88831, I suggest to push the packet to a queue and to loop get/send in // of `self.done.wait()` in `async def run(self)` bellow

apps/bench.py
```diff
@@ -0,0 +1,1206 @@
...
+                        f'!!! Unexpected packet, expected {self.current_packet_index} '
+                        f'but received {packet_index}'
+                    )
+                )
+
+        if packet_flags & PACKET_FLAG_LAST:
+            self.done.set()
+            return
+
+        asyncio.create_task(self.send_next_ping())
```
uael, line 425: Same as above

uael (approved):

apps/bench.py
barbibulle: Fixed in latest commit.

bumble/device.py
```diff
@@ -1242,6 +1243,13 @@ async def power_on(self) -> None:
         # Done
         self.powered_on = True
 
+    async def power_off(self) -> None:
+        if self.powered_on:
+            await self.host.reset()
```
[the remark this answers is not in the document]
barbibulle: Done.

docs/mkdocs/src/apps_and_tools/bench.md
```diff
@@ -0,0 +1,158 @@
...
+  --phy [1m|2m|coded]             PHY to use
+  --help                          Show this message and exit.
+```
+
+
+To test once device against another, one of the two devices must be running 
+the ``peripheral`` command and the other the ``central`` command. The device
+running the ``peripheral`` command will accept connections from the device
+running the ``central`` command.
+When using Bluetooth LE (all modes except for ``rfcomm-server`` and ``rfcomm`` client),
```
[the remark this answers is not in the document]
barbibulle: Fixed.

uael:
LGTM :+1:

apps/bench.py
uael: I just thought about that but we call it spawn instead of run?

barbibulle: Yes, good point, `spawn` is a better name for it.

PR #148, opened 2023-03-21
Add ASHA audio status point

yuyangh:


bumble/profiles/asha_service.py
```diff
@@ -81,9 +82,13 @@ def on_audio_control_point_write(_connection, value):
             elif opcode == AshaService.OPCODE_STATUS:
                 logger.info(f'### STATUS: connected={value[1]}')
 
-            # TODO Respond with a status
-            # asyncio.create_task(device.notify_subscribers(audio_status_characteristic,
-            # force=True))
+            # OPCODE_STATUS does not need audio status point update
+            if opcode != AshaService.OPCODE_STATUS:
+                asyncio.create_task(
```
barbibulle, line 87: As pointed out in PR #147 by @uael ,  there's a risk that the task may be garbage-collected when you don't keep a reference to it. You could use `AsyncRunner.run()` instead of `asyncio.create_task()` (when PR #147 is merged) to get around the problem.

barbibulle (approved):

bumble/profiles/asha_service.py
uael: * `AsynRunner.spawn()`

PR #150, opened 2023-03-22
Support BR/EDR role switch & change events

zxzxwu:
* Add role switch handler in host, device, device connection.
* There is a quirk that role change events may be received before HCI_Connection_Complete, especially on the responder side, or role switch is performed during connection establishment

Note: I didn't add any test because BR/EDR local test framework is not ready: https://github.com/google/bumble/pull/144

### DIFF
--- a/bumble/host.py
+++ b/bumble/host.py
@@ -603,7 +603,7 @@ class Host(AbortableEventEmitter):
                 BT_BR_EDR_TRANSPORT,
                 event.bd_addr,
                 None,
-                role,
+                BT_CENTRAL_ROLE,
                 None,
             )
         else:
@@ -744,12 +744,13 @@ class Host(AbortableEventEmitter):
                 event.bd_addr, BT_BR_EDR_TRANSPORT
             ):
                 connection.role = event.new_role
-             self.emit('role_change', event.bd_addr, event.new_role)
+            self.emit('role_change', event.bd_addr, event.new_role)
         else:
             logger.debug(
                 f'role change for {event.bd_addr} failed: '
                 f'{HCI_Constant.error_name(event.status)}'
             )
+            self.emit('role_change_failure', event.bd_addr, event.status)
 
     def on_hci_le_data_length_change_event(self, event):
         self.emit(
--- a/bumble/host.py
+++ b/bumble/host.py
@@ -24,7 +24,10 @@ from bumble.colors import color
 from bumble.l2cap import L2CAP_PDU
 from bumble.snoop import Snooper
 
+from typing import Optional
+
 from .hci import (
+    Address,
     HCI_ACL_DATA_PACKET,
     HCI_COMMAND_COMPLETE_EVENT,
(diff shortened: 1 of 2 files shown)


bumble/host.py
```diff
@@ -563,13 +585,19 @@ def on_hci_connection_complete_event(self, event):
                 f'### BR/EDR CONNECTION: [0x{event.connection_handle:04X}] '
                 f'{event.bd_addr}'
             )
+            # For connection initiator, role change may be issued if the remote device tries to switch role.
+            # For connection resopnder, role change is always issued.
```
barbibulle, line 589: nit: resopnder  -> responder

bumble/host.py
```diff
@@ -563,13 +585,19 @@ def on_hci_connection_complete_event(self, event):
...
+            if role := self.pending_connection_roles.get(event.bd_addr):
```
barbibulle, line 590: This should probably test for `is not None`, because if `pending_connection_roles[event.bd_addr]` is `BT_CENTRAL_ROLE`, (which has value 0), this will not delete the entry.

barbibulle (approved):

bumble/host.py
zxzxwu:
```
            if event.bd_addr in self.pending_connection_roles:
                role = self.pending_connection_roles[event.bd_addr]
                del self.pending_connection_roles[event.bd_addr]
```

bumble/host.py
```diff
@@ -135,13 +138,32 @@ def __init__(self, controller_source=None, controller_sink=None):
...
         self.snooper = None
+        self.pending_connection_roles = {}
 
         # Connect to the source and sink if specified
         if controller_source:
             controller_source.set_packet_sink(self)
         if controller_sink:
             self.set_packet_sink(controller_sink)
 
+    def find_connection_by_bd_addr(
```
uael, line 149: I suggest to move this logic to device instead, and to use the existing incomplete connections in `Device.pending_connections`

zxzxwu: That's much better than the original one. Thanks!

bumble/controller.py
```diff
@@ -47,6 +47,7 @@
     HCI_LE_Read_Remote_Features_Complete_Event,
     HCI_Number_Of_Completed_Packets_Event,
     HCI_Packet,
+    HCI_Role_Change_Event,
```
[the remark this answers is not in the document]
zxzxwu: Done

bumble/host.py
```diff
@@ -582,7 +603,7 @@ def on_hci_connection_complete_event(self, event):
                 BT_BR_EDR_TRANSPORT,
                 event.bd_addr,
                 None,
-                BT_CENTRAL_ROLE,
+                BT_BR_EDR_TRANSPORT,
```
uael, line 606: Are you sure about this line ?

zxzxwu: Thanks!

uael (approved):

PR #152, opened 2023-03-23
Add G722 decoder with pure python implementation

duohoo:
Add G722 decoder which will be used in ASHA media stream decoding stuffs.


bumble/decoder.py
```diff
@@ -0,0 +1,428 @@
...
+        self._band[1].det = 8
+
+    def decode_frame(self, encoded_data) -> bytearray:
+        result_array = bytearray(len(encoded_data) * 4)
+        self.g722_decode(result_array, encoded_data)
+        return result_array
+
+    def g722_decode(self, result_array, encoded_data) -> int:
+        """Decode the data frame using g722 decoder."""
+        rlow = 0
```
barbibulle, line 175: `rlow`, `rhigh`, `xout1`, `xout2`, `code` and `j` are not used before their first assignment, so it isn't necessary to initialize them here.

bumble/decoder.py
```diff
@@ -0,0 +1,428 @@
...
+        rhigh = 0
+        xout1 = 0
+        xout2 = 0
+        code = 0
+        result_length = 0
+        j = 0
+
+        for j in range(0, len(encoded_data)):
```
barbibulle, line 183: since the value of `j` isn't used other than to access elements of `encoded_data`, this could be simplified as: `for code in encoded_data`

bumble/decoder.py
```diff
@@ -0,0 +1,428 @@
...
+            code = encoded_data[j]
+            higher_bits = (code >> 6) & 0x03
+            lower_bits = code & 0x3F
+
+            rlow = self.lower_sub_band_decoder(lower_bits)
+            rhigh = self.higher_sub_band_decoder(higher_bits)
+
+            # Apply the receive QMF
+            for i in range(22):
```
barbibulle, line 192: it think it might be more efficient to do a block shift here rather than an element-wise shift. Something like `x[:22] = x[2:]`

bumble/decoder.py
```diff
@@ -0,0 +1,428 @@
...
+                self._x[i] = self._x[i + 2]
+
+            self._x[22] = rlow + rhigh
+            self._x[23] = rlow - rhigh
+            xout1 = 0
+            xout2 = 0
+
+            for i in range(12):
+                xout2 += self._x[2 * i] * QMF_COEFFS[i]
```
barbibulle, line 201: suggest: this could be easier to understand as a sum if written like: `sum(self._x[2 *i ] * QMF_COEFFS[i] for i in range(12))`. (and you could skip the `xout1 = 0` initialization.

barbibulle (approved):

bumble/decoder.py
duohoo: removed, thanks!

bumble/decoder.py
duohoo: updated, thanks

bumble/decoder.py
duohoo: updated, thanks

bumble/decoder.py
duohoo: it looks more clear after using `sum` here, thanks for the suggestions!

PR #153, opened 2023-03-23
Add 1 bug fix and a few features in bumble

benquike:
- bug fix:  [Fix a bug in HCI_Object.dict_from_bytes](https://github.com/google/bumble/commit/d55d53201ae65726ac0a76e7cda9d041bdd3e310)

- features
    -  [Allow Device.enable_classic to be configurable](https://github.com/google/bumble/commit/5b7966da54b93298b11e3945a6b16834399fec63)
    - [Add implemenetation of HCI_PIN_Code_Request_Reply_Command](https://github.com/google/bumble/commit/0e680e5cd465702f7abcd153b5bbab0066ee24f2)
    - [Add on_pin_code_request to support legacy BT classic pairing](https://github.com/google/bumble/commit/1036accba1fc857bba7c63efe53ebf5879b5e9b0)

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -94,6 +94,8 @@ from .hci import (
     HCI_LE_Set_Scan_Enable_Command,
     HCI_LE_Set_Scan_Parameters_Command,
     HCI_LE_Set_Scan_Response_Data_Command,
+    HCI_PIN_Code_Request_Reply_Command,
+    HCI_PIN_Code_Request_Negative_Reply_Command,
     HCI_Read_BD_ADDR_Command,
     HCI_Read_RSSI_Command,
     HCI_Reject_Connection_Request_Command,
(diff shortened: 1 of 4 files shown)


bumble/hci.py
```diff
@@ -1491,7 +1491,7 @@ def dict_from_bytes(data, offset, fields):
             elif field_type == -2:
                 # 16-bit signed
                 field_value = struct.unpack_from('<h', data, offset)[0]
-                offset += 1
+                offset += 2
```
barbibulle, line 1494: good catch! I dont think any of the commands use that field type yet, but that's a bug nonetheless.

bumble/host.py
```diff
@@ -797,7 +798,9 @@ def on_hci_pin_code_request_event(self, event):
         # For now, just refuse all requests
         # TODO: delegate the decision
         self.send_command_sync(
-            HCI_PIN_Code_Request_Negative_Reply_Command(bd_addr=event.bd_addr)
+            HCI_PIN_Code_Request_Reply_Command(bd_addr = event.bd_addr,
```
barbibulle, line 801: this hard-coded response doesn't seem to match the intent of responding to these events from device.py as supported in the other commit. Is that the intent here?

uael: It should be emitted to device I guess

uael (changes requested):

bumble/host.py
benquike: Updated, can you have another look?

bumble/device.py
```diff
@@ -2762,6 +2769,49 @@ async def get_number():
...
+            smp.SMP_KEYBOARD_DISPLAY_IO_CAPABILITY,
+        )
+
+        # respond the pin code
+        if can_input:
+            async def get_pin_code():
+                pin_code = await connection.abort_on(
+                    'disconnection', pairing_config.delegate.get_number()
+                )
+                pin_code = bytes(str(pin_code).zfill(6))
```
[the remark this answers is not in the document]
benquike: updated, can you have another look?

uael (changes requested):

uael (approved):

PR #158, opened 2023-03-24
Fix HCI_PIN_Code_Reply_Command 

benquike:
* fix HCI_PIN_Code_Reply_Command  implementation
* fix on Device.on_pin_code_request

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -2804,23 +2804,13 @@ class Device(CompositeEventEmitter):
 
             async def get_pin_code():
                 pin_code = await connection.abort_on(
-                    'disconnection', pairing_config.delegate.get_number()
+                    'disconnection', pairing_config.delegate.get_string(16)
                 )
 
                 if pin_code is not None:
-                    # TODO: make pin code input more flexible,
-                    # so far only numbers are allowed
-                    pin_code = bytes(str(pin_code), encoding='utf8')
+                    pin_code = bytes(pin_code, encoding='utf-8')
                     pin_code_len = len(pin_code)
-                    if pin_code_len > 16:
-                        logger.warning(
-                            'pin_code is longer than 16 bytes, truncated to 16'
-                        )
-                        pin_code_len = 16
-                        # the pin_code will be truncated according to pin_code_len
-                        # when converted to byte array in HCI_Object, so we
-                        # are not doing truncation here
-
+                    assert 0 < pin_code_len <= 16, "pin_code should be 1-16 bytes"
                     await self.host.send_command(
                         HCI_PIN_Code_Request_Reply_Command(
                             bd_addr=connection.peer_address,
@@ -2829,6 +2819,7 @@ class Device(CompositeEventEmitter):
                         )
                     )
                 else:
+                    logger.debug("delegate.get_string() returned None")
                     await self.host.send_command(
                         HCI_PIN_Code_Request_Negative_Reply_Command(
                             bd_addr=connection.peer_address
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -2808,11 +2808,23 @@ class Device(CompositeEventEmitter):
                 )
 
                 if pin_code is not None:
-                    pin_code = bytes(str(pin_code).zfill(6))
+                    # TODO: make pin code input more flexible,
+                    # so far only numbers are allowed
+                    pin_code = bytes(str(pin_code), encoding='utf8')
+                    pin_code_len = len(pin_code)
+                    if pin_code_len > 16:
+                        logger.warning(
+                            'pin_code is longer than 16 bytes, truncated to 16'
+                        )
+                        pin_code_len = 16
+                        # the pin_code will be truncated according to pin_code_len
+                        # when converted to byte array in HCI_Object, so we
+                        # are not doing truncation here
+
                     await self.host.send_command(
                         HCI_PIN_Code_Request_Reply_Command(
                             bd_addr=connection.peer_address,
-                            pin_code_length=len(pin_code),
+                            pin_code_length=pin_code_len,
                             pin_code=pin_code,
                         )
                     )
--- a/bumble/smp.py
+++ b/bumble/smp.py
@@ -525,6 +525,12 @@ class PairingDelegate:
     async def get_number(self) -> int:
         return 0
 
+    async def get_string(self, max_length) -> Optional[str]:
+        '''
+        Returns a string whose utf-8 encoding is up to max_length bytes.
+        '''
+        return None
+
     # pylint: disable-next=unused-argument
     async def display_number(self, number: int, digits: int) -> None:
         pass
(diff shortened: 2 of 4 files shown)


bumble/device.py
```diff
@@ -2808,11 +2808,23 @@ async def get_pin_code():
...
+                    pin_code = bytes(str(pin_code), encoding='utf8')
+                    pin_code_len = len(pin_code)
+                    if pin_code_len >= 0x10:
+                        logger.warning(
+                            'pin_code is longer than 16 bytes, truncated to 16'
+                        )
+                        pin_code_len = 0x10
+                        pin_code = pin_code[:16]
+                    else:
+                        pin_code = pin_code + b'0' * (16 - pin_code_len)
```
uael, line 2822: Shouldn't it need to be a prefix padding ?
`pin_code = bytes(str(pin_code).zfill(16))` should do the trick.
I think 6 bytes is enough

tests/hci_test.py
```diff
@@ -215,12 +215,13 @@ def test_HCI_Command():
 
 # -----------------------------------------------------------------------------
 def test_HCI_PIN_Code_Request_Reply_Command():
+    code = b'1234'
```
uael, line 218: should be `b'001234'`

bumble/device.py
barbibulle: The way I read the spec, the PIN code is a string, up to 16 bytes. So 0-padding past PIN_Code_Length should be right.

barbibulle, line 2822: You shouldn't need to manually pad the field here, the HCI object serializer automatically truncates/pads byte array fields when they are fixed length.

bumble/device.py
```diff
@@ -2808,11 +2808,23 @@ async def get_pin_code():
                 )
 
                 if pin_code is not None:
-                    pin_code = bytes(str(pin_code).zfill(6))
+                    # TODO: make pin code input more flexible,
```
barbibulle, line 2811: The `PairingDelegate` class should probably be enhanced with a `get_string` method (something like: `async def get_string(self, max_length)`

bumble/device.py
benquike: Whether padding before the pin code or after the pin code is something I am not sure.

bumble/device.py
benquike: Yes, that's what  I proposed in discussions with @uael. I will update it in another PR

bumble/device.py
```diff
@@ -2808,11 +2808,23 @@ async def get_pin_code():
...
+                    # so far only numbers are allowed
+                    pin_code = bytes(str(pin_code), encoding='utf8')
+                    pin_code_len = len(pin_code)
+                    if pin_code_len >= 16:
```
barbibulle, line 2815: should be > 16 I think.

tests/hci_test.py
benquike: Per our discussion, updated with padding zero bytes after the pin code

bumble/device.py
benquike: Updated.

bumble/device.py
benquike: Updated.

bumble/device.py
uael: Would you be able to fix it on this one to avoid the TODO ? Thanks!

bumble/device.py
benquike: To summarize, we should pad b'0x00's, but as padding is done for us, in final version of this change, we do not pad.

bumble/device.py
benquike: Sure, I will

uael: Sg!

bumble/smp.py
```diff
@@ -525,6 +525,9 @@ async def compare_numbers(self, number: int, digits: int) -> bool:
     async def get_number(self) -> int:
         return 0
 
+    async def get_string(self) -> str:
```
barbibulle, line 528: I suggest adding a `max_length` argument to let the delegate know what's the maximum number of encoded bytes that are expected. This can be useful for a user interface to limit the text input field.

barbibulle (approved):

bumble/smp.py
benquike: I will add it

bumble/smp.py
```diff
@@ -525,6 +525,9 @@ async def compare_numbers(self, number: int, digits: int) -> bool:
     async def get_number(self) -> int:
         return 0
 
+    async def get_string(self, max_length) -> str:
```
uael, line 528: Can you use `Optional[str]` instead ?

bumble/device.py
```diff
@@ -2804,24 +2804,32 @@ def on_pin_code_request(self, connection):
...
                 if pin_code is not None:
-                    pin_code = bytes(str(pin_code).zfill(6))
-                    await self.host.send_command(
-                        HCI_PIN_Code_Request_Reply_Command(
-                            bd_addr=connection.peer_address,
-                            pin_code_length=len(pin_code),
-                            pin_code=pin_code,
+                    pin_code = bytes(pin_code, encoding='utf8')
+                    pin_code_len = len(pin_code)
+                    if 0 < pin_code_len <= 16:
```
uael, line 2813: Since `max_length` was already supplied to `get_string` I would only assert here to make sure the API contract has been respected

uael (changes requested):

bumble/smp.py
```diff
@@ -525,6 +525,13 @@ async def compare_numbers(self, number: int, digits: int) -> bool:
     async def get_number(self) -> int:
         return 0
 
+    async def get_string(self, min_length=0, max_length=-1) -> Optional[str]:
```
uael, line 528: I'm not sure about `min_length`, @barbibulle what do you think ?

bumble/device.py
```diff
@@ -2804,24 +2804,30 @@ def on_pin_code_request(self, connection):
...
+                    assert 0 < pin_code_len <= 16, "pin_code should be 1-16 bytes"
                     await self.host.send_command(
                         HCI_PIN_Code_Request_Reply_Command(
                             bd_addr=connection.peer_address,
-                            pin_code_length=len(pin_code),
+                            pin_code_length=pin_code_len,
                             pin_code=pin_code,
                         )
                     )
+                    return
```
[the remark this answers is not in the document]
uael: Since the HCI packets are already logged I would keep the logs close to what we observe for a passkey request, as you want :smile:

bumble/smp.py
uael: I suggest to:
- remove `min_length` since `None` is available
- remove default value for `max_length`
- return `None` by default
As follow:
```python
async def get_string(self, max_length) -> Optional[str]:
    '''Returns a string of max_length bytes.'''
    return None
```

uael (changes requested):

bumble/smp.py
barbibulle: To be precise, the comment should probably say: """Returns a string whose utf-8 encoding is up to max_length bytes."""

uael (approved):

PR #160, opened 2023-03-25
Add some missing types to apps/console.py, bumble/gatt_client.py

AlanRosenthal:
Added via code inspection (not via a tool like pytype)

### DIFF
--- a/bumble/gatt_client.py
+++ b/bumble/gatt_client.py
@@ -23,9 +23,11 @@
 # -----------------------------------------------------------------------------
 # Imports
 # -----------------------------------------------------------------------------
+from __future__ import annotations
 import asyncio
 import logging
 import struct
+from typing import List, Optional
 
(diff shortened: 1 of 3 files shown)


bumble/gatt_client.py
```diff
@@ -332,8 +344,9 @@ async def discover_services(self, uuids=None):
                         '!!! unexpected error while discovering services: '
                         f'{HCI_Constant.error_name(response.error_code)}'
                     )
-                    # TODO raise appropriate exception
-                    return
+                    raise Exception(
```
barbibulle, line 347: suggest: create a more specific exception class. Maybe something like `ATT_Error(response.error_code)`, like we do for `HCI_Error` (i.e with an error code, error namespace and error name).

barbibulle (approved):

barbibulle (approved):

PR #161, opened 2023-03-27
smp: fix `PairingDelegate.get_number` return type

uael:
This function can return `None` to indicate a negative reply, update the type hint accordingly.

### DIFF
--- a/bumble/smp.py
+++ b/bumble/smp.py
@@ -522,7 +522,11 @@ class PairingDelegate:
     async def compare_numbers(self, number: int, digits: int) -> bool:
         return True
 
-    async def get_number(self) -> int:
+    async def get_number(self) -> Optional[int]:
+        '''
+        Returns an optional number as an answer to a passkey request.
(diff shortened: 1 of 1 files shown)


bumble/smp.py
```diff
@@ -522,7 +522,7 @@ async def confirm(self) -> bool:
     async def compare_numbers(self, number: int, digits: int) -> bool:
         return True
 
-    async def get_number(self) -> int:
+    async def get_number(self) -> Optional[int]:
```
barbibulle, line 525: For sure. Could you add a comment to specify that returning None means that the delegate was unable to return a valid number?

barbibulle (approved):

bumble/smp.py
uael: Done! But I'm not sure it's correctly expressed.

barbibulle: lgtm

barbibulle (approved):

PR #163, opened 2023-03-28
Add `show local-values`

AlanRosenthal:
This PR adds a way to display the local gatt characteristics/descriptors values

If no connections, it shows the value of every characteristic/descriptor. When there's a connection, it shows the value for each specific connection - CCCDs are connection specific

This screen auto-updates every second

### DIFF
--- a/bumble/gatt_server.py
+++ b/bumble/gatt_server.py
@@ -27,7 +27,7 @@ import asyncio
 import logging
 from collections import defaultdict
 import struct
-from typing import List, Tuple, Optional
+from typing import List, Tuple, Optional, TypeVar, Type
 from pyee import EventEmitter
 
 from .colors import color
@@ -134,6 +134,21 @@ class Server(EventEmitter):
                 self.attributes_by_handle[handle] = attribute
                 return attribute
         return None
+
+    AttributeGroupType = TypeVar('AttributeGroupType', Service, Characteristic)
+
+    def get_attribute_group(
+        self, handle: int, group_type: Type[AttributeGroupType]
+    ) -> Optional[AttributeGroupType]:
+        return next(
+            (
+                attribute
+                for attribute in self.attributes
+                if isinstance(attribute, group_type)
+                and attribute.handle <= handle <= attribute.end_group_handle
+            ),
+            None,
+        )
 
     def get_service_attribute(self, service_uuid: UUID) -> Optional[Service]:
         return next(
--- a/apps/console.py
+++ b/apps/console.py
@@ -28,6 +28,7 @@ from typing import Optional
 from collections import OrderedDict
 
 import click
+from prettytable import PrettyTable
 
 from prompt_toolkit import Application
 from prompt_toolkit.history import FileHistory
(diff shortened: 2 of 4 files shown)


bumble/gatt_server.py
```diff
@@ -135,6 +135,20 @@ def get_attribute(self, handle):
                 return attribute
         return None
 
+    def get_service_from_handle_in_range(self, handle: int) -> Optional[Service]:
```
barbibulle, line 138: You could unify this method and the next into a single generic one, since they only differ by a type.
One option may be to call it `get_attribute_group(self, handle: int, group_type: type)` --> Get the attribute of a given type that starts a group that contains a given handle. 
You then call `gatt_server.get_attribute(some_attribute, Service)` or `gatt_server.get_attribute(some_attribute, Characteristic)`. The type hinting system can let you express such generic methods so that it knows that if you pass a type T as the `group_type` argument, the method returns that same type T.

AlanRosenthal: I updated the code to do this, but I couldn't figure out how to get mypy to pass when using the generic methods

barbibulle: I think this can work as:
```
AttributeGroupType = TypeVar('GroupType', Service, Characteristic)
def get_attribute_group(self, handle: int, group_type: Type[AttributeGroupType]) -> AttributeGroupType
```

barbibulle (approved):

bumble/gatt_server.py
AlanRosenthal: this is the issue I'm running into:

```patch
diff --git a/bumble/gatt_server.py b/bumble/gatt_server.py
index e9e667f..538b021 100644
--- a/bumble/gatt_server.py
+++ b/bumble/gatt_server.py
@@ -27,7 +27,7 @@ import asyncio
 import logging
 from collections import defaultdict
 import struct
-from typing import List, Tuple, Optional, Union
+from typing import List, Tuple, Optional, Union, TypeVar
 from pyee import EventEmitter
 
 from .colors import color
@@ -135,9 +135,10 @@ class Server(EventEmitter):
                 return attribute
         return None
 
+    AttributeGroupType = TypeVar('AttributeGroupType', Service, Characteristic)
     def get_attribute_group(
-        self, handle: int, group_type
-    ) -> Optional[Union[Service, Characteristic]]:
+        self, handle: int, group_type: AttributeGroupType
+    ) -> Optional[AttributeGroupType]:
         return next(
             (
                 attribute
```

```
❯ invoke project.check-types
bumble/gatt_server.py:144: error: Generator has incompatible item type "Attribute"; expected "Optional[Service]"  [misc]
bumble/gatt_server.py:144: error: Generator has incompatible item type "Attribute"; expected "Optional[Characteristic]"  [misc]
bumble/gatt_server.py:146: error: Argument 2 to "isinstance" has incompatible type "Service"; expected "_ClassInfo"  [arg-type]
bumble/gatt_server.py:146: error: Argument 2 to "isinstance" has incompatible type "Characteristic"; expected "_ClassInfo"  [arg-type]
```

barbibulle: The type for `group_type` should be `Type[AttributeGroupType]`.

AlanRosenthal: ah thanks. found the doc section on this: https://docs.python.org/3/library/typing.html#typing.Type
it works now!

PR #164, opened 2023-03-28
Add `local-write` to bumble-console

AlanRosenthal:
Add a command to update the local gatt server, and notify/indicate subscribes (if any)

### DIFF
--- a/bumble/gatt.py
+++ b/bumble/gatt.py
@@ -28,7 +28,7 @@ import enum
 import functools
 import logging
 import struct
-from typing import Optional, Sequence
+from typing import Optional, Sequence, List, Any, Iterable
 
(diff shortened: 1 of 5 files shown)


bumble/gatt.py
```diff
@@ -325,6 +327,12 @@ def get_descriptor(self, descriptor_type):
 
         return None
 
+    def has_properties(self, properties: List[Any]):
```
barbibulle, line 330: The type could me more flexible. Accepting an `Iterable` instead of a`List` allows the caller to use tuples or other iterables. Also, shouldn't it be an iterable of `int` rather than `Any` here?

barbibulle (approved):

PR #166, opened 2023-03-30
att: fixed use of unknown attribute

uael:


AlanRosenthal (approved):
is there a way to catch this with some type of static analysis, pylint, mypy, etc?

uael:
I guess we should enable strict typing but it's not that easy with the current packet pack/unpack decorators since types are generated at runtime

PR #167, opened 2023-03-30
Create Characteristic.Property

AlanRosenthal:
Move all Characteristic properties into its own `enum.IntFlag` class

### DIFF
--- a/bumble/gatt.py
+++ b/bumble/gatt.py
@@ -28,7 +28,7 @@ import enum
 import functools
 import logging
 import struct
-from typing import Optional, Sequence, List, Any, Iterable
+from typing import Optional, Sequence, List
 
 from .colors import color
 from .core import UUID, get_dict_key_by_value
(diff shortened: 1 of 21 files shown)


bumble/gatt.py
```diff
@@ -259,63 +259,57 @@ class Characteristic(Attribute):
     See Vol 3, Part G - 3.3 CHARACTERISTIC DEFINITION
     '''
 
-    # Property flags
-    BROADCAST = 0x01
```
barbibulle, line 263: I recommend adding a backward-compatible "hoisting" of the new names so that they can still be used via the older names.
Something like:
```
# Deprecated - only for backward compatibility. Use the `Properties` enum instead.
BROADCAST = Properties.BROADCAST
READ = Properties.READ
...
```
Unfortunately  I don't think there's a way to annotation class properties with `@deprecated`

bumble/gatt.py
```diff
@@ -259,63 +259,57 @@ class Characteristic(Attribute):
...
-    def string_to_properties(properties_str: str):
-        return functools.reduce(
-            lambda x, y: x | get_dict_key_by_value(Characteristic.PROPERTY_NAMES, y),
-            properties_str.split(","),
-            0,
-        )
+    uuid: UUID
+    properties: Characteristic.Property
+
+    class Property(enum.IntFlag):
```
barbibulle, line 265: Since this represents a combination of properties, I'd recommend naming the class `Properties`.

barbibulle (approved):

PR #169, opened 2023-03-31
Add `show remote-values`

AlanRosenthal:
`gatt_client` caches values read/notifications/indications and displays the most recent value to the user


AlanRosenthal:
[Screencast from 2023-03-31 17-11-19.webm](https://user-images.githubusercontent.com/1288897/229234926-93626d85-edc7-4996-ab9c-6c495ea33f62.webm)

bumble/gatt_client.py
```diff
@@ -942,6 +948,11 @@ def on_att_handle_value_notification(self, notification):
         )
         if not subscribers:
             logger.warning('!!! received notification with no subscriber')
+
+        self.cached_values[notification.attribute_handle] = (
```
barbibulle, line 952: The same caching logic appears 3 times. This could be factored out in a `cache_value(handle, value)` method.

bumble/gatt_client.py
```diff
@@ -961,3 +977,32 @@ def on_att_handle_value_indication(self, indication):
 
         # Confirm that we received the indication
         self.send_confirmation(ATT_Handle_Value_Confirmation())
+
+    def get_attributes(
```
barbibulle, line 981: I'd suggest a more specific name (looking at the name and args, it looks like this would maybe look up an attribute by handle) . Maybe something like `get_attribute_grouping`.
Also, maybe move up this method earlier in the file, around the definition for `get_characteristics_by_uuid`

barbibulle (approved):

PR #171, opened 2023-04-03
use device public or static address for keystore namespace

barbibulle:
Instead of just using the device static random address, use the most likely identifier for the controller as the namespace for keystores:
  * Try the public address first if the controller has one
  * Second choice is the static random address assigned to the controller
  * Otherwise: default namespace.


uael (approved):
lgtm :+1:

PR #173, opened 2023-04-07
SMP: Determine initiator by direction instead of role

zxzxwu:
Though in the spec this is not allowed, but in some ambiguous cases such as SMP over BR/EDR or we want to test how remote devices handle invalid pairing requests, we still need to allow this behavior, with some logs to let users know it's invalid.

### DIFF
--- a/bumble/smp.py
+++ b/bumble/smp.py
@@ -645,7 +645,7 @@ class Session:
         },
     }
 
-    def __init__(self, manager, connection, pairing_config):
+    def __init__(self, manager, connection, pairing_config, is_initiator):
         self.manager = manager
         self.connection = connection
         self.preq = None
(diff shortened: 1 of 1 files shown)


uael (approved):
LGTM 👍

bumble/smp.py
```diff
@@ -1680,6 +1680,8 @@ def send_command(self, connection, command):
     def on_smp_pdu(self, connection, pdu):
         # Look for a session with this connection, and create one if none exists
         if not (session := self.sessions.get(connection.handle)):
+            if connection.role == BT_CENTRAL_ROLE:
+                logger.warn('Remote starts pairing as Peripheral!')
```
barbibulle, line 1684: `.warn` is deprecated, use `.warning` instead.

bumble/smp.py
```diff
@@ -1709,10 +1711,12 @@ def ecc_key(self):
 
     async def pair(self, connection):
         # TODO: check if there's already a session for this connection
+        if connection.role != BT_CENTRAL_ROLE:
+            logger.warn('Start pairing as Peripheral!')
```
barbibulle, line 1715: `.warning`

barbibulle (approved):

bumble/smp.py
zxzxwu: Done

PR #174, opened 2023-04-07
fix role state for classic connections

barbibulle:
Somehow the connection role wasn't handled quite right for Classic connections (because we don't know the role at the time of connection, the role is temporarily unknown). This caused CTKD to fail, because over a Classic connection, as a peripheral, the SMP would not distribute keys in response to an SMP pairing request since it incorrectly thought it was the central.
This PR removes the `role` attribute in the host-level connection object (it's actually not needed there).
Also, this removes the `resolved_peer_address` property for Classic connections, since there's no concept of resolvable keys there, that property isn't relevant.
Finally, there's limited support for disabling SMP over Classic connection, for when a device doesn't want to participate in Classic CTKD.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -595,7 +595,7 @@ class Connection(CompositeEventEmitter):
 
     # [Classic only]
     @classmethod
-    def incomplete(cls, device, peer_address):
+    def incomplete(cls, device, peer_address, role):
         """
         Instantiate an incomplete connection (ie. one waiting for a HCI Connection
         Complete event).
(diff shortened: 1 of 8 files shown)


bumble/device.py
```diff
@@ -1942,7 +1952,7 @@ def on_connection_failure(error):
 
         # Save pending connection
         self.pending_connections[peer_address] = Connection.incomplete(
-            self, peer_address
+            self, peer_address, BT_PERIPHERAL_ROLE
```
uael, line 1955: nit: Add a comment:
```
# Even if we requested a role switch in the HCI_Accept_Connection_Request command,
# This connection is still considered peripheral until an eventual role change event.
```

uael (approved):

bumble/device.py
barbibulle: done.

PR #175, opened 2023-04-08
Fix a bug in Device.on_pin_code_request

benquike:
SMP_KEYBOARD_DISPLAY_IO_CAPABILITY is not
present in Bluetooth classic.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -596,7 +596,7 @@ class Connection(CompositeEventEmitter):
 
     # [Classic only]
     @classmethod
-    def incomplete(cls, device, peer_address):
+    def incomplete(cls, device, peer_address, role):
         """
         Instantiate an incomplete connection (ie. one waiting for a HCI Connection
         Complete event).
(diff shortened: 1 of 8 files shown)


bumble/device.py
```diff
@@ -2844,9 +2844,8 @@ def on_pin_code_request(self, connection):
         # Ask what the pairing config should be for this connection
         pairing_config = self.pairing_config_factory(connection)
 
-        can_input = pairing_config.delegate.io_capability in (
-            smp.SMP_KEYBOARD_ONLY_IO_CAPABILITY,
-            smp.SMP_KEYBOARD_DISPLAY_IO_CAPABILITY,
+        can_input = (
```
barbibulle, line 2847: The API contract is a bit different (but isn't well documented): the intent si the the delegate chooses its `io_capability` from one of the 5 `SMP_XXX` constants (there should be an abstract class to represent the abstract I/O concepts, but since the constants from `SMP` are the most comprehensive of the two (between Classic and LE), that's what's currently used. Then, when in Classic mode, there's a mapping between the `SMP_XXX` constants to the most logical classic constant that would correspond (see line 2707 for the entire mapping). In the case of a PIN code request, `can_input` should be true if the delegate indicates that it has a keyboard, so both `SMP_KEYBOARD_ONLY_IO_CAPABILITY` and `SMP_KEYBOARD_DISPLAY_IO_CAPABILITY` would work.

benquike: I agree that using some abstract symbols for the io capabilities makes more sense.
But using SMP_XXXX symbols in methods related to classic pairing makes it very confusing as there is no such thing as SMP at all in bt classic.

barbibulle: I'll push out a PR later today where the io capabilities of the pairing delegate will use abstract values rather than directly the constants from the SMP protocol.

benquike: I also suggest moving pairing delegate and pairing config out of SMP, as we need them in classic pairing as well.

barbibulle: Yes, for sure. See PR #177 where that's done.

PR #177, opened 2023-04-12
refactor PairingDelegate

barbibulle:
This moves the PairingDelegate to its own file. It is thus clearer that the PairingDelegate interface can serve for both Classic and LE pairing.
Also included is a fix for one of the Classic pairing modes, where, according to the spec, the behavior of the responder depends on the `io_capability` of the initiator, so that value is now stored in the Connection object.
Tested against Android and macOS (with all 4 Classic `io_capability` types, both as an initiator and responder, even though the Android and macOS sides are always `DisplayYesNo`)

### DIFF
--- a/bumble/pairing.py
+++ b/bumble/pairing.py
@@ -1,0 +1,184 @@
+# Copyright 2021-2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# -----------------------------------------------------------------------------
+# Imports
+# -----------------------------------------------------------------------------
+import enum
+from typing import Optional, Tuple
+
+from .hci import (
+    HCI_NO_INPUT_NO_OUTPUT_IO_CAPABILITY,
+    HCI_DISPLAY_ONLY_IO_CAPABILITY,
+    HCI_DISPLAY_YES_NO_IO_CAPABILITY,
+    HCI_KEYBOARD_ONLY_IO_CAPABILITY,
+)
+from .smp import (
+    SMP_NO_INPUT_NO_OUTPUT_IO_CAPABILITY,
+    SMP_KEYBOARD_ONLY_IO_CAPABILITY,
+    SMP_DISPLAY_ONLY_IO_CAPABILITY,
+    SMP_DISPLAY_YES_NO_IO_CAPABILITY,
+    SMP_KEYBOARD_DISPLAY_IO_CAPABILITY,
+    SMP_ENC_KEY_DISTRIBUTION_FLAG,
+    SMP_ID_KEY_DISTRIBUTION_FLAG,
+    SMP_SIGN_KEY_DISTRIBUTION_FLAG,
+    SMP_LINK_KEY_DISTRIBUTION_FLAG,
+)
+
+
+# -----------------------------------------------------------------------------
+class PairingDelegate:
+    """Abstract base class for Pairing Delegates."""
+
+    # I/O Capabilities.
+    # These are defined abstractly, and can be mapped to specific Classic pairing
+    # and/or SMP constants.
+    class IoCapability(enum.IntEnum):
+        NO_OUTPUT_NO_INPUT = SMP_NO_INPUT_NO_OUTPUT_IO_CAPABILITY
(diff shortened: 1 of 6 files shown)


bumble/pairing.py
```diff
@@ -0,0 +1,184 @@
...
+
+        # This is just a 1-1 direct mapping
+        return self.io_capability
+
+    async def accept(self) -> bool:
+        """Accept or reject a Pairing request."""
+        return True
+
+    async def confirm(self) -> bool:
+        """Respond yes or not to a Pairing confirmation question."""
```
benquike, line 117: not -> no

bumble/pairing.py
```diff
@@ -0,0 +1,184 @@
...
+        DISPLAY_OUTPUT_AND_KEYBOARD_INPUT: HCI_DISPLAY_YES_NO_IO_CAPABILITY,
+    }
+
+    io_capability: IoCapability
+    local_initiator_key_distribution: KeyDistribution
+    local_responder_key_distribution: KeyDistribution
+
+    def __init__(
+        self,
+        io_capability=NO_OUTPUT_NO_INPUT,
```
benquike, line 88: As a pairing delegate object is always associated with a device, how about adding a device field in this class?

To make the io capabilities exposed to configuration, we can add classic and le io capabilies in the device,

Then the default implementations for `classic_io_capability` and `smp_io_capability` can also modified to get the classic and le io cap from device object.

bumble/pairing.py
```diff
@@ -0,0 +1,184 @@
...
+    def classic_io_capability(self) -> int:
+        """Map the abstract I/O capability to a Classic constant."""
+
+        # pylint: disable=line-too-long
+        return self.CLASSIC_IO_CAPABILITIES_MAP.get(
+            self.io_capability, HCI_NO_INPUT_NO_OUTPUT_IO_CAPABILITY
+        )
+
+    @property
+    def smp_io_capability(self) -> int:
```
benquike, line 106: I don't know how much overlap between classic and SMP, if there is not much overlay, I would suggest the APIs for classic and smp through separate sub-objects  and move the pairing protocol specific definitions there.

Using one API for different purposes makes it very confusing.

bumble/pairing.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+# -----------------------------------------------------------------------------
+class PairingDelegate:
+    """Abstract base class for Pairing Delegates."""
+
+    # I/O Capabilities.
+    # These are defined abstractly, and can be mapped to specific Classic pairing
+    # and/or SMP constants.
+    class IoCapability(enum.IntEnum):
```
benquike, line 47: Why is this defined inside PairingDelegate?

uael: I also suggest to move `IoCapability` out of the delegate, as well as `KeyDistribution`

uael (approved):

bumble/pairing.py
uael: They shall be the same so one field is enough

bumble/pairing.py
```diff
@@ -0,0 +1,184 @@
...
+        self.sc = sc
+        self.mitm = mitm
+        self.bonding = bonding
+        self.delegate = delegate or PairingDelegate()
+
+    def __str__(self) -> str:
+        return (
+            f'PairingConfig(sc={self.sc}, '
+            f'mitm={self.mitm}, bonding={self.bonding}, '
+            f'delegate[{self.delegate.io_capability}])'
```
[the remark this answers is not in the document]
uael: I think it is :thinking:

bumble/pairing.py
benquike: Not necessarily. At least on Android, there are separate variables are used to save io caps for classic and LE.

I am not sure of the reason through.

bumble/pairing.py
barbibulle: The point of the PairingDelegate is to allow users of the API to define how they want to make choices for pairing, in a dynamic way, including having a single policy used both for Classic and LE pairing. The property methods there are used to convert abstract io capabilities into concrete ones for the specific protocols.

bumble/pairing.py
barbibulle: This enum is really only used to allow a PairingDelegate to express its io capabilities, so it isn't needed outside that interface. That's why it's an inner definition inside the PairingDelegate class.

bumble/pairing.py
```diff
@@ -0,0 +1,184 @@
...
+        )
+
+
+# -----------------------------------------------------------------------------
+class PairingConfig:
+    """Configuration for the Pairing protocol."""
+
+    def __init__(
+        self,
+        sc: bool = True,
```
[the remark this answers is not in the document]
barbibulle: The notion of `mitm` and `bonding` are general, they apply to pairing both for Classic and LE. 
The PairingConfig class offers an abstract model of a superset of what Classic and LE pairing need, allowing a single config to be used for either type of pairing, thus not requiring two configs, one for each.

uael (approved):

PR #178, opened 2023-04-19
Overall fixes and improvements 

uael:
See commits for details

### DIFF
--- a/bumble/core.py
+++ b/bumble/core.py
@@ -152,7 +152,12 @@ class UUID:
     BASE_UUID = bytes.fromhex('00001000800000805F9B34FB')[::-1]  # little-endian
     UUIDS: List[UUID] = []  # Registry of all instances created
 
-    def __init__(self, uuid_str_or_int, name=None):
+    uuid_bytes: bytes
+    name: Optional[str]
+
+    def __init__(
+        self, uuid_str_or_int: Union[str, int], name: Optional[str] = None
+    ) -> None:
         if isinstance(uuid_str_or_int, int):
             self.uuid_bytes = struct.pack('<H', uuid_str_or_int)
         else:
@@ -172,7 +177,7 @@ class UUID:
             self.uuid_bytes = bytes(reversed(bytes.fromhex(uuid_str)))
         self.name = name
 
-    def register(self):
+    def register(self) -> UUID:
         # Register this object in the class registry, and update the entry's name if
         # it wasn't set already
         for uuid in self.UUIDS:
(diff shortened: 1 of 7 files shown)


uael:
I just added more details in the commit messages :+1:

bumble/core.py
```diff
@@ -288,8 +288,19 @@ def __str__(self):
 
         return result
 
-    def __repr__(self):
-        return str(self)
+    def __repr__(self) -> str:
```
AlanRosenthal, line 291: Could you add a couple of test cases for this

uael: I'm not sure what to test here, it's copy/pasted from `to_hex_str` with slightly modifications

AlanRosenthal: I'd just do something like:
```
  def test_uuid_repr():
    assert UUID("1234").__repr__() == "1234"
    assert UUID("df5ce654-e059-11ed-b5ea-0242ac120002").__repr__() == "DF5CE654-E059-11ED-B5EA-0242AC120002"

File: bumble/core.py [hunk 2]
```diff
@@ -288,8 +288,19 @@ def __str__(self):
...
+        if len(self.uuid_bytes) == 2 or len(self.uuid_bytes) == 4:
+            return bytes(reversed(self.uuid_bytes)).hex().upper()
+
+        return '-'.join(
```
barbibulle, line 295: Since this is the same as what's used in `__str__`, consider factoring out this full-size uuid->string conversion in its own method.
Maybe all we need is to add a `separator` argument to `to_hex_string` with a default of '', then `__repr__` becomes just `self.to_hex_string(separator="-")`, and `__str__` would also call `to_hex_string(separator="-")` for 16-byte UUIDs.

bumble/keys.py
```diff
@@ -307,3 +307,30 @@ async def get(self, name: str) -> Optional[PairingKeys]:
             return None
 
         return PairingKeys.from_dict(keys)
+
+
+# -----------------------------------------------------------------------------
+class InMemoryStore(KeyStore):
```
barbibulle, line 313: suggest: naming this class just `MemoryKeyStore` (having the word `Key` in there makes it explicit what this stores, and the `In` prefix isn't really needed. This also parallels the name of `JsonKeyStore`.

bumble/device.py
```diff
@@ -2778,84 +2778,60 @@ def on_authentication_user_confirmation_request(self, connection, code):
...
         io_capability = pairing_config.delegate.classic_io_capability
 
-        # Respond
+        # See Bluetooth spec @ Vol 3, Part C 5.2.2.6
         if io_capability == HCI_DISPLAY_YES_NO_IO_CAPABILITY:
-            if connection.peer_pairing_io_capability in (
-                HCI_DISPLAY_YES_NO_IO_CAPABILITY,
-                HCI_DISPLAY_ONLY_IO_CAPABILITY,
-            ):
+
```
barbibulle, line 2783: I don't think we can remove the check for the *peer* device's capabilities here. According to Table 5.7 in Vol 3, Part C 5.2.2.6, when the peer has `NoInputNoOutput`, the responder should confirm but not show the code ("Yes/No confirmation whether to pair on device B. Device B does not show the confirmation value.")
That makes sense, because if the initiator has no input and not output, there's no point showing the code on the responder, since the initiator will auto-accept without user intervention, there's nothing to compare the code against.

bumble/device.py
```diff
@@ -2778,84 +2778,60 @@ def on_authentication_user_confirmation_request(self, connection, code):
...
+                )
 
-            AsyncRunner.spawn(connection.abort_on('disconnection', confirm()))
-            return
-
-        if io_capability == HCI_DISPLAY_ONLY_IO_CAPABILITY:
-            # Display the code to the user
-            AsyncRunner.spawn(pairing_config.delegate.display_number(code, 6))
-
-        # Automatic confirmation
```
barbibulle, line 2818: I don't think we can remove this automatic confirmation without prompt. When the responder has no input (`DisplayOnly`), we can't ask the user to confirm. The spec does say "with automatic confirmation".

bumble/device.py
uael: I double checked and yes, let me re-add the check :+1:

bumble/core.py
uael: Ack :+1:

bumble/device.py
barbibulle: I think it makes sense to call the delegate even for auto-confirm cases. Maybe the way to do it is to add a `auto=False` parameter to `confirm()` so that a "real" delegate (i.e with a user) can know that it's being called but isn't supposed to actually prompt the user. For auto-confirm flows, we'd call `confirm(auto=True)`, and just `confirm()` in the non-auto confirm cases.

barbibulle (approved):

PR #183, opened 2023-04-30
fix keystore save implementation for windows

barbibulle:
use `os.replace` instead of `os.rename`.


uael (approved):
Lgtm!
For reference: https://stackoverflow.com/questions/69363867/difference-between-os-replace-and-os-rename

PR #186, opened 2023-05-05
GATT included service declaration & discovery

zxzxwu:


bumble/gatt.py
```diff
@@ -205,8 +205,16 @@ class Service(Attribute):
     '''
 
     uuid: UUID
+    characteristics: list[Characteristic]
```
barbibulle, line 208: since we still support Python 3.8, we can't subscript builtin types, we have to use `List[xxx]` instead of `list[xxx]`. I notice that there's a couple other places in this file that already had the wrong case before this PR. Could you fix those as well?

bumble/gatt.py
```diff
@@ -219,9 +227,10 @@ def __init__(self, uuid, characteristics: list[Characteristic], primary=True):
             uuid.to_pdu_bytes(),
         )
         self.uuid = uuid
-        # self.included_services = []
+        self.included_services = included_services
```
barbibulle, line 230: Should make a copy of the list here, like for `characteristics`, so that if the caller modifies  the list later on, this won't affect the local copy.

bumble/gatt.py
```diff
@@ -231,6 +240,19 @@ def get_advertising_data(self) -> Optional[bytes]:
         """
         return None
 
+    def get_included_service_declaration(self) -> Attribute:
```
barbibulle, line 243: suggest: create an `IncludedServiceDeclaration`, just like `CharacteristicDeclaration`. This class should have a `service` property that points to the service object it describes (like `CharacteristicDeclaration` has a pointer to its `characteristic`.

bumble/gatt.py
```diff
@@ -219,9 +227,10 @@ def __init__(self, uuid, characteristics: list[Characteristic], primary=True):
             uuid.to_pdu_bytes(),
         )
         self.uuid = uuid
-        # self.included_services = []
+        self.included_services = included_services
         self.characteristics = characteristics[:]
         self.primary = primary
+        self.registered = False
```
barbibulle, line 233: I think it would be better to keep track of that state in the server rather than in the service itself. It's a bit dangerous to have `Service` objects be mutable by the server. They really should behave like read-only objects. There's maybe a small performance penalty in keep the state out of the object, but since a GATT DB really can't get all that large, that would never be an issue in practice.

bumble/gatt_server.py
```diff
@@ -222,7 +222,13 @@ def add_service(self, service: Service):
         # Add the service attribute to the DB
         self.add_attribute(service)
 
-        # TODO: add included services
+        # Add all included service
+        for included_service in service.included_services:
+            # Not registered yet, register the included service first.
+            if not included_service.registered:
```
barbibulle, line 228: see earlier comment about keeping that state in the Server class.

bumble/gatt_server.py
```diff
@@ -274,6 +280,7 @@ def add_service(self, service: Service):
 
         # Update the service group end
         service.end_group_handle = self.attributes[-1].handle
+        service.registered = True
```
barbibulle, line 283: see earlier comment about the mutability of the Service object.

bumble/gatt.py
zxzxwu: Done

bumble/gatt.py
zxzxwu: Done

bumble/gatt.py
zxzxwu: Done

bumble/gatt.py
zxzxwu: Done

bumble/gatt_server.py
zxzxwu: Done

bumble/gatt_server.py
zxzxwu: Done

bumble/gatt.py
```diff
@@ -253,6 +261,33 @@ def __init__(self, characteristics, primary=True):
...
+# -----------------------------------------------------------------------------
+class IncludedServiceDefinition(Attribute):
+    '''
+    See Vol 3, Part G - 3.2 INCLUDE DEFINITION
+    '''
+
+    service: Service
+
+    def __init__(self, service):
+        definition_bytes = struct.pack(
```
barbibulle, line 273: nit: `definition_bytes` -> `declaration_bytes`

barbibulle (approved):

bumble/gatt.py
```diff
@@ -253,6 +261,33 @@ def __init__(self, characteristics, primary=True):
         super().__init__(self.UUID, characteristics, primary)
 
 
+# -----------------------------------------------------------------------------
+class IncludedServiceDefinition(Attribute):
```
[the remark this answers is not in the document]
zxzxwu: Done

bumble/gatt.py
zxzxwu: Done

uael (approved):

PR #187, opened 2023-05-09
classic: handle Simple Pairing Complete event

uael:


bumble/host.py
```diff
@@ -807,6 +807,19 @@ def on_hci_simple_pairing_complete_event(self, event):
             f'simple pairing complete for {event.bd_addr}: '
             f'status={HCI_Constant.status_name(event.status)}'
         )
+        # Notify the client
+        if event.status == HCI_SUCCESS:
+            self.emit('ssp_complete', event.bd_addr)
```
barbibulle, line 812: suggestion: we should call this even `simple_pairing`, to parallel the `pairing` event used for LE pairing. (we could even call it just `pairing` but then we'd have to modify the args, because for LE pairing we do pass the keys and the `sc` flag). Maybe `simple_pairing` and `simple_pairing_failure` is a good way to make it clear that this behaves differently from the LE case (also we don't have the `pairing_start` event here, and if we have CTKD with two pairings, differentiating between the two phases would be useful).

uael:
Fixed by https://github.com/google/bumble/pull/240

PR #190, opened 2023-05-17
pandora: import bumble pandora server from avatar

uael:
See [asop/2584359](https://android-review.git.corp.google.com/c/platform/external/python/bumble/+/2584359) for details.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,7 @@ url = https://github.com/google/bumble
 
 [options]
 python_requires = >=3.8
-packages = bumble, bumble.transport, bumble.profiles, bumble.apps, bumble.apps.link_relay
+packages = bumble, bumble.transport, bumble.profiles, bumble.apps, bumble.apps.link_relay, bumble.pandora
 package_dir =
     bumble = bumble
     bumble.apps = apps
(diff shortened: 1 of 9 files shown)


DeltaEvo (approved):

tasks.py
```diff
@@ -111,7 +111,10 @@ def release_test(ctx):
 # -----------------------------------------------------------------------------
 @task
 def lint(ctx, disable='C,R', errors_only=False):
-    options = []
+    options = [
+        # FIXME: pylint does not support stubs yet:
+        '--ignore pandora'
```
barbibulle, line 116: could this be done in pyproject.toml instead?

setup.cfg
```diff
@@ -60,6 +61,7 @@ console_scripts =
     bumble-usb-probe = bumble.apps.usb_probe:main
     bumble-link-relay = bumble.apps.link_relay.link_relay:main
     bumble-bench = bumble.apps.bench:main
+    bumble-pandora_server = bumble.apps.pandora_server:main
```
barbibulle, line 64: I recommend using `bumble-pandora-server` (`-` instead of `_`) to be consistent with the other entries.

barbibulle (approved):

PR #195, opened 2023-06-04
speaker app

barbibulle:
Simple command-line + web UI virtual speaker app.
Support A2DP sink with SBC and AAC.
Outputs: file, in-browser playback (with visualization), ffplay pipe (if installed).
Can be used in passive mode (wait for connection) or active mode (connect to remote device at the start).


uael (approved):
Since it's a bit big I only review files under `bumble/` and just a quick read on the rest, the overall LGTM !

PR #197, opened 2023-06-05
Add typing for HFP and RFCOMM

zxzxwu:


bumble/hfp.py
```diff
@@ -20,7 +20,8 @@
 import collections
 
 from .colors import color
-
+from .rfcomm import DLC
+from typing import Union
```
barbibulle, line 24: style: we usually try to keep the imports grouped, with the builtin modules first, followed by the 3rd party modules, then finally the local ones. So I'd recommend moving the typing import up to line 21

bumble/rfcomm.py
```diff
@@ -23,6 +23,7 @@
 from . import core
 from .colors import color
 from .core import BT_BR_EDR_TRANSPORT, InvalidStateError, ProtocolError
+from typing import Optional, Tuple, Callable, Dict
```
barbibulle, line 26: see comment about import grouping on the other commit.

bumble/rfcomm.py
```diff
@@ -368,23 +397,23 @@ def __init__(self, multiplexer, dlci, max_frame_size, initial_tx_credits):
...
 
-    def change_state(self, new_state):
+    def change_state(self, new_state: int) -> None:
         logger.debug(
             f'{self} state change -> {color(self.state_name(new_state), "magenta")}'
         )
         self.state = new_state
 
-    def send_frame(self, frame):
+    def send_frame(self, frame) -> None:
```
barbibulle, line 409: maybe `frame` should be typed too?

bumble/rfcomm.py
```diff
@@ -592,7 +621,13 @@ class Multiplexer(EventEmitter):
...
     }
 
-    def __init__(self, l2cap_channel, role):
+    connection_result: Optional[asyncio.Future]
+    disconnection_result: Optional[asyncio.Future]
+    open_result: Optional[asyncio.Future]
+    acceptor: Optional[Callable[[int], bool]]
+    dlcs: Dict[int, DLC]
+
+    def __init__(self, l2cap_channel, role) -> None:
```
barbibulle, line 630: maybe `l2cap_channel` and `role` should be typed?

barbibulle (approved):

bumble/rfcomm.py
zxzxwu: Done. BTW, we have many enum-like integer fields which cannot be handled by typing.Literal.

bumble/rfcomm.py
zxzxwu: Done

bumble/rfcomm.py
zxzxwu: Done

bumble/hfp.py
zxzxwu: Done

bumble/rfcomm.py
zxzxwu: Let's resolve this issue later. These are all internal states, so I think we can use some enum to replace them.

PR #199, opened 2023-06-07
Add support for legacy pairing over bt classic

benquike:
A


apps/pair.py
```diff
@@ -157,6 +157,26 @@ async def display_number(self, number, digits):
...
+        count = 0
+        while True:
+            response = await self.prompt('>>> Enter PIN (1-6 chars):')
+            if len(response) == 0:
+                count += 1
+                if count > 3:
+                    self.print('too many tries, stopping the pairing')
+                    return None
+
+                self.print('no PIN was not entered, try again')
```
barbibulle, line 176: `no PIN was not entered` -> `no PIN was entered`

barbibulle (approved):

apps/pair.py
benquike: Updated.

PR #203, opened 2023-06-18
realtek driver

barbibulle:
Add support for loading FW images and config on Realtek USB dongles


bumble/drivers/rtk.py
```diff
@@ -0,0 +1,636 @@
...
+        if (path := pathlib.Path(__file__).parent / "rtk_fw" / file_name).is_file():
+            logger.debug(f"{file_name} found in package dir")
+            return path
+
+        # Finally look in the current directory
+        if (path := pathlib.Path.cwd() / file_name).is_file():
+            logger.debug(f"{file_name} found in CWD")
+            return path
+
+        return None
```
uael, line 456: Do you think it's worth it to also look into `/lib/firmware/rtl_bt/` as a last chance ?

bumble/drivers/rtk.py
```diff
@@ -0,0 +1,636 @@
...
+        self.driver_info = driver_info
+        self.firmware = firmware
+        self.config = config
+
+    @staticmethod
+    async def drop_firmware(host):
+        host.send_hci_packet(HCI_RTK_Drop_Firmware_Command())
+
+        # Wait for the command to be effective (no response is sent)
+        await asyncio.sleep(Driver.POST_DROP_DELAY)
```
uael, line 546: nit: perform a reset instead ?

bumble/drivers/rtk.py
```diff
@@ -0,0 +1,636 @@
...
+
+    @classmethod
+    async def for_host(cls, host, force=False):
+        # Check that a driver is needed for this host
+        if not force and not cls.check(host):
+            return None
+
+        # Get the driver info
+        if (driver_info := await cls.driver_info_for_host(host)) is None:
+            return None
```
uael, line 511: We can have a last chance mode where driver info does not exists and fw & confirg name are generated using ?
```
soc_version = lmp_subversion << 4 | (hci_revision & 0xF)
fw_name = "rtl%x_%s.bin" % (soc_version, "fw")
config_name = "rtl%x_%s.bin" % (soc_version, "config")
```

uael (approved):

bumble/drivers/rtk.py
barbibulle: Yes, excellent idea.

bumble/drivers/rtk.py
barbibulle: I'm concerned that performing a reset instead of waiting may lead to a deadlock, because if the reset command is sent while the  dongle FW is restarting, it may be dropped, and we'd be waiting for a response for ever.

bumble/drivers/rtk.py
barbibulle: I'm worried that this could be misleading: if there's no entry in the list of known FW images, this could mislead the user into thinking that one does exist, with a specific name, and they would be looking for it online somehow, but we don't know if it would actually be the right one. For example, for some devices, the FW image name doesn't follow this pattern (ex: `rtl8822cu_fw.bin`)

bumble/drivers/rtk.py
uael: I've just realized it's only used for the command line tool so I guess it's fine!

bumble/drivers/rtk.py
uael: Cool 😄

bumble/drivers/rtk.py
uael: Sounds good to me!

PR #210, opened 2023-07-12
use r0 by default for classic connection requests

barbibulle:
I don't remember why this was set as r2 by default, but r0 is a better default, as it will connect better/faster.
(a future PR should allow this to be customized of course)


uael:
Actually, by experience using R0 will result in far more page timeouts. I need to find the reason why since I don't quite remember! Let me check the spec

uael:
Ok it depends on the page scan activity of the remote device:
- R0 listens continuously, the initiator will not miss it
- Listen every 1.28s, the initiator can miss it if it uses R0
- Listen every 2.56s, the initiator can miss it if it uses R0 or R1

Since we don't know the peer device, the safer mode is R2 which is more aggressive. But I don't think it will be slower 🤔

PR #212, opened 2023-07-12
smp: handle security request command

uael:


zxzxwu:
I think it might be better to emit a `security_request` event, and let pandora or other application handle it.

uael:
I'm OOO, feel free to take the ownership on this 👍

uael:
Supersede by #291

Issue #215, opened 2023-07-19
Possibility of connecting bumble tcp-server controller with Zephyr application?


vChavezB:
I am reading the docs but something was not clear to me.

My use case is the following, I am developing a BLE peripheral with Zephyr OS. This RTOS has the option of compiling as a linux posix binary to do unit tests, host debugging, etc. It also allows to use an hci linux interface to use the bluetooth controller from host.

https://github.com/zephyrproject-rtos/zephyr/blob/e47fc45c4d4ed3c04d6a147e10d2a52630b7f0d7/drivers/bluetooth/hci/userchan.c#L193

My question is if its possible to add a virtual hci with bumble and make my linux binary (BLE Peripheral) connect to it. Afterwards I would like to have a second virtual hci that I can connect to test that my bluetooth application (BLE Central) works (integration tests).

(BLE Peripheral)_________________________________________________________________(BLE Central)
Zephyr OS Linux Binary   <----> Virtual HCI    < -----> Virtual HCI <-----> Integration test application (python, C, java...)

From the bumble website I see that this would be covered in Use Case 4. Which example should I refer to create the two virtual HCI interfaces and link my "emulated devices" to these virtual HCIs.?

In addition I might be interested in Use Case 3, where I do my integration tests directly with Bumble. Which example should I refer to for this purpose?

barbibulle:
[...]

PR #216, opened 2023-07-22
accept Host Buffer Size Command in the controller

barbibulle:
The Zephyr stack sends this command. With this PR, the controller accepts it so it doesn't return an error, but it still doesn't honor the values sent.  So if the buffers are indeed restricted, it may not work.
This is a temporary fix until a more complete support for custom buffer sizes can be supported.

### DIFF
--- a/bumble/controller.py
+++ b/bumble/controller.py
@@ -654,7 +654,7 @@ class Controller:
 
     def on_hci_create_connection_command(self, command):
         '''
-        See Bluetooth spec Vol 2, Part E - 7.1.5 Create Connection command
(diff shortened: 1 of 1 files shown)


vChavezB:
As I mentioned here

https://github.com/google/bumble/issues/215#issuecomment-1646521126

I tested this PR but the Zephyr stack still gave me an error because the command "Set Controller To Host Flow Control Command" was missing.

I made a fork that solves this

I suggest to add the following commit to this PR 

https://github.com/vChavezB/bumble/commit/66620dbd4860889695c21ccd98223f23112c0d47

feel free to add this change to this PR or I can also submit a PR with the change 
BR

uael (approved):

PR #222, opened 2023-07-25
SMP: Do not send phase 2 commands in CTKD over Classic

zxzxwu:


tests/self_test.py
```diff
@@ -473,6 +477,85 @@ async def compare_numbers(self, number, digits):
...
+        two_devices.devices[1].public_address,
+        LINK_KEY,
+        HCI_AUTHENTICATED_COMBINATION_KEY_GENERATED_FROM_P_256_TYPE,
+    )
+    two_devices.devices[1].on_link_key(
+        two_devices.devices[0].public_address,
+        LINK_KEY,
+        HCI_AUTHENTICATED_COMBINATION_KEY_GENERATED_FROM_P_256_TYPE,
+    )
+    Connection.is_encrypted = lambda: True
```
uael, line 527:
```suggestion
    two_devices.connections[0].encryption = 1
    two_devices.connections[1].encryption = 1
```

zxzxwu: Done

bumble/smp.py
```diff
@@ -553,19 +554,23 @@ def resolve(self, address):
...
     OOB = 3
 
-    PAIRING_METHOD_NAMES = {
-        JUST_WORKS: 'JUST_WORKS',
-        NUMERIC_COMPARISON: 'NUMERIC_COMPARISON',
-        PASSKEY: 'PASSKEY',
-        OOB: 'OOB',
-    }
+    def __str__(self) -> str:
+        return {
```
barbibulle, line 565: `return self.name` should return the symbol name.

barbibulle (approved):

bumble/smp.py
zxzxwu: Done. Invoke `.name()` when printing might be a better idea.

zxzxwu:
I just realized that mock methods are not released after the test, though there is no following tests right now. To solve this problem, we need to wrap all test cases into test class.

zxzxwu:
> I just realized that mock methods are not released after the test, though there is no following tests right now. To solve this problem, we need to wrap all test cases into test class.

Oh not really. That can be protected by a `with` block.

zxzxwu:
Ready to review

uael (commented):
Code LGTM, let me check with the spec first before approval

bumble/smp.py
```diff
@@ -769,19 +769,23 @@ def get_long_term_key(self, rand: bytes, ediv: int) -> Optional[bytes]:
     def decide_pairing_method(
         self, auth_req: int, initiator_io_capability: int, responder_io_capability: int
     ) -> None:
+        if self.connection.transport == BT_BR_EDR_TRANSPORT:
```
uael, line 772: Are you sure about this ? Can you add a comment explaining a bit + link to the spec ?

zxzxwu: In Core Vol 3, Part H, Chapter 3.3 Command Format, it states that only the following commands are supported on BR/EDR transport:
* Pairing Request
* Pairing Response
* Pairing Fail
* Identity Info 
* Identity Address Info
* Signing Info

Which implied that Phase 2 authentication is impossible on Classic SMP channel - just like all CTKD-related spec, it's not very clear but that's something we can infer.

uael: sg

uael (approved):

PR #223, opened 2023-07-26
Add type hints to L2CAP module

zxzxwu:
Also slightly refactor L2CAP channel tables. The original implementation mix BR/EDR and LE channels.

### DIFF
--- a/bumble/l2cap.py
+++ b/bumble/l2cap.py
@@ -22,7 +22,18 @@ import struct
 
 from collections import deque
 from pyee import EventEmitter
-from typing import Dict, Type, List, Optional, Tuple, Callable, Any, Union, Deque
+from typing import (
+    Dict,
+    Type,
+    List,
+    Optional,
+    Tuple,
+    Callable,
+    Any,
+    Union,
+    Deque,
+    Iterable,
+)
 
 from .colors import color
 from .core import BT_CENTRAL_ROLE, InvalidStateError, ProtocolError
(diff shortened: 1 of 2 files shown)


zxzxwu:
This work is much larger than I expected...

uael:
We don't need to do the whole thing at once, a subset is good enough.
But I'm wondering if it's worth introducing breaking changes by refactoring at the same time? Maybe we should just add some typing at first, what do you think?

zxzxwu:
Right, let me think how to minimize this

zxzxwu:
I removed the refactor of code enum, and avoid any hints related to logic change.

bumble/l2cap.py
```diff
@@ -256,29 +271,29 @@ def inner(cls):
...
         if hasattr(self, 'fields') and kwargs:
             HCI_Object.init_from_fields(self, self.fields, kwargs)
-        if pdu is None:
-            data = HCI_Object.dict_to_bytes(kwargs, self.fields)
-            pdu = (
-                bytes([self.code, self.identifier])
-                + struct.pack('<H', len(data))
-                + data
-            )
+            if pdu is None:
```
uael, line 278: `if hasattr(self, 'fields') and pdu is None:` and keep depth as is ?

uael (approved):

bumble/l2cap.py
zxzxwu: Oh I missed kwarg. How about
```py
def __init__(self, pdu=None, **kwargs) -> None:
        self.identifier = kwargs.get('identifier', 0)
        if hasattr(self, 'fields'):
            if kwargs:
                HCI_Object.init_from_fields(self, self.fields, kwargs)
            if pdu is None:
                data = HCI_Object.dict_to_bytes(kwargs, self.fields)
                pdu = (
                    bytes([self.code, self.identifier])
                    + struct.pack('<H', len(data))
                    + data
                )
        self.pdu = pdu
```

uael: lgtm!

PR #226, opened 2023-07-27
pandora: refactor server a bit

uael:
The first commit is a working experiment of using only one HCI TCP connection.
The next one: Avatar improved it's logging, so the server one can decrease do debug logging only


barbibulle (approved):

uael:
We are still unsure of the effectiveness of this for avatar, marking this as a draft for now 👍

uael:
Not needed anymore :+1:

PR #227, opened 2023-07-27
compatibility with python 11

barbibulle:
Python 11's handling of enums has changed a bit, so the conversion to string behaves differently from before.
This should make the code and type checking compatible with the new modes.


barbibulle:
This breaks python < 11, I will push an update.

zxzxwu (approved):
LGTM

uael (approved):
I've not tested in AOSP yet but it should work! If any issues is encountered we can still fixes in follow up PR

PR #230, opened 2023-07-31
add support for field arrays in hci packet definitions

barbibulle:
This adds support for specifying field arrays in HCI packet definitions, in a way that's compatible with how this was done before with custom parsers (i.e for an HCI packet with an array of `some_field`, the object will have a `some_fields` property that is an array, including in the constructor.
To use this new syntax, use a array entry instead of a tuple for one of the `fields` definition in the class decorator (see `HCI_LE_Set_Extended_Advertising_Enable_Command` for an example.


uael (approved):
Cool :)

zxzxwu (approved):

PR #231, opened 2023-08-01
SMP: Delegate distributed address type

zxzxwu:
There might be some better solutions here?


uael (commented):
This actually sounds good to me, @barbibulle what do you think ?
Using the connection self address is a good default :+1:

barbibulle (approved):

PR #233, opened 2023-08-02
Pandora: Configure identity address type

zxzxwu:


uael (approved):
lgtm !

bumble/pandora/config.py
```diff
@@ -35,6 +36,7 @@ def load_from_dict(self, config: Dict[str, Any]) -> None:
             'io_capability', 'no_output_no_input'
         ).upper()
         self.io_capability = getattr(PairingDelegate, io_capability_name)
+        self.identity_address_type = config.get('identity_address_type', None)
```
uael, line 39: Add parsing from string ?

uael (changes requested):

bumble/pandora/config.py
uael:
```suggestion
        identity_address_type_name: str = config.get(
            'identity_address_type', 'random'
        ).upper()
        self.identity_address_type = getattr(PairingConfig.AddressType, identity_address_type_name)
```

zxzxwu: Done

bumble/pandora/config.py
```diff
@@ -35,6 +36,12 @@ def load_from_dict(self, config: Dict[str, Any]) -> None:
             'io_capability', 'no_output_no_input'
         ).upper()
         self.io_capability = getattr(PairingDelegate, io_capability_name)
+        identity_address_type_name: str = config.get(
```
uael, line 39: Ah, my bad. If you think `None` is a better default this need to be worked out a bit

zxzxwu: Actually I think that's a good idea to force setting a type when using Pandora, otherwise the default behavior is a little bit ambiguous which implies the device has 2 identity address. The thing left is I should remove `Optional`.

PR #234, opened 2023-08-02
Support address resolution offload

zxzxwu:
* Auto add device to resolving list after pairing
* Add a flag to determine whether address resolution offlaod is enabled

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -86,6 +86,7 @@ from .hci import (
     HCI_LE_Extended_Create_Connection_Command,
     HCI_LE_Rand_Command,
     HCI_LE_Read_PHY_Command,
+    HCI_LE_Set_Address_Resolution_Enable_Command,
     HCI_LE_Set_Advertising_Data_Command,
     HCI_LE_Set_Advertising_Enable_Command,
     HCI_LE_Set_Advertising_Parameters_Command,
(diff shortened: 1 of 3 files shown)


uael (approved):

bumble/device.py
```diff
@@ -778,6 +779,7 @@ def __init__(self) -> None:
         self.irk = bytes(16)  # This really must be changed for any level of security
         self.keystore = None
         self.gatt_services: List[Dict[str, Any]] = []
+        self.address_resolution = False
```
barbibulle, line 782: It might be better to call this `address_resolution_in_controller` or `address_resolution_offload`, because address resolution always happens, it's just whether it happens on the host or the controller that's different.

bumble/device.py
```diff
@@ -1314,6 +1299,24 @@ async def power_off(self) -> None:
             await self.host.flush()
             self.powered_on = False
 
+    async def refresh_resolving_list(self) -> None:
+        assert self.keystore
+
+        await self.send_command(HCI_LE_Clear_Resolving_List_Command())  # type: ignore[call-arg]
```
barbibulle, line 1305: this command should only be sent if address resolution offloading is enabled.

zxzxwu: Done

bumble/device.py
zxzxwu: Done

barbibulle (approved):

PR #238, opened 2023-08-03
controller extended advertising

barbibulle:
Still work in progress.
Enables basic support for extended advertising in the virtual controller object.


apps/controller_info.py
```diff
@@ -56,6 +60,38 @@ def command_succeeded(response):
...
+async def get_common_info(host):
+    if host.supports_command(HCI_READ_LOCAL_SUPPORTED_FEATURES_COMMAND):
+        response = await host.send_command(
+            HCI_Read_Local_Supported_Features_Command()
+        )
+        if response.return_parameters.status == HCI_SUCCESS:
+            print()
+            print(color('LMP Features:', 'yellow'))
+            # TODO: support printing discrete enum values
+            print('  ', response.return_parameters.lmp_features.hex())
```
zxzxwu, line 73: I think it would be helpful to parse features into `set` or `IntFlag` or something else because sometimes we need to check controller capabilities for both dev and product evaluation.

barbibulle: I agree. I'll try to add that in a separate PR.

apps/controller_info.py
```diff
@@ -56,6 +60,36 @@ def command_succeeded(response):
...
+
+    if host.supports_command(HCI_READ_LOCAL_EXTENDED_FEATURES_COMMAND):
+        response = await host.send_command(
+            HCI_Read_Local_Extended_Features_Command(page_number=0)
+        )
+        if response.return_parameters.status == HCI_SUCCESS:
+            if response.return_parameters.max_page_number > 0:
+                print()
```
zxzxwu, line 79: Is the empty line intended?

bumble/controller.py
```diff
@@ -78,6 +173,25 @@ class DataObject:
     pass
 
 
+# -----------------------------------------------------------------------------
+def le_supported_features_as_bytes(supported_features):
+    return struct.pack('<Q', sum((1 << bit for bit in supported_features)))
+
+
+# -----------------------------------------------------------------------------
+def supported_commands_as_bytes(supported_commands):
```
zxzxwu, line 182:
```suggestion
def supported_commands_as_bytes(supported_commands: Sequence[HCI_Command]) -> bytes:
```

bumble/controller.py
```diff
@@ -78,6 +173,25 @@ class DataObject:
     pass
 
 
+# -----------------------------------------------------------------------------
+def le_supported_features_as_bytes(supported_features):
```
zxzxwu, line 177:
```suggestion
def le_supported_features_as_bytes(supported_features: Sequence[int]) -> bytes:
```

bumble/controller.py
```diff
@@ -1255,6 +1625,190 @@ def on_hci_le_set_default_phy_command(self, command):
...
+        See Bluetooth spec Vol 2, Part E - 7.8.54 LE Set Extended Advertising Data
+        Command
+        '''
+        if (
+            advertiser := self.extended_advertisers.get(
+                command.advertising_handle, None
+            )
+        ) is None:
+            return bytes([HCI_UNKNOWN_ADVERTISING_IDENTIFIER_ERROR])
+
```
zxzxwu, line 1689: nit: Add a TODO here to replace literals with Operation enums

bumble/controller.py
```diff
@@ -975,24 +1346,25 @@ def on_hci_le_set_advertising_data_command(self, command):
...
         '''
         See Bluetooth spec Vol 4, Part E - 7.8.9 LE Set Advertising Enable Command
         '''
         if command.advertising_enable:
-            self.start_advertising()
+            self.legacy_advertiser.enabled = True
+            self.start_advertising_timer()
         else:
-            self.stop_advertising()
+            self.legacy_advertiser.enabled = True
```
zxzxwu, line 1367:
```suggestion
            self.legacy_advertiser.enabled = False
```

bumble/controller.py
```diff
@@ -623,32 +994,42 @@ def on_classic_role_change(self, peer_address, new_role):
...
     def send_advertising_data(self):
-        if self.link and self.advertising_data:
-            self.link.send_advertising_data(self.random_address, self.advertising_data)
+        # Legacy advertising
+        self.legacy_advertiser.send_advertising_data(self.link)
 
-    @property
-    def is_advertising(self):
-        return self.advertising_timer_handle is not None
+        # Extended advertising
```
zxzxwu, line 1030: Should all advertisers send data on each of them was scheduled?

bumble/controller.py
```diff
@@ -623,32 +994,42 @@ def on_classic_role_change(self, peer_address, new_role):
...
-    def start_advertising(self):
-        # Stop any ongoing advertising before we start again
-        self.stop_advertising()
+        self.send_advertising_data()
 
-        # Advertise now
-        self.advertising_timer_handle = asyncio.get_running_loop().call_soon(
-            self.on_advertising_timer_fired
+        # Compute the time of the next advertisement
+        next_advertisement = min(
```
zxzxwu, line 1002: nit: This could be replaced with a heapq
```py

# Advertiser
def send_advertising_data(self):
  ...
  heappush(timer_heap, (time.time()+interval, self))
  
def on_advertising_timer_fired(self):
  next_advertisement, advertiser = heappop(timer_heap)
  advertiser.send_advertising_data()
```

zxzxwu:
By the way, as RootCanal has implemented most of the feature, can we leverage the work instead of remaking the wheels?

barbibulle:
> By the way, as RootCanal has implemented most of the feature, can we leverage the work instead of remaking the wheels?

That’s exactly the plan: the local controller class is not intended to be compete and will not be enhanced in the future. It is only there to continue support for basic local unit testing and used by a few external users who have not moved to RootCanal yet. But now that RootCanal is available as a standalone project, the plan is to use that for any sort of full-feature need.

apps/controller_info.py
barbibulle: Yes. That's to ensure a blank line before the section.

bumble/controller.py
barbibulle: Done

bumble/controller.py
barbibulle: Done

bumble/controller.py
barbibulle: This is refactored now

PR #240, opened 2023-08-09
Handle SSP Complete events

zxzxwu:
A refined version of https://github.com/google/bumble/pull/187

`on_ssp_complete()` is removed due to no usage, and this logic should be handled by applications.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -2743,20 +2743,6 @@ class Device(CompositeEventEmitter):
         )
         connection.emit('connection_authentication_failure', error)
 
-    @host_event_handler
-    @with_connection_from_address
-    def on_ssp_complete(self, connection):
-        # On Secure Simple Pairing complete, in case:
-        # - Connection isn't already authenticated
-        # - AND we are not the initiator of the authentication
-        # We must trigger authentication to know if we are truly authenticated
-        if not connection.authenticating and not connection.authenticated:
-            logger.debug(
-                f'*** Trigger Connection Authentication: [0x{connection.handle:04X}] '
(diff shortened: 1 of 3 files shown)


zxzxwu:
Actually I am wondering if `simple` is needed, because legacy pairing doesn't produce any HCI events but `HCI_Link_Key_Notification`(which also means we cannot know it's failed).

barbibulle (approved):

barbibulle:
> Actually I am wondering if `simple` is needed, because legacy pairing doesn't produce any HCI events but `HCI_Link_Key_Notification`(which also means we cannot know it's failed).

Good point. I think maybe just `classic_pairing` should be enough. Either way is fine.

bumble/device.py
```diff
@@ -2743,20 +2743,6 @@ def on_connection_authentication_failure(self, connection, error):
...
         connection.emit('connection_authentication_failure', error)
 
-    @host_event_handler
-    @with_connection_from_address
-    def on_ssp_complete(self, connection):
-        # On Secure Simple Pairing complete, in case:
-        # - Connection isn't already authenticated
-        # - AND we are not the initiator of the authentication
-        # We must trigger authentication to know if we are truly authenticated
-        if not connection.authenticating and not connection.authenticated:
```
uael, line 2753: Even if this was dead code I think it still remains valid, what do you think?

zxzxwu: I highly disencourage this behavior, because it may raise authentication collision, and it's not something listed on the spec

zxzxwu:
I think this PR is ready :+1:

PR #242, opened 2023-08-10
Handle SMP_Security_Request

zxzxwu:


bumble/smp.py
```diff
@@ -1200,7 +1205,7 @@ def on_peer_key_distribution_complete(self) -> None:
         self.connection.abort_on('disconnection', self.on_pairing())
 
     def on_connection_encryption_change(self) -> None:
-        if self.connection.is_encrypted:
+        if self.connection.is_encrypted and self.state == Session.State.PAIRING:
```
uael, line 1208: Are we sure `on_pairing[_failure]` to not be called before this can happen ?

uael (approved):

PR #244, opened 2023-08-12
add sink method for lost transports

barbibulle:
This allows calls to be unblocked from waiting for a response when the transport goes away.
The new sink method `on_transport_lost` is optional for sinks. A source should call it when it is terminated.

### DIFF
--- a/bumble/host.py
+++ b/bumble/host.py
@@ -459,6 +459,8 @@ class Host(AbortableEventEmitter):
         if self.pending_response:
             self.pending_response.set_exception(TransportLostError('transport lost'))
 
+        self.emit('flush')
+
     def on_hci_packet(self, packet):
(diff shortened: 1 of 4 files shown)


bumble/host.py
```diff
@@ -455,6 +454,11 @@ def on_packet(self, packet):
         else:
             logger.debug('reset not done, ignoring packet from controller')
 
+    def on_transport_lost(self):
+        # Called by the source when the transport has been lost.
+        if self.pending_response:
```
uael, line 459: Emit the 'flush' event?

uael (approved):

bumble/host.py
barbibulle: Yes, that makes sense. I'll add it.

5678bgu: []()

5678bgu:
__

PR #248, opened 2023-08-17
pandora_server: add support to accept bumble config file

SilverBzH:


apps/pandora_server.py
```diff
@@ -18,13 +20,30 @@
...
+    if 'transport' not in bumble_config.keys():
+        bumble_config.update({'transport': transport})
+    device = PandoraDevice(bumble_config)
+
     logging.basicConfig(level=logging.DEBUG)
     asyncio.run(serve(device, port=grpc_port))
 
 
+def retrieve_config(config: str) -> Dict[str, Any]:
+    if len(config) == 0:
```
barbibulle, line 41: recommend: using "implicit" false test (https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations)

apps/pandora_server.py
```diff
@@ -18,13 +20,30 @@
...
+        return {}
+    else:
```
barbibulle, line 43: `else` not needed here.

barbibulle (approved):

apps/pandora_server.py
SilverBzH: Indeed this is cleaner, thanks !

apps/pandora_server.py
SilverBzH: Of course 👍

PR #249, opened 2023-08-18
Support SBC in speaker.app

zxzxwu:


barbibulle (approved):

barbibulle:
Thanks for the PR. Somehow I thought that ffplay didn't support SBC over a pipe, but it seems that it does.

PR #253, opened 2023-08-21
Adding more typing in rfcomm.py

zxzxwu:


bumble/rfcomm.py
```diff
@@ -263,7 +266,7 @@ def __init__(
         self.window_size = window_size
 
     @staticmethod
-    def from_bytes(data: bytes):
+    def from_bytes(data: bytes) -> 'RFCOMM_MCC_PN':
```
barbibulle, line 270: this should use the class object rather than the name as a string.

bumble/rfcomm.py
```diff
@@ -318,7 +323,7 @@ def __init__(self, dlci: int, fc: int, rtc: int, rtr: int, ic: int, dv: int):
         self.dv = dv
 
     @staticmethod
-    def from_bytes(data: bytes):
+    def from_bytes(data: bytes) -> 'RFCOMM_MCC_MSC':
```
barbibulle, line 327: this should use the class object rather than the name as a string.

bumble/rfcomm.py
```diff
@@ -375,8 +380,12 @@ class DLC(EventEmitter):
     sink: Optional[Callable[[bytes], None]]
 
     def __init__(
-        self, multiplexer, dlci: int, max_frame_size: int, initial_tx_credits: int
-    ):
+        self,
+        multiplexer: 'Multiplexer',
```
barbibulle, line 376: this should use the class object rather than the name as a string.

bumble/rfcomm.py
```diff
@@ -846,7 +855,7 @@ class Client:
     multiplexer: Optional[Multiplexer]
     l2cap_channel: Optional[l2cap.Channel]
 
-    def __init__(self, device, connection) -> None:
+    def __init__(self, device: 'Device', connection: 'Connection') -> None:
```
barbibulle, line 830: this should use the class object rather than the name as a string.

bumble/rfcomm.py
```diff
@@ -886,7 +895,7 @@ async def shutdown(self) -> None:
 class Server(EventEmitter):
     acceptors: Dict[int, Callable[[DLC], None]]
 
-    def __init__(self, device) -> None:
+    def __init__(self, device: 'Device') -> None:
```
barbibulle, line 870: this should use the class object rather than the name as a string.

barbibulle (approved):

bumble/rfcomm.py
zxzxwu: (Also in other packet classes) This seems like a Pylance issue(I am using VS Code). Methods referencing its own class cannot be resolved, but mypy considers it's fine.

zxzxwu: Oh I got it on https://stackoverflow.com/questions/33533148

barbibulle (approved):

PR #254, opened 2023-08-21
eSCO codec/HCI definitions + Host support 

zxzxwu:
I only added HCI packets and eSCO codec definitions, as well as host handler. For `device`, it's not easy to integrate SCO into current handlers.

I also tested the SCO HCI path on Android, with IBluetoothHci. Links can be created and controllers consistently report packets, but later packets are all 0. It's probably because the host didn't send any SCO packet?

Most USB dongles don't support eSCO, so I haven't tried to setup them legacy SCO.

### DIFF
--- a/bumble/hci.py
+++ b/bumble/hci.py
@@ -17,6 +17,7 @@
 # -----------------------------------------------------------------------------
 from __future__ import annotations
 import collections
+import enum
 import functools
 import logging
 import struct
@@ -1368,6 +1369,7 @@ HCI_LE_SUPPORTED_FEATURES_NAMES = {
     if feature_name.startswith('HCI_') and feature_name.endswith('_LE_SUPPORTED_FEATURE')
 }
 
+
 # fmt: on
 # pylint: enable=line-too-long
 # pylint: disable=invalid-name
@@ -1923,6 +1925,9 @@ class HCI_Packet:
         if packet_type == HCI_ACL_DATA_PACKET:
             return HCI_AclDataPacket.from_bytes(packet)
 
+        if packet_type == HCI_SYNCHRONOUS_DATA_PACKET:
+            return HCI_SynchronousDataPacket.from_bytes(packet)
+
         if packet_type == HCI_EVENT_PACKET:
             return HCI_Event.from_bytes(packet)
 
@@ -2295,6 +2300,19 @@ class HCI_Read_Clock_Offset_Command(HCI_Command):
 @HCI_Command.command(
     fields=[
         ('bd_addr', Address.parse_address),
+        ('reason', {'size': 1, 'mapper': HCI_Constant.error_name}),
+    ],
+)
+class HCI_Reject_Synchronous_Connection_Request_Command(HCI_Command):
+    '''
+    See Bluetooth spec @ 7.1.28 Reject Synchronous Connection Request Command
+    '''
+
+
+# -----------------------------------------------------------------------------
+@HCI_Command.command(
+    fields=[
+        ('bd_addr', Address.parse_address),
         ('io_capability', {'size': 1, 'mapper': HCI_Constant.io_capability_name}),
         ('oob_data_present', 1),
         (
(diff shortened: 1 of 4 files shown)


zxzxwu:
Some thoughts:
* Probably not all controllers support PCM IO? Also, may it reduce the bandwidth/quality?
* If we use transparent IO, ffplay/liblc3 may handle the co/decoding, but not for CVSD.

bumble/hfp.py
```diff
@@ -747,3 +748,156 @@ def sdp_records(
...
+    SCO_CVSD_D1 = enum.auto()
+    ESCO_CVSD_S1 = enum.auto()
+    ESCO_CVSD_S2 = enum.auto()
+    ESCO_CVSD_S3 = enum.auto()
+    ESCO_CVSD_S4 = enum.auto()
+    ESCO_MSBC_T1 = enum.auto()
+    ESCO_MSBC_T2 = enum.auto()
+
+
+_ESCO_PARAMETERS_COMMON = {
```
barbibulle, line 769: maybe we could use a class (like a dataclass) instead of a dictionary here?

bumble/hci.py
```diff
@@ -5653,6 +5716,64 @@ def __str__(self):
         )
 
 
+# -----------------------------------------------------------------------------
+class HCI_SynchronousDataPacket:
+    '''
+    See Bluetooth spec @ 5.4.2 HCI ACL Data Packets
```
barbibulle, line 5722: Should be `5.4.3 HCI Synchronous Data packets`

bumble/hci.py
```diff
@@ -5653,6 +5716,64 @@ def __str__(self):
...
+    '''
+
+    hci_packet_type = HCI_SYNCHRONOUS_DATA_PACKET
+
+    @staticmethod
+    def from_bytes(packet: bytes) -> 'HCI_SynchronousDataPacket':
+        # Read the header
+        h, data_total_length = struct.unpack_from('<HB', packet, 1)
+        connection_handle = h & 0xFFF
+        packat_status = (h >> 12) & 3
```
barbibulle, line 5732: typo: packet

bumble/host.py
```diff
@@ -484,6 +487,10 @@ def on_hci_acl_data_packet(self, packet):
         if connection := self.connections.get(packet.connection_handle):
             connection.on_hci_acl_data_packet(packet)
 
+    def on_hci_sco_data_packet(self, packet):
+        # Look for the connection to which this data belongs
```
barbibulle, line 491: I don't think this comment applies here.

barbibulle:
> Some thoughts:
> 
> * Probably not all controllers support PCM IO? Also, may it reduce the bandwidth/quality?
> * If we use transparent IO, ffplay/liblc3 may handle the co/decoding, but not for CVSD.

We could implement CVSD in python, and leave other formats for external libraries (it would also be interesting to try and package liblc3 as a wasm binary and use it within python using a wasm interpreter)

barbibulle:
> I only added HCI packets and eSCO codec definitions, as well as host handler. For `device`, it's not easy to integrate SCO into current handlers.

I have some ideas on how to support eSCO connections at the Device level. I can try to push up a PR for that once this one is finished and merged.

bumble/hfp.py
zxzxwu: Modified

bumble/hci.py

bumble/hci.py

bumble/host.py
zxzxwu: Done

uael (approved):

bumble/hfp.py
```diff
@@ -747,3 +748,198 @@ def sdp_records(
...
+    input_transport_unit_size: int = 0
+    output_transport_unit_size: int = 0
+    input_bandwidth: int = 16000
+    output_bandwidth: int = 16000
+    transmit_bandwidth: int = 8000
+    receive_bandwidth: int = 8000
+    transmit_codec_frame_size: int = 60
+    receive_codec_frame_size: int = 60
+
+    def to_dict(self):
```
barbibulle, line 812: This is probably not necessary, since @dataclass already has `asdict` which does the same thing.

barbibulle (approved):

bumble/hfp.py
zxzxwu: Done

PR #258, opened 2023-08-25
Add support for Zephyr HCI VSC set TX power command

mogenson:
Add HCI Zephyr vendor commands to read and write TX power
    
Create platforms/zephyr/hci.py with definitions of vendor HCI commands
to read and write TX power.
    
Add documentation for how to prepare an nRF52840 dongle with a Zephyr
HCI USB firmware application that includes dynamic TX power support and
how to send a write TX power vendor HCI command from Bumble

### DIFF
--- a/apps/show.py
+++ b/apps/show.py
@@ -104,7 +104,7 @@ class SnoopPacketReader:
 )
 @click.option(
     '--vendors',
-    type=click.Choice(['android']),
+    type=click.Choice(['android', 'zephyr']),
     multiple=True,
     help='Support vendor-specific commands (list one or more)',
 )
@@ -114,6 +114,8 @@ def main(format, vendors, filename):
     for vendor in vendors:
         if vendor == 'android':
             import bumble.vendor.android.hci
+        elif vendor == 'zephyr':
+            import bumble.vendor.zephyr.hci
 
     input = open(filename, 'rb')
     if format == 'h4':
--- a/bumble/vendor/zephyr/hci.py
+++ b/bumble/vendor/zephyr/hci.py
@@ -1,0 +1,88 @@
+# Copyright 2021-2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# -----------------------------------------------------------------------------
+# Imports
+# -----------------------------------------------------------------------------
+from bumble.hci import (
+    hci_vendor_command_op_code,
+    HCI_Command,
+    STATUS_SPEC,
+)
+
+
+# -----------------------------------------------------------------------------
+# Constants
+# -----------------------------------------------------------------------------
+
+# Zephyr RTOS Vendor Specific Commands and Events.
+# Only a subset of the commands are implemented here currently.
+#
+# pylint: disable-next=line-too-long
+# See https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/bluetooth/hci_vs.h
+HCI_WRITE_TX_POWER_LEVEL_COMMAND = hci_vendor_command_op_code(0x000E)
+HCI_READ_TX_POWER_LEVEL_COMMAND = hci_vendor_command_op_code(0x000F)
+
+HCI_Command.register_commands(globals())
+
+
+# -----------------------------------------------------------------------------
+class TX_Power_Level_Command:
+    '''
+    Base class for read and write TX power level HCI commands
+    '''
+
+    TX_POWER_HANDLE_TYPE_ADV = 0x00
+    TX_POWER_HANDLE_TYPE_SCAN = 0x01
+    TX_POWER_HANDLE_TYPE_CONN = 0x02
+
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -63,3 +63,23 @@ This is not an official Google product.
 This library is in alpha and will be going through a lot of breaking changes. While releases will be stable enough for prototyping, experimentation and research, we do not recommend using it in any production environment yet.
 Expect bugs and sharp edges.
 Please help by trying it out, reporting bugs, and letting us know what you think!
+
+# TX Power Control for nrf52840 dongles
```
barbibulle, line 67: Maybe this doc should live in a section under `docs/mkdocs/src` (maybe `platforms/zephyr.md`?) rather than the top-level README, which should remain very generic.

bumble/hci.py
```diff
@@ -557,6 +557,13 @@ def phy_list_to_bits(phys):
 HCI_LE_SET_DEFAULT_SUBRATE_COMMAND                                       = hci_command_op_code(0x08, 0x007D)
 HCI_LE_SUBRATE_REQUEST_COMMAND                                           = hci_command_op_code(0x08, 0x007E)
 
+# Vendor specific HCI commands
+HCI_VSC_TX_POWER_HANDLE_TYPE_ADV = 0x00
```
barbibulle, line 561: Those 3 constants should be class members of the `HCI_VSC_Write_Tx_Power_Level_Command` class I think.

bumble/hci.py
```diff
@@ -2053,6 +2061,37 @@ def __str__(self):
...
+@HCI_Command.command(
+    fields=[('handle_type', 1), ('connection_handle', 2), ('tx_power_level', -1)],
+    return_parameters_fields=[
+        ('status', STATUS_SPEC),
+        ('handle_type', 1),
+        ('connection_handle', 2),
+        ('selected_tx_power_level', -1),
+    ],
+)
+class HCI_VSC_Write_Tx_Power_Level_Command(HCI_Command):
```
barbibulle, line 2074: Is there a doc/spec we can point to for this vendor command?

hci_usb.zip
barbibulle, line 1: We should find a way to host this type of pre-built firmware images on the static pages rather than in the Git repo.

mogenson:
Hi @barbibulle thanks for the comments. The vendor HCI command for TX power is here: https://github.com/zephyrproject-rtos/zephyr/blob/a15e7d2326d9e8896401be479fae56f30dde71bb/include/zephyr/bluetooth/hci_vs.h#L163

However, I'm going to close this PR because I don't think it makes sense to have vendor commands for specific hardware that require custom built firmware as part of Bumble. I will add the definition for these commands at the application level for our Python project. I just wanted to use this PR to demonstrate the feasibility of setting TX power.

barbibulle:
Actually, I think it would be great to make these HCI commands and tool extension part of the main project. I’m sure others will find it useful, even if Nordic/Zephyr specific.

barbibulle:
It might be good to put the vendor HCI packet definitions in their own file. Maybe hci_vendor.py. I’d leave it as low level commands for now. For normal HCI commands, there’s a way to query the controller if the command is supported, but I don’t know of a way to do that for vendor commands (the answer comes in the form of a long bit mask, so not extensible to custom extensions). It’s probably safe to let users decide if they want/need to send those commands, and let them catch errors if they send the command to a controller without knowing if their controller supports it. Since the high level APIs would not send those commands directly, we wouldn’t need a config entry to enable/disable them.

barbibulle:
> It might be good to put the vendor HCI packet definitions in their own file. Maybe `hci_vendor.py` or `hci_vsc.py`. 

I will push a PR for review with a structure and supporting functions for separating vendor HCI commands in discrete modules.

mogenson:
> 

Great thanks!!

uael:
> It might be good to put the vendor HCI packet definitions in their own file. Maybe `hci_vendor.py` or `hci_vsc.py`. I’d leave it as low level commands for now. For normal HCI commands, there’s a way to query the controller if the command is supported, but I don’t know of a way to do that for vendor commands (the answer comes in the form of a long bit mask, so not extensible to custom extensions). It’s probably safe to let users decide if they want/need to send those commands, and let them catch errors if they send the command to a controller without knowing if their controller supports it. Since the high level APIs would not send those commands directly, we wouldn’t need a config entry to enable/disable them.

I suggest to put them in their own files like in `bumble/vendor/{zephyr,rtk}.py` and for ex. with realtek use the commands in `bumble/drivers/rtk.py`

barbibulle:
> > It might be good to put the vendor HCI packet definitions in their own file. Maybe `hci_vendor.py` or `hci_vsc.py`. I’d leave it as low level commands for now. For normal HCI commands, there’s a way to query the controller if the command is supported, but I don’t know of a way to do that for vendor commands (the answer comes in the form of a long bit mask, so not extensible to custom extensions). It’s probably safe to let users decide if they want/need to send those commands, and let them catch errors if they send the command to a controller without knowing if their controller supports it. Since the high level APIs would not send those commands directly, we wouldn’t need a config entry to enable/disable them.
> 
> I suggest to put them in their own files like in `bumble/vendor/{zephyr,rtk}.py` and for ex. with realtek use the commands in `bumble/drivers/rtk.py`

PR up for review: #269

mogenson:
@barbibulle @uael This PR is now ready to be reviewed and can be merged.

I was searching for a good way to host the prebuilt hci_usb.zip file as part of the online docs. I ended up creating a discussion thread [here](https://github.com/google/bumble/discussions/274) and uploading the file. It appears that there is a static Github URL that can be include in the mkdocs.

mogenson:
@barbibulle Changed the Zephyr firmware link to point to `downloads/zephyr/hci_usb.zip` from PR #276.

barbibulle (approved):

PR #261, opened 2023-08-28
Rust tools for working with Realtek firmware

marshallpierce:
Further adventures in porting tools to Rust to flesh out the supported API.

These tools didn't feel like `example`s, so I made a top level `bumble` CLI tool that hosts them all as subcommands. I also moved the usb probe not-really-an-`example` into it as well. I'm open to suggestions on how best to organize the subcommands to make them intuitive to explore with `--help`, and how to leave room for other future tools.

I also adopted the per-OS project data dir for a default firmware location so that users can download once and then use those .bin files from anywhere without having to sprinkle .bin files in project directories or reaching inside the python package dir hierarchy.


tools/rtk_util.py
```diff
@@ -61,9 +61,8 @@ async def do_load(usb_transport, force):
         # Get the driver.
         driver = await rtk.Driver.for_host(host, force)
         if driver is None:
-            if not force:
```
marshallpierce, line 64: I think it's erroneous to only return if `not force` -- otherwise, it would then try to `driver.download_firmware()`, which would surely fail if `driver is None`.

rust/Cargo.toml
```diff
@@ -23,7 +23,16 @@ hex = "0.4.3"
 itertools = "0.11.0"
 lazy_static = "1.4.0"
 thiserror = "1.0.41"
+
+# CLI
 anyhow = { version = "1.0.71", optional = true }
```
marshallpierce, line 28: I haven't figured out if it does any harm to users of bumble as a library to have dependencies that are only used by a binary, so until that's sorted out, I'm using a default-enabled feature for these so that library users can turn them off if needed.

marshallpierce:
Here's the current subcommand hierarchy. Is this intuitive? Does it leave the necessary conceptual room for more tools?

```
% ... cg run --bin bumble
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/bumble`
Usage: bumble <COMMAND>

Commands:
  firmware  Manage device firmware
  usb       USB operations
  help      Print this message or the help of the given subcommand(s)
...
```

```
% bumble usb

USB operations

Usage: bumble usb <COMMAND>

Commands:
  probe  Probe the USB bus for Bluetooth devices
  help   Print this message or the help of the given subcommand(s)
```

```
% bumble usb probe --help
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/bumble usb probe --help`
Probe the USB bus for Bluetooth devices

Usage: bumble usb probe [OPTIONS]

Options:
      --verbose  Show additional info for each USB device
```

```
% bumble firmware

Manage device firmware

Usage: bumble firmware <COMMAND>

Commands:
  realtek  Manage Realtek chipset firmware
```

```
% bumble firmware realtek

Manage Realtek chipset firmware

Usage: bumble firmware realtek <COMMAND>

Commands:
  download  Download Realtek firmware
  drop      Drop firmware from a USB device
  info      Show driver info for a USB device
  load      Load firmware onto a USB device
  parse     Parse a firmware file
```

rust/src/wrapper/drivers/rtk.rs
```diff
@@ -0,0 +1,306 @@
...
+            if length == 0 {
+                // report what nom likely would have done, if nom was good at parsing backwards
+                return Err(nom::Err::Error(error::Error::new(
+                    chunk,
+                    error::ErrorKind::Verify,
+                )));
+            }
+            if opcode == 0 && length == 1 {
+                project_id = offset
+                    .checked_sub(1)
```
marshallpierce, line 245: The equivalent logic in python doesn't do this check before reading `offset - 1`, so I think on a suitably mangled file it could read from the header instead of stopping short.

rust/src/wrapper/drivers/rtk.rs
```diff
@@ -0,0 +1,306 @@
...
+/// Realtek firmware file contents
+pub struct Firmware {
+    version: u32,
+    project_id: u8,
+    patches: Vec<Patch>,
+}
+
+impl Firmware {
+    /// Parse a `*_fw.bin` file
+    pub fn parse(input: &[u8]) -> Result<Self, nom::Err<error::Error<&[u8]>>> {
```
marshallpierce, line 151: is it ok to check in realtek firmware files so this can have tests?

rust/src/wrapper/drivers/rtk.rs
```diff
@@ -0,0 +1,306 @@
...
+                    },
+                )(input)
+                .map(|(_rem, output)| output)
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        // look for project id from the end
+        let mut offset = payload.len();
+        let mut project_id: Option<u8> = None;
+        while offset >= 2 {
```
whitevegagabriel, line 226: Does this mean that a project ID can be found within a patch?

rust/src/wrapper/drivers/rtk.rs
```diff
@@ -0,0 +1,306 @@
...
+            // Won't panic, since offset >= 2
+            let chunk = &payload[offset - 2..offset];
+            let length: usize = chunk[0].into();
+            let opcode = chunk[1];
+            offset -= 2;
+
+            if opcode == 0xFF {
+                break;
+            }
+            if length == 0 {
```
whitevegagabriel, line 236: Is this just a courtesy check, since theoretically there may be 0-length opcodes elsewhere?

rust/src/wrapper/drivers/rtk.rs
marshallpierce: I'm assuming that there is some undocumented footer of opcodes after the patches, but I don't know for sure

rust/src/wrapper/drivers/rtk.rs
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    /// ROM id
+    pub fn rom(&self) -> PyResult<u32> {
+        Python::with_gil(|py| self.0.getattr(py, intern!(py, "rom"))?.as_ref(py).extract())
+    }
+}
+
+/// Realtek firmware file contents
+pub struct Firmware {
```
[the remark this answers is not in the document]
marshallpierce: It's not really a wrapper, but it's also certainly part of the coherent concept of a bumble API. I'm not really sure what the right option is. Maybe a separate top level module that's non-public, from which we `pub use` to drop the names into places that would make sense to the user?

rust/src/wrapper/drivers/rtk.rs
marshallpierce: No idea, just something the python parser did.

rust/src/wrapper/drivers/rtk.rs
```diff
@@ -0,0 +1,306 @@
...
+    pub async fn check(host: &Host) -> PyResult<bool> {
+        Python::with_gil(|py| {
+            PyModule::import(py, intern!(py, "bumble.drivers.rtk"))?
+                .getattr(intern!(py, "Driver"))?
+                .call_method1(intern!(py, "check"), (&host.obj,))
+                .and_then(|obj| obj.extract::<bool>())
+        })
+    }
+
+    /// Check if the host has a known driver.
```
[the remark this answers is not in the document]
marshallpierce: Oops!

rust/src/cli/usb/mod.rs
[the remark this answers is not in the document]
marshallpierce: I'm not sure... if people are motivated enough to be looking for rust implementations, they would hopefully have at least run `--help` and seen the `usb` option (and presumably gone to look at the source), right?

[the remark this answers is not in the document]
whitevegagabriel: Yeah, there is definitely a minimum amount of motivation involved there. I see you also made a note in the Readme, so that's probably good enough.

rust/src/wrapper/drivers/rtk.rs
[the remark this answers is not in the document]
whitevegagabriel: I can see that being a solution that bridges that gap nicely.

rust/src/wrapper/drivers/rtk.rs
whitevegagabriel: If the opcodes are to be found after the patches (or not at all), then it might be more correct to only check the post-patch bytes (if present). 

But I don't personally know the firmware format, so don't know what's the most correct.

marshallpierce: yep, we could chop the search down to only the space after patches, but unless we know that's actually valid, I don't think it makes sense to do anything other than blindly follow in the footsteps of the Python impl.

rust/Cargo.toml
marshallpierce: After some experimentation, it looks like the cost is compile time, but if the user enables LTO, there's no binary size impact. Nonetheless, since I expect most users will be using the CLI via a downloaded binary, it's not much of a hassle to make the feature non-default so that lib users aren't affected at all.

rust/src/wrapper/drivers/rtk.rs
marshallpierce: This code got moved; it's now at https://github.com/google/bumble/pull/261/files#diff-1a70751eee41dc84f6ba4d3247b476d244e914ad403c219c88d4c154f9b0c991R103.

whitevegagabriel (approved):
Approval with the assumption that the rtk firmware parsing in Python is also correct (aside from the unchecked sub when getting `project_id`)

rust/src/cli/firmware/realtek.rs
```diff
@@ -0,0 +1,265 @@
+// Copyright 2023 Google LLC
```
barbibulle, line 1: maybe this could be named `rtk.rs` to be consistent with `drivers/rtk.rs` ?

rust/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+        transport: String,
+    },
+    /// Show driver info for a USB device
+    Info {
+        /// Bumble transport spec. Must be for a USB device.
+        ///
+        /// <https://google.github.io/bumble/transports/index.html>
+        #[arg(long)]
+        transport: String,
+        /// Try to resolve driver info even if USB info is not available.
```
barbibulle, line 109: or the USB (vendor,product) tuple is not in the list of known compatible RTK USB dongles.

barbibulle (approved):

bumble/drivers/__init__.py
```diff
@@ -66,3 +69,21 @@ async def get_driver_for_host(host):
         return driver
 
     return None
+
+
+def project_data_dir() -> pathlib.Path:
+    """
+    :return: A path to an OS-specific directory for bumble data.
```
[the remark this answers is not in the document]
marshallpierce: Done

rust/src/cli/firmware/realtek.rs
marshallpierce: Done

rust/src/main.rs
marshallpierce: Done

rust/src/wrapper/drivers/rtk.rs
barbibulle: The python implementation just mirrors the Linux driver implementation, where looking for the project ID is here: https://github.com/torvalds/linux/blob/99d99825fc075fd24b60cc9cf0fb1e20b9c16b0f/drivers/bluetooth/btrtl.c#L677
And you're right that the search for opcodes should stop at the end of the patch block. I think the C implementation takes the shortcut of assuming that there will always be an opcode list with at least an `end` opcode, so it cheats a bit by not calculating the size of the patch block. There is still a check to stop at the end of the header, because that's easy, fixed size, and that ensures the search doesn't get out of bounds of the firmware buffer at least. But a malformed file with no `end` or `project_id` opcode after the patch block could lead to incorrectly finding a lookalike inside the patch block.

barbibulle (approved):

PR #268, opened 2023-09-07
Raise an exception when an L2cap connection fails

benquike:


bumble/l2cap.py
```diff
@@ -2078,7 +2078,8 @@ async def connect(self, connection: Connection, psm: int) -> Channel:
         # Connect
         try:
             await channel.connect()
-        except Exception:
+        except Exception as e:
```
barbibulle, line 2081: maybe use  `try: xxx finally: del yyy` ?

barbibulle (approved):

bumble/l2cap.py
benquike: moving del to finally would make it wrong, right?
we only want to delete when exception happens?

uael: @benquike No you could move the delete to finally 👌

uael: I miss-read it!

Issue #266, opened 2023-09-06
Cleanup hfp.py


zxzxwu:
Some typos and type mismatching.

PR #270, opened 2023-09-07
Fix typos

zxzxwu:


zxzxwu:
Shall we also add Frontend checking?

zxzxwu:
Fixing https://github.com/google/bumble/issues/266

bumble/hfp.py
```diff
@@ -390,7 +391,7 @@ async def execute_command(
                 )
                 if result.code == 'OK':
                     if response_type == AtResponseType.SINGLE and len(responses) != 1:
-                        raise ProtocolError("NO ANSWER")
+                        raise InvalidStateError("NO ANSWER")
```
barbibulle, line 394: I think `ProtocolError` is more appropriate here. An `InvalidStateError` is usually more representative of a condition like trying to close a stream that's not open. a `ProtocolError` reflects an error in an exchange between two entities, where the syntax or sequence of messages isn't what's expected or allowed, which seems to be the case here.

zxzxwu: I extend BaseError to show non-digital errors.

barbibulle (approved):

PR #271, opened 2023-09-07
Typing transport and relateds.

zxzxwu:


zxzxwu:
Is it a good idea to make all sink/source inherited from a base class? Or a typing protocol is enough?

barbibulle:
> Is it a good idea to make all sink/source inherited from a base class? Or a typing protocol is enough?

I would prefer a `Protocol` based approach. The sinks and source concepts are very much intended for a Duck Typing approach. Definining a `Sink` and `Source` `Protocol` type rather than base classes would be great.

bumble/transport/file.py
```diff
@@ -28,7 +30,31 @@
 
 
 # -----------------------------------------------------------------------------
-async def open_file_transport(spec):
+class FileTransport(Transport):
```
barbibulle, line 33: It is simpler and more encapsulated to have `FileTransport` as a local class in `open_file_transport`. That class isn't really meant to be used by itself, it is just a local helper to tie the source and sink together into a `Transport` object. `open_file_transport` should be typed to return just a `Transport`, without having to reveal its internal details.
The same applies to all the other `open_xxx_transport` functions.

bumble/transport/tcp_server.py
```diff
@@ -15,19 +15,25 @@
...
-from .common import Transport, StreamPacketSource
+from .common import Transport, StreamPacketSource, BaseSink
 
 # -----------------------------------------------------------------------------
 # Logging
 # -----------------------------------------------------------------------------
 logger = logging.getLogger(__name__)
 
 
+class TcpServerTransport(Transport):
```
barbibulle, line 30: Same remark as above: this should remain a function-local helper class, we don't want to leak the internal details of the implementation outside of the constructor.

bumble/transport/ws_client.py
```diff
@@ -27,7 +27,7 @@
 
 
 # -----------------------------------------------------------------------------
-async def open_ws_client_transport(spec):
+async def open_ws_client_transport(spec: str) -> PumpedTransport:
```
barbibulle, line 30: Prefer returning just `Transport` to avoid leaking internal details.

bumble/transport/__init__.py
```diff
@@ -114,12 +113,14 @@ async def _open_transport(name: str) -> Transport:
     if scheme == 'pty':
         from .pty import open_pty_transport
 
-        return await open_pty_transport(spec[0] if spec else None)
+        assert spec is not None
```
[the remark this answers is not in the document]
zxzxwu: Done

bumble/transport/file.py
zxzxwu: Done

bumble/transport/ws_client.py
zxzxwu: Done

bumble/host.py
```diff
@@ -21,7 +21,8 @@
 import logging
 import struct
 
-from typing import Optional
+from collections.abc import Mapping, Callable, Awaitable
```
barbibulle, line 24: `Callable` and `Awaitable` now exist in the `typing` module. `Mapping` exists there too, but maybe we can use the more direct `Dict` type here? (I think the rule of thumb for `Dict` vs `Mapping` is to use `Mapping` for function arguments if possibe (to make it more flexible for the caller), but use `Dict` for return parameters or class members, to make it more precise.

bumble/host.py
```diff
@@ -116,10 +121,21 @@ def on_acl_pdu(self, pdu: bytes) -> None:
 
 # -----------------------------------------------------------------------------
 class Host(AbortableEventEmitter):
-    def __init__(self, controller_source=None, controller_sink=None):
+    connections: Mapping[int, Connection]
```
barbibulle, line 124: suggest: `Dict`

bumble/transport/android_emulator.py
```diff
@@ -33,7 +33,7 @@
 
 
 # -----------------------------------------------------------------------------
-async def open_android_emulator_transport(spec):
+async def open_android_emulator_transport(spec: str | None) -> PumpedTransport:
```
barbibulle, line 36: suggest: returning just `Transport` to allow the internal type to change in the future without affecting callers.

bumble/transport/common.py
```diff
@@ -20,11 +20,13 @@
 import struct
 import asyncio
 import logging
-from typing import ContextManager
+import io
+from collections.abc import Mapping
```
barbibulle, line 24: use`typing` package

bumble/transport/common.py
```diff
@@ -36,39 +38,54 @@
 # Information needed to parse HCI packets with a generic parser:
 # For each packet type, the info represents:
 # (length-size, length-offset, unpack-type)
-HCI_PACKET_INFO = {
+HCI_PACKET_INFO: Mapping[int, Tuple[int, int, str]] = {
```
barbibulle, line 41: suggest: `Dict`

bumble/transport/common.py
```diff
@@ -36,39 +38,54 @@
...
+# -----------------------------------------------------------------------------
+# Typing Protocols
+# -----------------------------------------------------------------------------
+class TransportSink(Protocol):
+    def on_packet(self, packet: bytes) -> None:
+        ...
+
+
+class TransportSource(Protocol):
+    def __init__(self) -> None:
```
barbibulle, line 67: Is this necessary as part of the method signature? What if you had a `TransportSource` implementation that has an `__init__` method with constructor arguments?

barbibulle (approved):

bumble/host.py
zxzxwu: Done

bumble/host.py
zxzxwu: Done

bumble/transport/common.py
zxzxwu: Done

bumble/transport/common.py
zxzxwu: Done

bumble/transport/common.py
zxzxwu: Done

bumble/transport/android_emulator.py
zxzxwu: Done

PR #272, opened 2023-09-07
Bring HfpProtocol back

zxzxwu:
There are still many users using the original protocols. Though we are still in an early stage, deprecation should be slowly and responsible. Also, there is no alternative for AG and low-level usage.

### DIFF
--- a/bumble/hfp.py
+++ b/bumble/hfp.py
@@ -15,16 +15,19 @@
 # -----------------------------------------------------------------------------
 # Imports
 # -----------------------------------------------------------------------------
+import collections.abc
 import logging
 import asyncio
 import dataclasses
 import enum
 import traceback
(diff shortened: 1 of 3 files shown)


barbibulle (approved):
It's good to have a way to support the AG mode. I think we should try to merge the lower-level functionalities of HfpProtocol in a way that can be shared with the HfProtocol class here, in a future PR.

hchataing (approved):
I was planning on working on the AG side as well by have been delayed by other tasks ...

zxzxwu:
Yes, from my perspective `HfProtocol` is more like a servicer above the protocol. Though I marked the protocol deprecate here, I would prefer HF&AG wrappers working over `HfpProtocol`.

PR #275, opened 2023-09-08
Add license header check for rust files

whitevegagabriel:
Added binary that can check for and add Apache 2.0 licenses. Run this binary during the build-rust workflow.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -24,6 +24,10 @@ itertools = "0.11.0"
 lazy_static = "1.4.0"
 thiserror = "1.0.41"
 
+# Dev tools
+file-header = { version = "0.1.2", optional = true }
+globset = { version = "0.4.13", optional = true }
+
 # CLI
 anyhow = { version = "1.0.71", optional = true }
 clap = { version = "4.3.3", features = ["derive"], optional = true }
@@ -53,9 +57,14 @@ env_logger = "0.10.0"
 rustdoc-args = ["--generate-link-to-definition"]
 
 [[bin]]
+name = "file-header"
+path = "tools/file_header.rs"
+required-features = ["dev-tools"]
+
+[[bin]]
 name = "gen-assigned-numbers"
 path = "tools/gen_assigned_numbers.rs"
-required-features = ["bumble-codegen"]
+required-features = ["dev-tools"]
 
 [[bin]]
 name = "bumble"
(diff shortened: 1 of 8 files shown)


rust/Cargo.toml
```diff
@@ -24,6 +24,12 @@ itertools = "0.11.0"
 lazy_static = "1.4.0"
 thiserror = "1.0.41"
 
+# Dev tools
+file-header = { version = "0.1.1", optional = true }
+globset = { version = "0.4.13", optional = true }
+# anyhow
+# clap
```
marshallpierce, line 31: leftovers?

rust/tools/file_header.rs
```diff
@@ -0,0 +1,71 @@
...
+use std::{env, path::PathBuf};
+
+fn main() -> anyhow::Result<()> {
+    let rust_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR")?);
+    let ignore_globset = ignore_globset()?;
+    // Note: when adding headers, there is a bug where the line spacing is off for Apache 2.0 (see https://github.com/spdx/license-list-XML/issues/2127)
+    let header = APACHE_2_0.build_header(YearCopyrightOwnerValue::new(2023, "Google LLC".into()));
+
+    let cli = Cli::parse();
+
```
marshallpierce, line 30: rustfmt appears to be upset here

rust/tools/file_header.rs
```diff
@@ -0,0 +1,71 @@
...
+    APACHE_2_0, YearCopyrightOwnerValue
+}};
+use globset::{Glob, GlobSet, GlobSetBuilder};
+use std::{env, path::PathBuf};
+
+fn main() -> anyhow::Result<()> {
+    let rust_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR")?);
+    let ignore_globset = ignore_globset()?;
+    // Note: when adding headers, there is a bug where the line spacing is off for Apache 2.0 (see https://github.com/spdx/license-list-XML/issues/2127)
+    let header = APACHE_2_0.build_header(YearCopyrightOwnerValue::new(2023, "Google LLC".into()));
```
barbibulle, line 27: will that not conflict with headers that have a date range for the copyright? (like `Copyright 2021-2022 Google LLC`)?

marshallpierce: The search pattern used to check for an existing license is just the date-free ASL2 header, so it shouldn't care about date ranges: https://github.com/google/file-header/blob/main/src/license/spdx/mod.rs#L327.
However, that does raise the question of perhaps we should be using the current date's year when _adding_ licenses 🤔

rust/Cargo.toml
whitevegagabriel: I wanted to signify that these dependencies are used by Dev tools but declared elsewhere. Though I guess its not important. If the dependencies are deleted, the compiler will complain.

rust/tools/file_header.rs
whitevegagabriel: Ok, so for now we can probably keep as-is.

rust/Cargo.toml
whitevegagabriel: Removed.

whitevegagabriel:
There was an issue in a dependency, so waiting for that to be resolved before I continue here.

rust/tools/file_header.rs
whitevegagabriel: Should be fixed now

whitevegagabriel:
Ok, issue in dependency has been fixed.

barbibulle:
Ready to mege?

whitevegagabriel:
Ready for merge 👍

barbibulle (approved):

PR #277, opened 2023-09-09
Add connection parameter update from peripheral

wescande:


uael (approved):
Sgtm !

barbibulle (commented):
Don’t merge this yet, I'd like to suggest a simple enhancement. I need to describe the details, which I will do in a following note (I don’t have the time today).

bumble/l2cap.py
```diff
@@ -1865,11 +1867,45 @@ def on_l2cap_connection_parameter_update_request(
                 ),
             )
 
+    async def update_connection_parameters(
```
barbibulle, line 1870: I think it would be good to also expose this functionality via the existing `Device.update_connection_parameters` (and indirectly via `Connection.update_parameters`). Maybe adding a `use_l2cap` bool parameters, with default to `False` (and checking that we're in the peripheral role when that parameter is `True`).
We can keep this function here and delegate to it from the `Device` API.

barbibulle (approved):

bumble/l2cap.py
wescande: Done. Please let me know if there is room for improvement

bumble/device.py
```diff
@@ -2110,11 +2110,29 @@ async def update_connection_parameters(
         supervision_timeout,
         min_ce_length=0,
         max_ce_length=0,
-    ):
+        use_l2cap=False,
```
barbibulle, line 2113: Would be good to also add this parameter to the `Connection.update_parameters` method that calls this.

bumble/device.py
```diff
@@ -2110,11 +2110,29 @@ async def update_connection_parameters(
...
+    ) -> bool:
```
barbibulle, line 2114: I don't think there's a need to return a boolean here. This function either just returns (None), or raises an exception.

bumble/device.py
```diff
@@ -2110,11 +2110,29 @@ async def update_connection_parameters(
...
+            l2cap_result = (
+                await self.l2cap_channel_manager.update_connection_parameters(
+                    connection,
+                    connection_interval_min,
+                    connection_interval_max,
+                    max_latency,
+                    supervision_timeout,
+                )
+            )
+            return l2cap_result == 0
```
barbibulle, line 2134: This should raise an error (`HCI_StatusError(l2cap_result)`) if the result is not HCI_SUCCESS, just like the non-l2cap version below.

bumble/device.py
```diff
@@ -2124,10 +2142,11 @@ async def update_connection_parameters(
                 supervision_timeout=supervision_timeout,
                 min_ce_length=min_ce_length,
                 max_ce_length=max_ce_length,
-            )
+            )  # type: ignore[call-arg]
         )
         if result.status != HCI_Command_Status_Event.PENDING:
             raise HCI_StatusError(result)
+        return True
```
barbibulle, line 2149: should return `None` (see earlier comment)

bumble/device.py
wescande: Now raising `ConnectionParameterUpdateError` defined in core.py

bumble/device.py
```diff
@@ -2110,11 +2113,30 @@ async def update_connection_parameters(
...
+            l2cap_result = (
+                await self.l2cap_channel_manager.update_connection_parameters(
+                    connection,
+                    connection_interval_min,
+                    connection_interval_max,
+                    max_latency,
+                    supervision_timeout,
+                )
+            )
+            if l2cap_result != 0:
```
barbibulle, line 2137: nit: use the `L2CAP_CONNECTION_PARAMETERS_ACCEPTED_RESULT` constant.

barbibulle (approved):

PR #278, opened 2023-09-09
Typing GATT

zxzxwu:


bumble/att.py
```diff
@@ -720,47 +722,67 @@ class ATT_Handle_Value_Confirmation(ATT_PDU):
...
-                f"Attribute::permissions error:\nExpected a string containing any of the keys, separated by commas: {','.join(Attribute.PERMISSION_NAMES.values())}\nGot: {permissions_str}"
-            ) from exc
+    def write(self, connection, value: bytes) -> None:
+        ...
 
-    def __init__(self, attribute_type, permissions, value=b''):
+
+# -----------------------------------------------------------------------------
+class Attribute(EventEmitter):
+    class Permission(enum.IntFlag):
```
barbibulle, line 735: Since this type consists of a combination of permissions, I suggest we name if `Permissions` (with an `s).

bumble/att.py
```diff
@@ -720,47 +722,67 @@ class ATT_Handle_Value_Confirmation(ATT_PDU):
...
+        WRITE_REQUIRES_AUTHENTICATION = 0x20
+        READ_REQUIRES_AUTHORIZATION = 0x40
+        WRITE_REQUIRES_AUTHORIZATION = 0x80
+
+        @classmethod
+        def from_string(cls, permissions_str: str) -> Attribute.Permission:
+            try:
+                return functools.reduce(
+                    lambda x, y: x | Attribute.Permission[y],
+                    permissions_str.split(","),
```
barbibulle, line 750: It might be nice to also allow `|` as a separator, in addition to `,`.

bumble/att.py
```diff
@@ -778,22 +800,26 @@ def __init__(self, attribute_type, permissions, value=b''):
...
-    def encode_value(self, value):
+    def encode_value(self, value: bytes) -> bytes:
         return value
 
-    def decode_value(self, value_bytes):
+    def decode_value(self, value_bytes: bytes) -> bytes:
         return value_bytes
 
-    def read_value(self, connection: Connection):
+    def read_value(self, connection: Connection | None) -> bytes:
```
barbibulle, line 809: the `|` symbol for unions is only for Python >= 3.10, but we still support >= 3.8, so we should still use `Union[]` here.

bumble/gatt.py
```diff
@@ -187,7 +187,7 @@
 # -----------------------------------------------------------------------------
 
 
-def show_services(services):
+def show_services(services: List[Service]) -> None:
```
barbibulle, line 190: since we only iterate over the services here, we could be more flexible an accept `Iterable[Service]` or `Sequence[Service]`

bumble/gatt_client.py
```diff
@@ -112,13 +126,13 @@ class ServiceProxy(AttributeProxy):
     characteristics: List[CharacteristicProxy]
     included_services: List[ServiceProxy]
 
-    @staticmethod
-    def from_client(service_class, client, service_uuid):
+    @classmethod
```
barbibulle, line 129: This change might not always work: the `service_class` passed here may not necessarily be `ServiceProxy` or a subclass.

bumble/gatt_client.py
```diff
@@ -78,16 +92,16 @@
...
+        self, client: Client, handle: int, end_group_handle: int, attribute_type: UUID
+    ) -> None:
         EventEmitter.__init__(self)
         self.client = client
         self.handle = handle
         self.end_group_handle = end_group_handle
         self.type = attribute_type
 
-    async def read_value(self, no_long_read=False):
+    async def read_value(self, no_long_read: bool = False) -> bytes:
```
barbibulle, line 104: Because of the polymorphic nature of `decode`, the return type here should really be `Any`, because `AttributeProxy` is a generic superclass, characteristic adapters may decode byte arrays into any type (profile dependent). It's not easy to see it, because that happens though the CharacteristicAdapter class that does some of the patching at runtime (which is always confusing for the type checker).

bumble/gatt_client.py
```diff
@@ -97,13 +111,13 @@ async def write_value(self, value, with_response=False):
             self.handle, self.encode_value(value), with_response
         )
 
-    def encode_value(self, value):
+    def encode_value(self, value: bytes) -> bytes:
```
barbibulle, line 114: Similar to `read_value`, this should be generic, with `value: Any`.

barbibulle (approved):

bumble/gatt_client.py
```diff
@@ -216,8 +232,10 @@ class ProfileServiceProxy:
     Base class for profile-specific service proxies
     '''
 
+    SERVICE_CLASS: type[TemplateService]
```
barbibulle, line 235: using lowercase `type[]` is only for Python >= 3.9. For Python 3.8, we still need to use `Type[]`.

barbibulle (approved):

PR #289, opened 2023-09-14
Refactor L2CAP API

zxzxwu:


zxzxwu:
I would like to keep the name `register_l2cap_server()` and `connect()`, but appending more arguments.

zxzxwu:
Need some reviews to decide if I should go further

bumble/l2cap.py
```diff
@@ -2095,7 +2155,34 @@ async def open_le_coc(
 
         return channel
 
-    async def connect(self, connection: Connection, psm: int) -> Channel:
+    async def connect(
```
barbibulle, line 2158: Having a single method here to dispatch to `_open_le_credit_based_channel` or `_open_classic_channel` depending on `mode` is a bit misleading, because not all parameters apply to all modes. Having a separate function for each mode is cleaner, as there are not extra/unused arguments. Also, calling `connect` for a connectionless channel is surprising.  Maybe `Device.connect_l2cap_le_credit_based_channel()` for LE COC, and `Device.open_l2cap_channel()` for non-connection-oriented channels?

bumble/l2cap.py
```diff
@@ -1491,7 +1500,49 @@ def deregister_fixed_channel(self, cid: int) -> None:
         if cid in self.fixed_channels:
             del self.fixed_channels[cid]
 
-    def register_server(self, psm: int, server: Callable[[Channel], Any]) -> int:
+    def register_server(
```
barbibulle, line 1503: similar comment as for the open/connect comment: having  `max_credits`, `mtu` and `mps` argument for non-connection non-credit channels is misleading. Having two functions would be clearer. Maybe we can name them `Device.register_l2cap_server` and `Device.register_l2cap_le_credit_based_server` to keep a naming convention consistent with `Device.connect_l2cap_le_credit_based_channel()`  and `Device.open_l2cap_channel()`.  At the `l2cap.py` level, we can drop the `l2cap` part of the name and have just `register_server` and `register_l2_credit_based_server`.

bumble/device.py
```diff
@@ -1181,12 +1191,19 @@ def find_connection_by_bd_addr(
         return None
 
     def create_l2cap_connector(self, connection, psm):
+        warnings.warn('Please use l2cap_channel_manager.connect', DeprecationWarning)
         return lambda: self.l2cap_channel_manager.connect(connection, psm)
 
     def create_l2cap_registrar(self, psm):
+        warnings.warn(
```
barbibulle, line 1198: I think we can just get rid of this, it is only used in one place internally, and I doubt there are many external callers.

bumble/l2cap.py
zxzxwu: The problem here is that it makes function name very long, and will require a 2(for client/server) * 2(for device/l2cap) of public methods for each mode which is a little bit fat. Also, it is not an anti-pattern to ignore some parameters. For example, `asyncio.open_connection()` has SSL parameters but they will be ignored in pure TCP mode.

bumble/device.py
```diff
@@ -1181,12 +1191,19 @@ def find_connection_by_bd_addr(
         return None
 
     def create_l2cap_connector(self, connection, psm):
+        warnings.warn('Please use l2cap_channel_manager.connect', DeprecationWarning)
```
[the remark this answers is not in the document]
zxzxwu: Got it!

bumble/l2cap.py
```diff
@@ -1342,6 +1345,12 @@ def __str__(self) -> str:
         )
 
 
+# -----------------------------------------------------------------------------
+# Typing
+# -----------------------------------------------------------------------------
+ChannelMode = TypeVar('ChannelMode', Channel, LeConnectionOrientedChannel)
```
barbibulle, line 1351: Maybe we should use this refactor as an opportunity to rename `LeConnectionOrientedChannel` to `LeCreditBasedChannel`?

bumble/l2cap.py
zxzxwu: In that case, would it be better to saperate `open_channel()` and `connect()`, where `open_channel()` resides in both `device` and `l2cap` only returning an object from a type, while `connect()` only resides in l2cap, taking mode-specific parameters?
Example:
```py
# device.py

def open_l2cap_channel(mode: ...) -> l2cap.Channel:
  ...

# l2cap.py
class ClassicBasicChannel:
  ...
  def connect(psm: int):
    ...

class LeCreditBasedChannel:
  ...
  def connect(psm: int, credit, mtu, mps):
    ...

def open_channel(mode: ...) -> l2cap.Channel:
  ...

```

bumble/l2cap.py
zxzxwu: Yes that's the next step

bumble/l2cap.py
zxzxwu: For server it will be a little bit complicated, because the API doesn't return any object.

zxzxwu: Or with a new proxy...
```py
class ClassicBasicChannelServer:
    def __init__(self, psm: int) -> None:
        self.psm = psm

    async def serve(self, handler: Callable[[Channel], Any]) -> None:
        ...


class LeCreditBasedChannelServer:
    def __init__(self, psm: int) -> None:
        self.psm = psm

    async def serve(
        self, handler: Callable[[LeConnectionOrientedChannel], Any], credit, mtu, mps
    ) -> None:
        ...

ServerMode = TypeVar('ServerMode', ClassicBasicChannelServer, LeCreditBasedChannelServer)

class ChannelManager:
    def open_server(mode: ServerMode) -> ServerMode:
        ...

```

barbibulle (approved):
Nice work!

zxzxwu:
Well somehow Rust raises an inrelated error

uael (approved):
Good job Josh!

PR #292, opened 2023-09-19
add pyee group util

barbibulle:
Implementation of an idea by @zxzxwu 
Allows registering event handlers as a group (they get removed as a group when any event fires)

### DIFF
--- a/bumble/utils.py
+++ b/bumble/utils.py
@@ -19,9 +19,19 @@ import asyncio
 import logging
 import traceback
 import collections
+from contextlib import ExitStack
 import sys
-from typing import Awaitable, Set, TypeVar
-from functools import wraps
+from typing import (
+    Awaitable,
+    Callable,
+    ContextManager,
+    Mapping,
+    Optional,
+    Set,
+    Tuple,
+    TypeVar,
+)
+import functools
 from pyee import EventEmitter
 
 from .colors import color
@@ -167,7 +177,7 @@ class AsyncRunner:
         """
 
         def decorator(func):
-            @wraps(func)
+            @functools.wraps(func)
             def wrapper(*args, **kwargs):
                 coroutine = func(*args, **kwargs)
                 if queue is None:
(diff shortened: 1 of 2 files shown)


zxzxwu:
For reference, here is my implementation

```py

_F = TypeVar('_F', bound=Callable)


class EventWatcherContext(object):
    handlers: List[Tuple[EventEmitter, str, Callable[..., Any]]]

    def __init__(self) -> None:
        self.handlers = []

    def on(
        self, emitter: EventEmitter, event: str
    ) -> Callable[[_F], _F]:
        def wrapper(f: _F):
            self.handlers.append((emitter, event, f))
            emitter.on(event, f)

        return wrapper

    def once(
        self, emitter: EventEmitter, event: str
    ) -> Callable[[_F], _F]:
        def wrapper(f: _F):
            self.handlers.append((emitter, event, f))
            emitter.once(event, f)

        return wrapper

    def __enter__(self) -> EventWatcherContext:
        return self

    def __exit__(self, type, value, traceback) -> Any:
        for emitter, event, handler in self.handlers:
            emitter.remove_listener(event, handler)
#------------------------
# usage

                with EventWatcherContext() as context:

                    @context.on(connection, 'pairing')
                    def on_pairing(*_: Any) -> None:
                        security_result.set_result('success')

                    @context.on(connection, 'pairing_failure')
                    def on_pairing_failure(*_: Any) -> None:
                        security_result.set_result('pairing_failure')

                    @context.on(connection, 'disconnection')
                    def on_disconnection(*_: Any) -> None:
                        security_result.set_result('connection_died')


```

I found some advantages here:
1. Can register multiple handlers on the same (emitter, event) pair
2. Can be implemented as decorator
3. Handler mapping is not very type-friendly

zxzxwu:
And your proposal can be implemented as well:
```py

def once_for_group(
    emitter: EventEmitter, handlers: Mapping[str, Callable[..., Any]]
) -> None:
    # Just as a container
    context = EventWatcherContext()

    def on_event(handler: Callable[..., Any], *args, **kwargs) -> None:
        handler(*args, **kwargs)
        context.remove_all_listeners()

    for event, handler in handlers.items():
        context.once(emitter, event, partial(on_event, handler))
```

barbibulle:
Yes, that makes sense. I now understand better what you were looking for, and I was shooting for a different goal (auto-remove of "once" handlers when one of them is invoked). As you point out, that behavior can be layered on top of the context manager approach, so we can start from your version.

PR #294, opened 2023-09-20
Make cryptography a valid dependency for emscripten targets

### LINKED ISSUE #284: Loading in Pyodide fails due to missing `cryptography` package
We are loading bumble in Pyodide in a browser, and running into an issue where it says the `cryptography` package is not installed.

The installation code for bumble looks like this"

```ts
await pyodide.loadPackage('micropip');
await pyodide.runPythonAsync('import micropip; await micropip.install("bumble")')
```

And the error message is

```
PythonError: Traceback (most recent call last):
  File "/lib/python311.zip/_pyodide/_base.py", line 540, in eval_code_async
    await CodeRunner(
  File "/lib/python311.zip/_pyodide/_base.py", line 363, in run_async
    coroutine = eval(self.code, globals, locals)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<exec>", line 22, in <module>
  File "/lib/python3.11/site-packages/bumble/device.py", line 158, in <module>
    from .pairing import PairingConfig
  File "/lib/python3.11/site-packages/bumble/pairing.py", line 28, in <module>
    from .smp import (
  File "/lib/python3.11/site-packages/bumble/smp.py", line 59, in <module>
    from . import crypto
  File "/lib/python3.11/site-packages/bumble/crypto.py", line 28, in <module>
    from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
ModuleNotFoundError: The module 'cryptography' is included in the Pyodide distribution, but it is not installed.
You can install it by calling:
  await micropip.install("cryptography") in Python, or
  await pyodide.loadPackage("cryptography") in JavaScript
See https://pyodide.org/en/stable/usage/loading-packages.html for more details.
```

I can easily add the `micropip.install("cryptography")` call, but I am wondering if `setup.cfg` can be updated to avoid the issue in the first place?

https://github.com/google/bumble/blob/5af347b499b26c2d0431f2a00a3c4f44290e6ec2/setup.cfg#L38

Thank you for your time!

mauricelam:
Since only the special cryptography package bundled with pyodide can be used, relax the version requirement to anything that's version `39.*`.

Fix #284

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ install_requires =
     bt-test-interfaces >= 0.0.2; platform_system!='Emscripten'
     click == 8.1.3; platform_system!='Emscripten'
     cryptography == 39; platform_system!='Emscripten'
-    # Pyodide has bundles a version of cryptography that is built for wasm, which may not match the
+    # Pyodide bundles a version of cryptography that is built for wasm, which may not match the
     # versions available on PyPI. Relax the version requirement since it's better than being
     # completely unable to import the package in case of version mismatch.
     cryptography >= 39.0; platform_system=='Emscripten'
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ install_requires =
     appdirs >= 1.4; platform_system!='Emscripten'
     bt-test-interfaces >= 0.0.2; platform_system!='Emscripten'
     click == 8.1.3; platform_system!='Emscripten'
-    cryptography == 39; platform_system!='Emscripten'
+    cryptography == 39.*
     grpcio == 1.57.0; platform_system!='Emscripten'
     humanize >= 4.6.0; platform_system!='Emscripten'
     libusb1 >= 2.0.1; platform_system!='Emscripten'
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,11 @@ install_requires =
     appdirs >= 1.4; platform_system!='Emscripten'
     bt-test-interfaces >= 0.0.2; platform_system!='Emscripten'
     click == 8.1.3; platform_system!='Emscripten'
-    cryptography == 39.*
+    cryptography == 39; platform_system!='Emscripten'
+    # Pyodide has bundles a version of cryptography that is built for wasm, which may not match the
+    # versions available on PyPI. Relax the version requirement since it's better than being
+    # completely unable to import the package in case of version mismatch.
+    cryptography >= 39.0; platform_system=='Emscripten'
     grpcio == 1.57.0; platform_system!='Emscripten'
     humanize >= 4.6.0; platform_system!='Emscripten'
     libusb1 >= 2.0.1; platform_system!='Emscripten'
--- a/web/bumble.js
+++ b/web/bumble.js
@@ -74,7 +74,6 @@ export async function loadBumble(pyodide, bumblePackage) {
     await pyodide.loadPackage("micropip");
     await pyodide.runPythonAsync(`
         import micropip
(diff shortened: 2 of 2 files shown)


setup.cfg
```diff
@@ -35,7 +35,7 @@ install_requires =
     appdirs >= 1.4; platform_system!='Emscripten'
     bt-test-interfaces >= 0.0.2; platform_system!='Emscripten'
     click == 8.1.3; platform_system!='Emscripten'
-    cryptography == 39; platform_system!='Emscripten'
```
barbibulle, line 38: I don't think that's going to work. As far as I can tell, Pyiodide only supported pure-python packages when loaded as package dependencies. This is why `cryptography` is excluded here for `Emscripten`, and manually loaded in `bumble.js` (the manual loading way supports loading built-in packages)

mauricelam: The micropip folks recommended doing this in https://github.com/pyodide/micropip/issues/84. I tested the change with `inv build`, host a static server at the bumble directory, and then navigated to `http://localhost:8000/web/scanner/scanner.html?package=/dist/bumble-0.0.170.dev2%2bg98bf77d.d20230920-py3-none-any.whl` (the last part replaced with the whl file name printed out in `inv build`), and that worked.

(I also locally added `from bumble import crypto` to make sure the import is triggered)

Is there additional testing I should do to verify against the issues you ran into earlier?

barbibulle: Ah, good to know! 
The next question we need to answer is: how will pyodide manage its package versions. For now, pyodide comes with `cryptography == 39.x`, but that will evolve. What would be a good strategy to manage compatibility for static pages that load pyodide and want to use the latest Bumble release? Would we need to always pin the version of the Bumble package that the page wants to load, to ensure that the version of pyodide it loads comes with a compatible `cryptography` version?

barbibulle: Maybe we should allow a wider range of `cryptography` versions, to allow loading older, compatible versions, yet allow newer versions to be used (from PyPI), even if not yet present in Pyiodide's builtin package list?

mauricelam: Good questions. For the first question, I think pinning both the bumble version and the pyodide version would be a good recommendation to ensure compatibility. I wish there is a way to have more relaxed version specification during development that gets frozen when the static HTML is built, but that build step doesn't always exist in general.

For the second question, maybe it makes sense for bumble to be optimistic and make the version `>= 39.0; platform_system=='Emscripten'`, that way it would attempt to use any version of `cryptography` that Pyodide happens to provide us with, the argument being that if there are API breaking changes, it's not worse than failing with "cannot find compatible version of `cryptography`". Having `=='Emscripten'` listed separately also allows other platforms to update its dependency version independently of when Pyodide updates. How does that sound?

barbibulle: That sounds good. `cryptography` has been very good with API stability, so I'm not worried about having a loose version match there.

mauricelam: Done in the new commit

barbibulle (approved):

setup.cfg
```diff
@@ -35,7 +35,11 @@ install_requires =
     appdirs >= 1.4; platform_system!='Emscripten'
     bt-test-interfaces >= 0.0.2; platform_system!='Emscripten'
     click == 8.1.3; platform_system!='Emscripten'
-    cryptography == 39.*
+    cryptography == 39; platform_system!='Emscripten'
+    # Pyodide has bundles a version of cryptography that is built for wasm, which may not match the
```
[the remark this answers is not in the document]
mauricelam: Done

PR #295, opened 2023-09-21
Bumble hid

SneKarnataki:
Bumble HID Host implementation.


bumble/hid.py
```diff
@@ -0,0 +1,246 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# -----------------------------------------------------------------------------
+# Imports
+# -----------------------------------------------------------------------------
+import logging
```
zxzxwu, line 18: `from __future__ import annotation`

bumble/hid.py
```diff
@@ -0,0 +1,246 @@
...
+# Control Operations
+HID_SUSPEND              = 0x03
+HID_EXIT_SUSPEND         = 0x04
+HID_VIRTUAL_CABLE_UNPLUG = 0x05
+
+# -----------------------------------------------------------------------------
+class HIDHost(EventEmitter):
+    l2cap_channel: Optional[l2cap.Channel]
+
+    def __init__(self, device, connection) -> None:
```
zxzxwu, line 78: type hint for arguments

bumble/hid.py
```diff
@@ -0,0 +1,246 @@
...
+            )
+        except ProtocolError as error:
+            logger.error(f'L2CAP connection failed: {error}')
+            raise
+
+        assert self.l2cap_intr_channel is not None
+        # Become a sink for the L2CAP channel
+        self.l2cap_intr_channel.sink = self.on_intr_pdu
+
+    async def interrupt_channel_disconnect(self):
```
zxzxwu, line 117: Return types

bumble/hid.py
```diff
@@ -0,0 +1,246 @@
...
+        await self.l2cap_intr_channel.disconnect()  # type: ignore
```
zxzxwu, line 118:
```py
if self.l2cap_intr_channel is not None:
  raise InvalidStateError('invalid state')
```

bumble/hid.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def on_ctrl_pdu(self, pdu: bytes) -> None:
+        logger.debug(f'<<< HID CONTROL PDU: {pdu.hex()}')
+        # Here we will receive all kinds of packets, parse and then call respective callbacks
+        message_type = pdu[0] >> 4
+        param = pdu[0] & 0x0f
+        if (message_type == HID_HANDSHAKE):
+            logger.debug('<<< HID HANDSHAKE')
+            self.handle_handshake(param)
+            self.emit('handshake', )
```
zxzxwu, line 144: self.emit('handshake', )

bumble/hid.py
```diff
@@ -0,0 +1,246 @@
...
+                logger.debug('<<< HID CONTROL OPERATION UNSUPPORTED')
+        else:
+            logger.debug('<<< HID CONTROL DATA')
+            self.emit('data', pdu)
+
+    def on_intr_pdu(self, pdu: bytes) -> None:
+        logger.debug(f'<<< HID INTERRUPT PDU: {pdu.hex()}')
+        self.emit("data", pdu)
+
+    def register_data_cb(self, data_cb):
```
zxzxwu, line 168: This doesn't look necessary as watcher can watch by themselves?

bumble/hid.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def register_handshake_cb(self, handshake_cb):
+        self.on('handshake', handshake_cb)
+
+    def register_virtual_cable_unplug(self, virtual_cable_unplug_cb):
+        self.on('virtual_cable_unplug', virtual_cable_unplug_cb)
+
+    def get_report(self, report_type, report_id, buffer_size):
+        if(report_type == HID_OTHER_REPORT):
+            param = report_type
```
zxzxwu, line 179: Typing

bumble/hid.py
```diff
@@ -0,0 +1,246 @@
...
+        else:
+            param = 0x08 | report_type
+        header = ((HID_GET_REPORT << 4) | param)
+        msg = bytes([header, report_id, (buffer_size & 0xff), ((buffer_size >> 8) & 0xff)])
```
zxzxwu, line 183: Looks like a packet format. Can it be a packet class?

bumble/sdp.py
```diff
@@ -167,7 +167,7 @@ class DataElement:
         UUID: lambda x: DataElement(
             DataElement.UUID, core.UUID.from_bytes(bytes(reversed(x)))
         ),
-        TEXT_STRING: lambda x: DataElement(DataElement.TEXT_STRING, x.decode('utf8')),
+        TEXT_STRING: lambda x: DataElement(DataElement.TEXT_STRING, x),
```
barbibulle, line 170: We don't want to keep strings as byte arrays here. If you're encountering a byte array that can't be decoded as UTF-8, we need to understand why. The spec isn't very clear about SDP text strings in that regard (doesn't specify if it is assumed to be ASCII or UTF-8, but UTF-8 seems like a reasonable choice given that ASCII strings are also valid UTF-8 strings). If for some reason there are strings that are encoded differently (like with an ISO-LATIN-1 character set for example), then we could have a safe conversion function here that could try decoding the string with a few common encodings, then if all fails, replace non-ascii characters with a placeholder).

barbibulle: Do you have an example os an SDP record that has this 0xa1 byte in it?

bumble/hid.py
```diff
@@ -0,0 +1,246 @@
...
+HID_BOOT_PROTOCOL_MODE      = 0x00
+HID_REPORT_PROTOCOL_MODE    = 0x01
+
+# Control Operations
+HID_SUSPEND              = 0x03
+HID_EXIT_SUSPEND         = 0x04
+HID_VIRTUAL_CABLE_UNPLUG = 0x05
+
+# -----------------------------------------------------------------------------
+class HIDHost(EventEmitter):
```
barbibulle, line 75: I suggest naming this class just `Host`, since it lives in the namespaced `hid` module (hence `hid.Host`).

bumble/hid.py
```diff
@@ -0,0 +1,246 @@
...
+# -----------------------------------------------------------------------------
+logger = logging.getLogger(__name__)
+
+
+# -----------------------------------------------------------------------------
+# Constants
+# -----------------------------------------------------------------------------
+# fmt: off
+
+HID_CTRL_PSM            = 0x0011
```
barbibulle, line 39: prefer full names rather than contractions: `HID_CONTROL_PSM` and `HID_INTERRUPT_PSM`

bumble/hid.py
```diff
@@ -0,0 +1,246 @@
...
+        self.device = device
+        self.connection = connection
+        self.l2cap_ctrl_channel= None
+        self.l2cap_intr_channel = None
+
+        # Register ourselves with the L2CAP channel manager
+        device.register_l2cap_server(HID_CTRL_PSM, self.on_connection)
+        device.register_l2cap_server(HID_INTR_PSM, self.on_connection)
+
+    async def control_channel_connect(self) -> None:
```
barbibulle, line 89: suggest: `connect_control_channel` (and `disconnect_control_channel` `connect_interrupt_channel` `disconnect_interrupt_channel` below)

examples/run_hid_host.py
```diff
@@ -0,0 +1,461 @@
...
+                print("12. Virtual Cable Unplug")
+                print("13. Disconnect device")
+                print("14. Delete Bonding")
+                print("15. Re-connect to device")
+                print("\nEnter your choice : \n")
+
+                choice = await reader.readline()
+                choice = choice.decode('utf-8').strip()
+
+                if (choice == '1'):
```
barbibulle, line 324: nit: all the parentheses in this test and the elif tests below are not necessary.

bumble/hid.py
```diff
@@ -0,0 +1,246 @@
...
+        self.on('data', data_cb)
+
+    def register_handshake_cb(self, handshake_cb):
+        self.on('handshake', handshake_cb)
+
+    def register_virtual_cable_unplug(self, virtual_cable_unplug_cb):
+        self.on('virtual_cable_unplug', virtual_cable_unplug_cb)
+
+    def get_report(self, report_type, report_id, buffer_size):
+        if(report_type == HID_OTHER_REPORT):
```
barbibulle, line 178: style: parentheses not necessary

bumble/hid.py
```diff
@@ -0,0 +1,246 @@
...
+HID_SET_PROTOCOL        = 0x07
+HID_DATA                = 0x0A
+
+# Report types
+HID_OTHER_REPORT        = 0x00
+HID_INPUT_REPORT        = 0x01
+HID_OUTPUT_REPORT       = 0x02
+HID_FEATURE_REPORT      = 0x03
+
+# Handshake parameters
```
barbibulle, line 57: These handshake constants should be defined in the `Host` class, they are not needed outside. Same for the control operations below.

bumble/hid.py
SneKarnataki: Added to hid.py

bumble/hid.py
SneKarnataki: Done

bumble/hid.py
SneKarnataki: Done

bumble/hid.py
SneKarnataki: Added

bumble/hid.py
SneKarnataki: Updated

bumble/hid.py
SneKarnataki: Taken Care

bumble/hid.py
SneKarnataki: done

bumble/hid.py
SneKarnataki: done HIDPacket class

bumble/sdp.py
SneKarnataki: Changed the encoding to "latin1", we are able to decode/encode HID Descriptors.
Previously in utf8, if failed for below example
HID Report DescriptorList
HID_KEYBOARD_REPORT_MAP = bytes(
    [
        0x05,
        0x01,  # Usage Page (Generic Desktop Controls)
        0x09,
        0x06,  # Usage (Keyboard)
        0xA1,
        0x01,  # Collection (Application)
        0x85,
        0x01,  # . Report ID (1)
        0x05,
        0x07,  # . Usage Page (Keyboard/Keypad)
        0x19,
        0xE0,  # . Usage Minimum (0xE0)
        0x29,
        0xE7,  # . Usage Maximum (0xE7)
        0x15,
        0x00,  # . Logical Minimum (0)
        0x25,
        0x01,  # . Logical Maximum (1)
        0x75,
        0x01,  # . Report Size (1)
        0x95,
        0x08,  # . Report Count (8)
        0x81,
        0x02,  # . Input (Data,Var,Abs,No Wrap,Linear,Preferred State,No Null Position)
        0x95,
        0x01,  # . Report Count (1)
        0x75,
        0x08,  # . Report Size (8)
        0x81,
        0x01,  # . Input (Const,Array,Abs,No Wrap,Linear,Preferred State,No Null Position)
        0x95,
        0x06,  # . Report Count (6)
        0x75,
        0x08,  # . Report Size (8)
        0x15,
        0x00,  # . Logical Minimum (0x00)
        0x25,
        0x94,  # . Logical Maximum (0x94)
        0x05,
        0x07,  # . Usage Page (Keyboard/Keypad)
        0x19,
        0x00,  # . Usage Minimum (0x00)
        0x29,
        0x94,  # . Usage Maximum (0x94)
        0x81,
        0x00,  # . Input (Data,Array,Abs,No Wrap,Linear,Preferred State,No Null Position)
        0x95,
        0x05,  # . Report Count (5)
        0x75,
        0x01,  # . Report Size (1)
        0x05,
        0x08,  # . Usage Page (LEDs)
        0x19,
        0x01,  # . Usage Minimum (Num Lock)
        0x29,
        0x05,  # . Usage Maximum (Kana)
        0x91,
        0x02,  # . Output (Data,Var,Abs,No Wrap,Linear,Preferred State,No Null Position,Non-volatile)
        0x95,
        0x01,  # . Report Count (1)
        0x75,
        0x03,  # . Report Size (3)
        0x91,
        0x01,  # . Output (Const,Array,Abs,No Wrap,Linear,Preferred State,No Null Position,Non-volatile)
        0xC0,  # End Collection
    ]

bumble/hid.py
SneKarnataki: We have Host.py with Host Class and device.py with Device class. These classes are used in HID Profile. Hence we are maintaining the name with HID initials.

bumble/hid.py
SneKarnataki: done

bumble/hid.py
SneKarnataki: done

examples/run_hid_host.py
SneKarnataki: done

bumble/hid.py
SneKarnataki: done

bumble/hid.py
SneKarnataki: We will be using constants in HID Device role, hence maintained outside Host Class.

SneKarnataki:
Hi zxzxwu, barbibulle,

Thank you for your comment. We have checked-in the updated files in bumble_hid branch. 
Request you to provide any feedback.

Regards,
Mindtree Team

bumble/hid.py
barbibulle: The BT spec calls those Message. Please use `class Message` rather than `class HIDPacket`

bumble/hid.py
```diff
@@ -0,0 +1,295 @@
...
+                 protocol_mode: Optional[int] = None,
+                 data: Optional[bytes] = None) -> None:
+
+        self.report_type = report_type
+        self.report_id = report_id
+        self.buffer_size = buffer_size
+        self.protocol_mode = protocol_mode
+        self.data = data
+
+    def to_bytes_gr(self) -> bytes:
```
barbibulle, line 91: Instead of having several `to_bytes_xxx` methods, each Message type should be a subclass, each with its own `__bytes__` method.

bumble/hid.py
```diff
@@ -0,0 +1,295 @@
...
+HID_BOOT_PROTOCOL_MODE      = 0x00
+HID_REPORT_PROTOCOL_MODE    = 0x01
+
+# Control Operations
+HID_SUSPEND              = 0x03
+HID_EXIT_SUSPEND         = 0x04
+HID_VIRTUAL_CABLE_UNPLUG = 0x05
+
+
+class HIDPacket():
```
barbibulle, line 77: As per the spec, this class should be named `Message`

bumble/hid.py
barbibulle: Those should then probably be defined as class properties of the `Message` class (see below about renaming `HIDPacket` to `Message`

bumble/hid.py
```diff
@@ -0,0 +1,295 @@
...
+    def __init__(self,
```
barbibulle, line 78: With subclasses, some of those fields should move to subclass constructors, as they only apply to certain message types.

bumble/sdp.py
```diff
@@ -167,7 +167,7 @@ class DataElement:
         UUID: lambda x: DataElement(
             DataElement.UUID, core.UUID.from_bytes(bytes(reversed(x)))
         ),
-        TEXT_STRING: lambda x: DataElement(DataElement.TEXT_STRING, x.decode('utf8')),
+        TEXT_STRING: lambda x: DataElement(DataElement.TEXT_STRING, x.decode('latin1')),
```
barbibulle, line 170: After checking the spec, since no specific encoding is required for Data Element objects of type String, I think the best option here is to leave it as a byte array and then let the user of the object perform encoding or decoding, as appropriate depending on their context and profile (the encoding may be specified elsewhere for certain use cases)

bumble/sdp.py
```diff
@@ -376,7 +376,9 @@ def __bytes__(self):
                 raise ValueError('invalid value_size')
         elif self.type == DataElement.UUID:
             data = bytes(reversed(bytes(self.value)))
-        elif self.type in (DataElement.TEXT_STRING, DataElement.URL):
+        elif self.type == DataElement.TEXT_STRING:
+            data = self.value.encode('latin1')
```
barbibulle, line 380: See comment above: we should leave strings as unencoded byte arrays

bumble/hid.py
uael, line 77: This class seems to be a good candidate for [`dataclass`](https://docs.python.org/3/library/dataclasses.html)

bumble/hid.py
```diff
@@ -0,0 +1,295 @@
...
+        device.register_l2cap_server(HID_INTERRUPT_PSM, self.on_connection)
+
+    async def connect_control_channel(self) -> None:
+        # Create a new L2CAP connection - control channel
+        try:
+            self.l2cap_ctrl_channel = await self.device.l2cap_channel_manager.connect(
+                self.connection, HID_CONTROL_PSM
+            )
+        except ProtocolError as error:
+            logger.error(f'L2CAP connection failed: {error}')
```
uael, line 151: nit: use `logging.exception` instead and remove the `as error`

bumble/hid.py
```diff
@@ -0,0 +1,295 @@
...
+        self.l2cap_ctrl_channel.sink = self.on_ctrl_pdu
+
+    async def connect_interrupt_channel(self) -> None:
+        # Create a new L2CAP connection - interrupt channel
+        try:
+            self.l2cap_intr_channel = await self.device.l2cap_channel_manager.connect(
+                self.connection, HID_INTERRUPT_PSM
+            )
+        except ProtocolError as error:
+            logger.error(f'L2CAP connection failed: {error}')
```
uael, line 165: nit: use `logging.exception` instead and remove the `as error`

bumble/hid.py
```diff
@@ -0,0 +1,295 @@
...
+    async def disconnect_control_channel(self) -> None:
+        if self.l2cap_ctrl_channel is None:
+            raise InvalidStateError('invalid state')
+        await self.l2cap_ctrl_channel.disconnect()  # type: ignore
+
+    def on_connection(self, l2cap_channel: l2cap.Channel) -> None:
+        logger.debug(f'+++ New L2CAP connection: {l2cap_channel}')
+        l2cap_channel.on('open', lambda: self.on_l2cap_channel_open(l2cap_channel))
+
+    def on_l2cap_channel_open(self, l2cap_channel: l2cap.Channel) -> None:
```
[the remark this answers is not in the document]
barbibulle: I don't see an opportunity for inlining here: when `on_connection` is invoked, the channel isn't open yet, it will be open after some transitions of the state machine, at which point `open` will be emitted. (unless I misunderstood the question)

SneKarnataki:
Hi all,

We have updated review comments.
Thank you.

Regards,
LTIMindtree Team

bumble/hid.py
```diff
@@ -0,0 +1,301 @@
...
+            self.emit('data', pdu)
+
+
+    def on_intr_pdu(self, pdu: bytes) -> None:
+        logger.debug(f'<<< HID INTERRUPT PDU: {pdu.hex()}')
+        self.emit("data", pdu)
+
+    def get_report(self, report_type: int, report_id: int, buffer_size: int) -> None:
+        msg = GetReportMessage(report_type = report_type , report_id = report_id , buffer_size = buffer_size)
+        hid_message = msg.__bytes__()
```
barbibulle, line 250: to convert an object to bytes via its `__bytes__` method, the idiomatic syntax is: `bytes(msg)` rather than invoking the dunder method directly.

bumble/hid.py
```diff
@@ -0,0 +1,301 @@
...
+            param = self.report_type
+        else:
+            param = 0x08 | self.report_type
+        header = ((Message.MessageType.GET_REPORT << 4) | param)
+        packet_bytes = bytearray()
+        packet_bytes.append(header)
+        packet_bytes.append(self.report_id)
+        packet_bytes.extend([(self.buffer_size & 0xff), ((self.buffer_size >> 8) & 0xff)])
+        return bytes(packet_bytes)
+@dataclass
```
barbibulle, line 105: there should be 2 blank lines between top-level function and class definitions (see [pep8)](https://peps.python.org/pep-0008/#blank-lines)

bumble/hid.py
```diff
@@ -0,0 +1,301 @@
...
+class SetReportMessage(Message):
+    report_type: int
+    data : bytes
+
+    def __bytes__(self) -> bytes:
+        header = ((Message.MessageType.SET_REPORT << 4) | self.report_type)
+        packet_bytes = bytearray()
```
barbibulle, line 112: creating a `bytearray` here doesn't seem necessary. you could just:
```
return header + self.data
```

Event better, the code to compute the top 4 bits of the header is pretty much the same for all the message subclasses, so you could factor it out in a Message method.
For example, you could have a `message_type: MessageType` member in the `Message` superclass, and a
```
class Message:
    message_type: MessageType
    ...

    @classmethod
    def header(cls, lower_bits: int):
        return bytes([(cls.message_type << 4) | lower_bits])
```

Then a subclass like `SetReportMessage` would have:
```
@dataclass
class SetReportMessage(Message):
    message_type = Message.MessageType.SET_REPORT
    report_type: int
    data : bytes

    def __bytes__(self) -> bytes:
        return self.header(self.report_type) + self.data
```

(the same comment applies to most other Message classes below)

bumble/hid.py
```diff
@@ -0,0 +1,301 @@
...
+
+    data : bytes
+
+    def __bytes__(self) -> bytes:
+        header = ((Message.MessageType.DATA << 4) | Message.ReportType.OUTPUT_REPORT)
+        packet_bytes = bytearray()
+        packet_bytes.append(header)
+        packet_bytes.extend(self.data)
+        return bytes(packet_bytes)
+# -----------------------------------------------------------------------------
```
barbibulle, line 145: missing two blank lines above the separator

examples/run_hid_host.py
```diff
@@ -0,0 +1,465 @@
...
+                print(color('  HID Profileversion number : ', 'cyan'), hex(attribute.value.value[0].value[1].value))
+
+            elif  attribute.id == SDP_ADDITIONAL_PROTOCOL_DESCRIPTOR_LIST_ATTRIBUTE_ID :
+                print(color('  BT_L2CAP_PROTOCOL_ID : ', 'cyan'), attribute.value.value[0].value[0].value[0].value)
+                print(color('  PSM for Bluetooth HID Interrupt channel : ', 'cyan'), hex(attribute.value.value[0].value[0].value[1].value))
+                print(color('  BT_HIDP_PROTOCOL_ID : ', 'cyan'), attribute.value.value[0].value[1].value[0].value)
+
+            elif  attribute.id == SDP_HID_SERVICE_NAME_ATTRIBUTE_ID :
+                print(color('  Service Name: ', 'cyan'), attribute.value.value)
+                HID_Service_Name = attribute.value.value
```
barbibulle, line 136: all these `HID_Abc_Def` variables are set here but don't seem to be used anywhere. What is their role?

bumble/hid.py
SneKarnataki: Done

bumble/hid.py
SneKarnataki: Thank you, have fixed the same

examples/run_hid_host.py
SneKarnataki: SDP HID related attributes were maintained for future use. But since they are not currently used by application, removed the variables and we are just printing them.

bumble/hid.py
SneKarnataki: Added

SneKarnataki:
Hi barbibulle,
Thank you for inputs.
We have updated the file.

Regards,
LTIMindtree team

barbibulle:
> Hi barbibulle, Thank you for inputs. We have updated the file.
> 
> Regards, LTIMindtree team

This looks good, thanks. Please see a couple new (minor) comments I added as I was reading through the code again.

Also, please run the formatter before this can be merged (see the output of the GitHub actions to see the pre-commit checks that failed). You can run the formatter by running `invoke project.format` from the top level (you'll need to `pip install -e ".[development]" to ensure you have the proper linter and formatter installed so you can run the pre-commit checks including the linter and formatter.

Thanks.

SneKarnataki:
> > Hi barbibulle, Thank you for inputs. We have updated the file.
> > Regards, LTIMindtree team
> 
> This looks good, thanks. Please see a couple new (minor) comments I added as I was reading through the code again.
> 
> Also, please run the formatter before this can be merged (see the output of the GitHub actions to see the pre-commit checks that failed). You can run the formatter by running `invoke project.format` from the top level (you'll need to `pip install -e ".[development]" to ensure you have the proper linter and formatter installed so you can run the pre-commit checks including the linter and formatter.
> 
> Thanks.


Thank you barbibulle. I was able to perform "invoke project.format". But I am unable to view "_Please see a couple new (minor) comments I added as I was reading through the code again_." comments in PR.

barbibulle:
> > > Hi barbibulle, Thank you for inputs. We have updated the file.
> > > Regards, LTIMindtree team
> > 
> > 
> > This looks good, thanks. Please see a couple new (minor) comments I added as I was reading through the code again.
> > Also, please run the formatter before this can be merged (see the output of the GitHub actions to see the pre-commit checks that failed). You can run the formatter by running `invoke project.format` from the top level (you'll need to `pip install -e ".[development]" to ensure you have the proper linter and formatter installed so you can run the pre-commit checks including the linter and formatter.
> > Thanks.
> 
> Thank you barbibulle. I was able to perform "invoke project.format". But I am unable to view "_Please see a couple new (minor) comments I added as I was reading through the code again_." comments in PR.

Sorry about that, the two comments were pending and I forgot to send them. They should be visible now.

bumble/hid.py
```diff
@@ -0,0 +1,301 @@
...
+# Logging
+# -----------------------------------------------------------------------------
+logger = logging.getLogger(__name__)
+
+
+# -----------------------------------------------------------------------------
+# Constants
+# -----------------------------------------------------------------------------
+# fmt: off
```
uael, line 44: suggest: re-enable the formater

bumble/hid.py
```diff
@@ -0,0 +1,302 @@
...
+    data : bytes
+    message_type = Message.MessageType.DATA
+
+    def __bytes__(self) -> bytes:
+        return self.header(Message.ReportType.OUTPUT_REPORT) + self.data
+
+
+# -----------------------------------------------------------------------------
+class Host(EventEmitter):
+    l2cap_channel: Optional[l2cap.Channel]
```
uael, line 150:
```suggestion
    l2cap_ctrl_channel: Optional[l2cap.Channel]
    l2cap_intr_channel: Optional[l2cap.Channel]
```

uael (approved):
Good job :tada:

barbibulle:
@SneKarnataki the "Code format and lint check" GitHub action still fails on this PR (a couple places where the typing needs to be fixed). Can you check the output of the GitHub action and fix the issue before we can merge?
(I recommend running `invoke project.pre-commit` locally before pushing commits, this way you can catch most of these error in advance).

uael:
One more format issue :pray: 
```patch
diff --git a/examples/hid_report_parser.py b/examples/hid_report_parser.py
index e5f407f..61561b3 100644
--- a/examples/hid_report_parser.py
+++ b/examples/hid_report_parser.py
@@ -139,7 +139,6 @@ class Mouse:
 
 # ------------------------------------------------------------------------------
 class ReportParser:
-
     @staticmethod
     def parse_input_report(input_report: bytes) -> None:
```

SneKarnataki:
Thank you all for guiding our first GitHub checkin and helping to make it optimized.

Regards,
LTIMindtree

PR #298, opened 2023-09-22
use Address.to_string instead of manual suffix replacement

barbibulle:


uael (approved):

apps/console.py
```diff
@@ -1172,7 +1172,7 @@ def to_display_string(self):
             name = ''
 
         # Remove any '/P' qualifier suffix from the address string
-        address_str = str(self.address).replace('/P', '')
+        address_str = self.address.to_string(False)
```
AlanRosenthal, line 1175: it's not obvious by reading this LOC what False is referring to. is that param required or could we change it to optional?

barbibulle: This param is `with_type_qualifier`, it is already optional, in the sense that it has a default value (`True`). To make it clearer, I'll update this call to say `to_string(with_type_qualifier=False)` explicitly.

PR #303, opened 2023-09-27
Ability to send HCI commands from Rust

whitevegagabriel:
* Autogenerate packet code in Rust from PDL (packet file copied from rootcanal)
* Implement parsing of packets that have a type header
* Expose Python APIs for sending HCI commands
* Expose Python APIs for instantiating a local controller


bumble/controller.py
```diff
@@ -194,6 +194,21 @@ def __init__(
 
         self.terminated = asyncio.get_running_loop().create_future()
 
+    @classmethod
```
barbibulle, line 197: It's maybe too verbose to have a `create` method with all the same args as `__init__`.
A generic solution would be cleaner.
Something like:
```
async def async_call(function, *args, **kwargs):
    return function(*args, **kwargs)
```
(which could be put in `utils.py`)
Can be invoked either with an argument-less lambda, or passing the func and its args. Examples:
```
c = await async_call(Controller, name, ...)
```
or
```
c = await async_call(lambda: Controller(name, ...))
```

Or if you actually need to create a function object, this decorator would be convenient:
```
def asyncify(function):
    return partial(async_call, function)
```
which you'd use like:
```
constructor = asyncify(Controller)
...
controller = await constructor(name, ...)
```

whitevegagabriel: Excellent idea.

I've moved forward with your first suggested solution, and it works great.

I suppose creating a function object would make it even cleaner on the Rust side, but I'm only using `async_call` in one place, so I've only added minimum necessary.

rust/src/wrapper/common.rs
```diff
@@ -0,0 +1,34 @@
...
+impl ToPyObject for TransportSink {
+    fn to_object(&self, _py: Python<'_>) -> PyObject {
+        self.0.clone()
+    }
+}
+
+/// Represents the source for some transport mechanism
+pub struct TransportSource(pub(crate) PyObject);
+
+impl ToPyObject for TransportSource {
```
nabilwadih, line 30: could use the `From/Into` traits eg: `impl Into<PyObject> for Transport`

rust/src/wrapper/hci.rs
```diff
@@ -134,12 +112,45 @@ impl Address {
...
+
+/// Implements minimum necessary interface to be treated as bumble's [HCI_Command].
+/// While pyo3's macros do not support generics, this could probably be refactored to allow multiple
+/// implementations of the HCI_Command methods in the future, if needed.
+#[pyclass]
+pub(crate) struct HciCommandWrapper(pub(crate) Command);
+
+#[pymethods]
+impl HciCommandWrapper {
+    fn __bytes__(&self, py: Python) -> PyResult<PyObject> {
```
nabilwadih, line 129: whats the meaning of the double underscores ?

nabilwadih (approved):

bumble/controller.py
whitevegagabriel: Thinking about it further, I decided that its worth having a cleaner solution, so I also added the "asyncify" function/decorator.

rust/src/internal/hci/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+                    value: value1,
+                    type_: type_1,
+                },
+                Error::InvalidEnumValueError {
+                    obj: obj2,
+                    field: field2,
+                    value: value2,
+                    type_: type_2,
+                },
+            ) => obj1 == obj2 && field1 == field2 && value1 == value2 && type_1 == type_2,
```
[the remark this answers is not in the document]
whitevegagabriel: In an ideal world, I would do that. But the packets::Error enum declaration is autogenerated by PDL, so I think my only good option is to file a bug (https://github.com/google/pdl/issues/71) and implement here for now.

But if there are other ways of deriving PartialEq in a separate file than the declaration, I'm all ears!

rust/src/internal/hci/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+) -> Result<T, PacketTypeParseError>
+where
+    F: Fn(&[u8]) -> Result<T, E>,
+    PacketTypeParseError: From<E>,
+{
+    let (first_byte, packet_bytes) = bytes
+        .split_first()
+        .ok_or(PacketTypeParseError::EmptySlice)?;
+    let actual_packet_type = PacketType::try_from(*first_byte)?;
+    if PacketType::try_from(*first_byte)? == expected_packet_type {
```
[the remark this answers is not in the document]
whitevegagabriel: Oops, good catch!

rust/src/wrapper/common.rs
whitevegagabriel: Normally I would do that, but `ToPyObject` is the standard trait requirement for converting to PyObject throughout `pyo3` apis, so only implementing Into<PyObject> would end up making it more verbose elsewhere.

rust/src/wrapper/hci.rs
whitevegagabriel: This is the Python function that gets called internally when a caller (in Python) invokes `bytes(some_obj)` (https://docs.python.org/3/reference/datamodel.html?highlight=__bytes__#object.__bytes__)

As to why Python uses double underscore before and after such methods... Probably to discourage calling them directly? They're typically magic methods that are special to Python.

rust/src/internal/hci/mod.rs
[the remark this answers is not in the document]
whitevegagabriel: PDL recently released 0.2.0 where PartialEq is implemented for Error, so I have updated to that version.

barbibulle (approved):

PR #305, opened 2023-09-28
Pin type

benquike:
Add commands for reading/writing pin type

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -74,6 +74,7 @@ from .hci import (
     HCI_REMOTE_USER_TERMINATED_CONNECTION_ERROR,
     HCI_SUCCESS,
     HCI_WRITE_LE_HOST_SUPPORT_COMMAND,
+    HCI_WRITE_PIN_TYPE_COMMAND,
     Address,
     HCI_Accept_Connection_Request_Command,
     HCI_Authentication_Requested_Command,
@@ -128,6 +129,7 @@ from .hci import (
     HCI_Write_Inquiry_Mode_Command,
     HCI_Write_LE_Host_Support_Command,
     HCI_Write_Local_Name_Command,
+    HCI_Write_PIN_Type_Command,
     HCI_Write_Scan_Enable_Command,
     HCI_Write_Secure_Connections_Host_Support_Command,
     HCI_Write_Simple_Pairing_Mode_Command,
@@ -1307,6 +1309,11 @@ class Device(CompositeEventEmitter):
                 )
 
         if self.classic_enabled:
+            if self.host.supports_command(HCI_WRITE_PIN_TYPE_COMMAND):
+                # setup pin type for legacy pairing
+                await self.send_command(
+                    HCI_Write_PIN_Type_Command(pin_type=self.config.classic_pin_type)
+                )
             await self.send_command(
                 HCI_Write_Local_Name_Command(local_name=self.name.encode('utf8'))  # type: ignore[call-arg]
             )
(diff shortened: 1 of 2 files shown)


bumble/hci.py
```diff
@@ -2694,17 +2709,45 @@ class HCI_Reset_Command(HCI_Command):
...
+class HCI_Read_PIN_Type_Command(HCI_Command):
+    '''
+    See Bluetooth spec @ 7.3.5 Read PIN Type Command
+    '''
+
+
+# -----------------------------------------------------------------------------
+@HCI_Command.command(
+    fields=[
+        ('pin_type', 1),
```
barbibulle, line 2739: This should probably use `PIN_TYPE_SPEC` here too.

bumble/hci.py
```diff
@@ -2694,17 +2709,45 @@ class HCI_Reset_Command(HCI_Command):
 
 # -----------------------------------------------------------------------------
 @HCI_Command.command(
-    [
+    fields=[
         ('filter_type', 1),
-        ('filter_condition', '*'),
-    ]
+        # TODO: separate this field into condition_type and condition fields
```
barbibulle, line 2714: Unfortunately, there's no "easy" way to do this just from the fields description, so this would have to be done in code rather than data, as class methods (as is done for a small number of other commands that also need some level of custom parsing/serializing)

barbibulle (approved):

bumble/hci.py
benquike: Updated the comments

bumble/hci.py
benquike: Updated

uael:
I'm not sure to understand the need for this, the default is already VARIABLE and `rootcanal` does not support fixed pins.
Do you have a use case ?

benquike:
> I'm not sure to understand the need for this, the default is already VARIABLE and `rootcanal` does not support fixed pins. Do you have a use case ?

The need: 
support for testing different settings of BR/EDR Legacy pairing, which  is part of the tests.

On support on root-canal
If rootcanal does not support it, I think we we need to add support for it.

Even if rootcanal does not support it, we still need it to test with physical controllers.

DeltaEvo:
> On support on root-canal
> If rootcanal does not support it, I think we we need to add support for it.

AFAIU, this setting has no effect from the peer's Host point of view, it's why I didn't implemented it on root-canal.

Do you have some examples on how this setting affect the behavior of the DUT and why we want to alter it during tests ?

> Even if rootcanal does not support it, we still need it to test with physical controllers.

Do you want to test the controller as well or the host on top of it ?

bumble/device.py
```diff
@@ -1301,6 +1308,10 @@ async def power_on(self) -> None:
                 )
 
         if self.classic_enabled:
+            # setup pin type for legacy pairing
+            await self.send_command(
```
DeltaEvo, line 1312: This command is optional, I think we should check if it's supported before sending it.

benquike:
BLUETOOTH CORE SPECIFICATION Version 5.4 | Vol 2, Part C 

4.2.2.1 Responder accepts pairing and has a variable PIN

4.2.2.2 Responder accepts pairing and has a fixed PIN

uael:
> Please refer to the following sections for reference.
> 
> BLUETOOTH CORE SPECIFICATION Version 5.4 | Vol 2, Part C 4.2.2.1 Responder accepts pairing and has a variable PIN 4.2.2.2 Responder accepts pairing and has a fixed PIN

But none of the procedures detailed above change the behavior on Android (Host) side, only bumble side

uael:
ie. the received HCI events on the Android side will remain the same.
Or I'm wrong here ?

barbibulle:
You'll need to fix the linter errors before we can merge this. See the output of the GH actions. You can run `invoke project.pre-commit` to perform a local check before pushing a commit.

uael:
After an offline discussion it has been decided that this is not needed and so not worth the addition. If it changes, we can still re-open tho

PR #306, opened 2023-09-29
ci: add python avatar tests

uael:


.github/workflows/python-avatar.yml
```diff
@@ -0,0 +1,46 @@
...
+        ]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set Up Python 3.11
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.11
+      - name: Install
+        run: |
+          pip install --upgrade pip
```
barbibulle, line 35: I think the recommended invocation for pip in this type of action is `python -m pip` rather than just `pip` (I forget the exact reason, but I remember making that change for the other GitHub actions in this project)

barbibulle (approved):

.github/workflows/python-avatar.yml
zxzxwu: Because pip works in global, so when the project runs in a venv, it will fail. Though workflows may run in a container, `python -m venv` still avoid potential failures.

uael: Done!
I wasn't aware of that, I'm gonna have a look :+1:

.github/workflows/python-avatar.yml
```diff
@@ -0,0 +1,45 @@
...
+          python -m pip install --upgrade pip
+          python -m pip install rootcanal
+          python -m pip install pandora-avatar
+          python -m pip install .
+      - name: Rootcanal
+        run: nohup python -m rootcanal > rootcanal.log &
+      - name: Test
+        run: |
+          avatar --list | grep -Ev '^=' > test-names.txt
+          timeout 5m avatar --test-beds bumble.bumbles --tests $(split test-names.txt -n l/${{ matrix.shard }})
```
zxzxwu, line 43: As avatar tests take quite a long time even for virtual-bumble-bumble cases, I think it's better to pick only critical cases to test availability coverage.

zxzxwu: Oh it seems you parallelized the test, this is good even for local testing.

zxzxwu (approved):

.github/workflows/python-avatar.yml
zxzxwu: Another problem is that: bring a downstream project to CI probably means it's actually a submodule of the upstream?

uael: > Another problem is that: bring a downstream project to CI probably means it's actually a submodule of the upstream?

I'm not sure to understand :thinking:

uael: > Oh it seems you parallelized the test, this is good even for local testing.

Yes! The whole test suite run under 3 minutes most of the time

zxzxwu: As a public project, the best practice is to depend on a specific version which itself works I think? Forcing the upstream to run tests for downstream is a kind of wear(it's not our internal monorepo).

zxzxwu: And, even you add CI here, Avatar is still depending on a specific release of Bumble(until now) so it always passes.

uael: I was thinking of pining the versions of avatar and rootcanal (if this is what you mean).
But I was afraid that the versions were left un-updated, so I then considered that they were now stable enough to trust their latest releases (ie. to not break bumble master).
What do you think ?
@barbibulle I would like you pov too

uael: > And, even you add CI here, Avatar is still depending on a specific release of Bumble(until now) so it always passes.

The current sources replace the avatar one using `pip install .` after the avatar installation, although I tested only with `bt-test-interfaces`

barbibulle: It does seem somewhat odd that the dependency is setup in this direction. Since developers who contribute to Bumble don't all have the ability to change Avatar, running the avatar tests from the Bumble project actions could lead to a condition where a change in Avatar breaks the test, and they couldn't fix that. Why not run those tests as an action from the Avatar project side?

uael: I will fix the version then, cause we need to check that bumble pass those tests and that the bumble Pandora server is working for every update, so it makes sense to add avatar to the bumble CI

barbibulle: Pinning the version would definitely help avoid surprises. But we still need to figure out how to deal with the case where a Bumble change would require an Avatar update. To avoid a deadlock, the PR should be able to still be merged, with a note that it is “temporarily incompatible with Avatar”. When a compatible PR is merged on Avatar, we should then mark it back as compatible going forward.

uael: Actually this is very unlikely to happen, since the bumble Pandora server has been moved to bumble, avatar now only contains the test framework and a test suite, only breaking changes to a very small subset of bumble APIs can break the tests, and in that case it's mostly something that should be addressed right away.
Also the Avatar test suite covers very well defined Bluetooth scenarios that should always pass on bumble, so I see it as a very good addition to the bumble CI 🙂

barbibulle: Ok, sounds good. I recommend we create an “avatar” section in setup.cfg (with avatar, RootCanal, etc) so that we can maintain dependency versions there rather than hard coded in the GH action. The GH action should use, then, a pip qualifier to install the avatar deps before running the test.

uael: Ok sgtm! Let me update the PR later today 👌

PR #308, opened 2023-10-03
Minimal LE Audio implementation

zxzxwu:
With the example, it can "cheat" Android source that Bumble is an available sink. However, volume control somehow doesn't work anymore.


zxzxwu:
The complete procedure to create Unicast streaming is very long. https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/16212-BAP-html5/out/en/index-en.html#UUID-e610da3a-d737-b38e-7302-6c9e934a99f6

zxzxwu:
I am hesitating because LEA services are not "profiles" - but actually they are separated as several profiles like BAP, CAP, CSIP, so maybe we should name modules after profiles instead of services?

bumble/gatt.py
```diff
@@ -99,12 +99,21 @@
 GATT_VOLUME_OFFSET_CONTROL_SERVICE          = UUID.from_16_bits(0x1845, 'Volume Offset Control')
 GATT_COORDINATED_SET_IDENTIFICATION_SERVICE = UUID.from_16_bits(0x1846, 'Coordinated Set Identification Service')
 GATT_DEVICE_TIME_SERVICE                    = UUID.from_16_bits(0x1847, 'Device Time')
-GATT_MEDIA_CONTROL_SERVICE                  = UUID.from_16_bits(0x1848, 'Media Control Service')
-GATT_GENERIC_MEDIA_CONTROL_SERVICE          = UUID.from_16_bits(0x1849, 'Generic Media Control Service')
+GATT_MEDIA_CONTROL_SERVICE                  = UUID.from_16_bits(0x1848, '[LE Audio] Media Control Service')
```
barbibulle, line 102: I think we should keep all GATT UUID names without a `[Profile]` prefix, so that they are just as listed in the BT spec (also, it is possible, in theory, that the same characteristic could be used in multiple profiles, so tying them to a specific one isn't safe)

bumble/profiles/pacs.py
```diff
@@ -0,0 +1,464 @@
...
+    VOICE_ASSISTANTS = 0x0020
+    LIVE             = 0x0040
+    SOUND_EFFECTS    = 0x0080
+    NOTIFICATIONS    = 0x0100
+    RINGTONE         = 0x0200
+    ALERTS           = 0x0400
+    EMERGENCY_ALARM  = 0x0800
+
+
+class SampleFrequency(enum.IntEnum):
```
barbibulle, line 102: This should be `SamplingFrequency`

bumble/profiles/pacs.py
```diff
@@ -0,0 +1,464 @@
...
+    '''Bluetooth Assigned Numbers, Section 6.12.5.1 - Sample Frequency'''
```
barbibulle, line 103: `Sample` -> `Sampling`

bumble/profiles/pacs.py
```diff
@@ -0,0 +1,464 @@
...
+# -----------------------------------------------------------------------------
+
+
+@dataclasses.dataclass
+class CodecSpecificCapabilities:
+    '''Bluetooth Assigned Numbers, Section 6.12.5 - Codec Specific Capabilities LTV Structures.'''
+
+    class Type(enum.IntEnum):
+        # fmt: off
+        SAMPLE_FREQUENCY     = 0x01
```
barbibulle, line 253: `SAMPLING_FREQUENCY`

bumble/profiles/pacs.py
```diff
@@ -0,0 +1,464 @@
...
+        FRAME_DURATION       = 0x02
+        AUDIO_CHANNEL_COUNT  = 0x03
+        OCTETS_PER_SAMPLE    = 0x04
+        CODEC_FRAMES_PER_SDU = 0x05
+
+    supported_sample_frequencies: SupportedSampleFrequency
```
barbibulle, line 259: `supported_sampling_frequencies`

bumble/profiles/pacs.py
```diff
@@ -0,0 +1,464 @@
...
+            self.min_octets_per_sample,
+            self.max_octets_per_sample,
+            2,
+            CodecSpecificCapabilities.Type.CODEC_FRAMES_PER_SDU,
+            self.supported_max_codec_frames_per_sdu,
+        )
+
+
+@dataclasses.dataclass
+class PAC_Record:
```
barbibulle, line 320: maybe `PacRecord` to keep a consistent class naming?

bumble/profiles/csis.py
```diff
@@ -0,0 +1,127 @@
...
+# -----------------------------------------------------------------------------
+# Server
+# -----------------------------------------------------------------------------
+class Service(gatt.TemplateService):
+    UUID = gatt.GATT_COORDINATED_SET_IDENTIFICATION_SERVICE
+
+    def __init__(
+        self,
+        sirk: bytes,
+        set_size: Optional[int] = None,
```
barbibulle, line 46: I'd recommend keeping the characteristic names exactly as in the spec, even if that's sometimes a bit long, it makes it very clear and unambiguous how the profile implementation matches the spec.
Here, we'd have  `set_identity_resolving_key`, `coordinated_set_size`, `set_member_lock` and `set_member_rank`, as listed in P.5 of the CISC profile spec.

bumble/profiles/ascs.py
```diff
@@ -0,0 +1,155 @@
...
+        pass
+
+
+# -----------------------------------------------------------------------------
+# Client
+# -----------------------------------------------------------------------------
+class ServiceProxy(gatt_client.ProfileServiceProxy):
+    SERVICE_CLASS = Service
+
+    sirk: gatt_client.CharacteristicProxy
```
barbibulle, line 130: I think this has been copy/pasted by mistake from another profile

bumble/profiles/csis.py
zxzxwu: Done.

bumble/profiles/pacs.py
zxzxwu: Done.

bumble/profiles/pacs.py
zxzxwu: Done.

bumble/profiles/ascs.py
zxzxwu: No, why do you think so?

PR #309, opened 2023-10-04
asha: import ASHA Pandora service from AOSP

uael:
- [ ] https://github.com/google/bt-test-interfaces/pull/15
- [ ] pandora Pypi release

### DIFF
--- a/bumble/profiles/asha_service.py
+++ b/bumble/profiles/asha_service.py
@@ -32,6 +32,7 @@ from ..gatt import (
     Characteristic,
     CharacteristicValue,
 )
+from ..l2cap import Channel
 from ..utils import AsyncRunner
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


bumble/profiles/asha_service.py
```diff
@@ -101,49 +104,59 @@ def on_audio_control_point_write(connection: Connection, value):
                     )
                 )
 
+        def on_read_only_properties_read(connection: Connection) -> bytes:
+            value = (
```
barbibulle, line 108: is this value expected to change during the lifetime of the characteristic? If not, why not construct it just once at init time?

bumble/profiles/asha_service.py
```diff
@@ -101,49 +104,59 @@ def on_audio_control_point_write(connection: Connection, value):
...
+                        self.capability,
+                    ]
+                )
+                + bytes(self.hisyncid)
+                + bytes(AshaService.FEATURE_MAP)
+                + bytes(AshaService.RENDER_DELAY)
+                + bytes(AshaService.RESERVED_FOR_FUTURE_USE)
+                + bytes(AshaService.SUPPORTED_CODEC_ID)
+            )
+            self.emit("read_only_properties", connection, value)
```
barbibulle, line 121: The event name is not typical. Event names are normally nouns. The even here is that there is a read request on the characteristic. The event name should probably, then, just be 'read'.
But more important question: is this only for testing? Maybe the Attribute class itself should emit a 'read' event (like it emits a 'write' event), to make this observability more generally available?

bumble/profiles/asha_service.py
```diff
@@ -101,49 +104,59 @@ def on_audio_control_point_write(connection: Connection, value):
...
+            return value
+
+        def on_le_psm_out_read(connection: Connection) -> bytes:
+            self.emit("le_psm_out", connection, self.psm)
```
barbibulle, line 125: same comment about event names as above.

barbibulle:
closing as obsolete

PR #310, opened 2023-10-04
Typing AVDTP

zxzxwu:
Still, many commands wrappers are left untyped. Maybe in another PR we can replace `send_command()` with an overloaded form, which mapping command to their responses.

### DIFF
--- a/bumble/avdtp.py
+++ b/bumble/avdtp.py
@@ -587,6 +587,8 @@ class ServiceCapabilities:
 # -----------------------------------------------------------------------------
 class MediaCodecCapabilities(ServiceCapabilities):
     media_codec_information: Union[bytes, SupportsBytes]
+    media_type: int
+    media_codec_type: int
 
     def init_from_bytes(self) -> None:
         self.media_type = self.service_capabilities_bytes[0]
--- a/bumble/avdtp.py
+++ b/bumble/avdtp.py
@@ -20,8 +20,23 @@ import asyncio
 import struct
 import time
 import logging
+import enum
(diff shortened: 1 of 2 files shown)


bumble/avdtp.py
```diff
@@ -385,32 +408,37 @@ async def stop(self):
 
 
 # -----------------------------------------------------------------------------
-class MessageAssembler:  # pylint: disable=attribute-defined-outside-init
-    def __init__(self, callback):
-        self.callback = callback
```
barbibulle, line 390: As a general design pattern, I'm not a big fan of using event emitters for direct 1:1 relationships like this, where there's normally just one client that wants to receive the data. Having a direct function allows better typing that generic events. Events are better suited for 1:M relationships, where you can dynamically add/remove listeners.

zxzxwu: Done

bumble/avdtp.py
```diff
@@ -551,7 +586,9 @@ def __str__(self):
 
 # -----------------------------------------------------------------------------
 class MediaCodecCapabilities(ServiceCapabilities):
-    def init_from_bytes(self):
+    media_codec_information: Union[bytes, SupportsBytes]
+
```
barbibulle, line 590: missing `media_type` and `media_codec_type`?

barbibulle (approved):

bumble/avdtp.py
zxzxwu: They can be inferred from `__init__`, but I added hints still.

PR #313, opened 2023-10-07
Allow turning on BLE in classic pairing mode

benquike:


apps/pair.py
```diff
@@ -326,7 +327,6 @@ async def pair(
         # Select LE or Classic
         if mode == 'classic':
             device.classic_enabled = True
-            device.le_enabled = False
```
barbibulle, line 329: Maybe what we need instead of always enabling BLE is to add two modes: `classic+le` which would do Classic pairing with LE enabled, and `le+classic` which would do LE pairing with Classic enabled, while just 'classic' would do Classic pairing with LE disabled and 'le' would do LE pairing with Classic disabled.

benquike: A device config file is already used  to define which (classic, le, or classic+le) mode the device runs on, here we do not need a second option (command line) for that.

If we use, we need to define some rules of precedence.

barbibulle: You're right, I forgot about the config file.

barbibulle (approved):

PR #316, opened 2023-10-10
Add support for extended advertising via Rust-only API

whitevegagabriel:
* Extended functionality is gated on an "unstable" feature
* Designed for very simple use and minimal interferance with existing legacy implementation
* Intended to be temporary, until bumble can integrate extended advertising into its core functionality
* Added error case for controller's `on_hci_le_set_scan_parameters_command` (matching the BLE spec) to test error flow from Rust
* Dropped `HciCommandWrapper` in favor of using bumble's `HCI_Command.from_bytes` for converting from PDL into bumble implementation
* Refactored Address and Device constructors to better match what the python constructors expect


rust/src/wrapper/device/mod.rs
```diff
@@ -230,35 +263,159 @@ impl Device {
...
+        let properties = AdvertisingEventProperties {
+            connectable: 0,
+            scannable: 0,
+            directed: 0,
+            high_duty_cycle: 0,
+            legacy: 0,
+            anonymous: 0,
+            tx_power: 0,
+        };
+        let extended_advertising_params_cmd = LeSetExtendedAdvertisingParametersBuilder {
```
whitevegagabriel, line 325: @barbibulle are these good default parameters?

rust/src/wrapper/device/mod.rs
```diff
@@ -230,35 +263,159 @@ impl Device {
...
+                ConversionError::Native(e) => PyErr::new::<PyException, _>(format!("{e:?}")),
+            })?;
+        let random_address_cmd = LeSetAdvertisingSetRandomAddressBuilder {
+            advertising_handle,
+            random_address,
+        };
+        self.send_command(random_address_cmd.into(), true).await?;
+
+        // set adv data
+        let advertising_data_cmd = LeSetExtendedAdvertisingDataBuilder {
```
whitevegagabriel, line 358: @barbibulle are these good default parameters?

rust/examples/broadcast.rs
```diff
@@ -63,17 +63,28 @@ async fn main() -> PyResult<()> {
...
-    device.set_advertising_data(adv_data)?;
     device.power_on().await?;
 
-    println!("Advertising...");
-    device.start_advertising(true).await?;
+    if cli.extended {
+        println!("Starting extended advertisement...");
+        device.start_advertising_extended(adv_data).await?;
+    } else {
+        device.set_advertising_data(adv_data)?;
```
nabilwadih, line 72: why does the non-extended advertising need set advertisement data separately?

rust/src/wrapper/device/mod.rs
```diff
@@ -230,35 +263,159 @@ impl Device {
...
         Python::with_gil(|py| {
-            self.0
+            self.obj
                 .getattr(py, intern!(py, "host"))
                 .map(|obj| obj.into_option(Host::from))
         })
     }
 
     /// Start advertising the data set with [Device.set_advertisement].
     pub async fn start_advertising(&mut self, auto_restart: bool) -> PyResult<()> {
```
nabilwadih, line 273: nit: Docs for that auto-restart parameter, what does it do?

rust/src/wrapper/device/mod.rs
```diff
@@ -230,35 +263,159 @@ impl Device {
...
+        if self.advertising_status == AdvertisingStatus::AdvertisingExtended {
```
nabilwadih, line 274: What about if the device is already advertising in legacy mode? Would that also be an error, or would it get overridden?

rust/src/wrapper/device/mod.rs
```diff
@@ -230,35 +263,159 @@ impl Device {
...
+        match self.advertising_status {
+            AdvertisingStatus::AdvertisingLegacy => return Err(PyErr::new::<PyException, _>("Already advertising in legacy mode. Stop the existing legacy advertisement to start an extended advertisement.")),
+            // Stop the current extended advertisement before advertising with new data.
+            // We could just issue an LeSetExtendedAdvertisingData command, but this approach
+            // allows better future flexibility if `start_advertising_extended` were to change.
+            AdvertisingStatus::AdvertisingExtended => self.stop_advertising_extended().await?,
+            _ => {}
+        }
+
+        // if you change this, don't forget to change the same handle in `stop_advertising_extended`
```
nabilwadih, line 312: can it be moved to a const that is shared between them both?

rust/src/wrapper/device/mod.rs
```diff
@@ -230,35 +263,159 @@ impl Device {
...
+        let advertising_handle = 0x00;
+
+        // set extended params
+        let properties = AdvertisingEventProperties {
```
nabilwadih, line 316: Is there any way for the caller of the library to set these? Should they be exposed in the public API?

rust/src/wrapper/device/mod.rs
```diff
@@ -230,35 +263,159 @@ impl Device {
...
+            self.advertising_status = AdvertisingStatus::NotAdvertising;
+        }
+        Ok(())
+    }
+
+    /// Stop advertising extended.
+    #[cfg(feature = "unstable_extended_adv")]
+    pub async fn stop_advertising_extended(&mut self) -> PyResult<()> {
+        if AdvertisingStatus::AdvertisingExtended != self.advertising_status {
+            return Ok(());
```
nabilwadih, line 402: Should this be return Error?

rust/examples/broadcast.rs
whitevegagabriel: Non-extended advertising sets advertisement data separately to mirror the python behavior.

In adding extended advertising, I don't want to interfere with how `advertisement_data` is used for legacy use cases, so it takes advertisement data as a parameter.

rust/src/wrapper/device/mod.rs
whitevegagabriel: Done.

rust/src/wrapper/device/mod.rs
whitevegagabriel: Currently, bumble ignores the command if its already advertising. Added a comment with rationale for the current implementation.

```
// Bumble allows (and currently ignores) calling `start_advertising` when already
// advertising. Because that behavior may change in the future, we continue to delegate the
// handling to bumble.
```

rust/src/wrapper/device/mod.rs
whitevegagabriel: Doh! That is definitely preferable.

Done.

rust/src/wrapper/device/mod.rs
whitevegagabriel: Currently, my rationale is that since there are plans to eventually integrate extended advertising into the core APIs, it makes more sense to only implement minimal functionality while still being useful.

rust/src/wrapper/device/mod.rs
```diff
@@ -230,35 +263,159 @@ impl Device {
...
+            enable: Enable::Enabled,
+            enabled_sets: vec![EnabledSet {
+                advertising_handle,
+                duration: 0,
+                max_extended_advertising_events: 0,
+            }],
+        };
+        self.send_command(extended_advertising_enable_cmd.into(), true)
+            .await?;
+
```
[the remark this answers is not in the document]
whitevegagabriel: A "Success" response from the controller means that is successfully received and processes the command. If the command was to start advertising, then "Success" would mean that it started advertising.

By setting `check_result` to true in the API call, then Python will throw an exception if the controller response is not "Success", and thus here a successful call means that the command to start advertising was successfully processed by the controller.

I have added some info in the `send_command` API as to what `check_result` does.

Another option would be to set `check_result` to false, and then check here that the event code was "success".

rust/src/wrapper/device/mod.rs
```diff
@@ -230,35 +263,159 @@ impl Device {
...
-            self.0
+            self.obj
                 .call_method0(py, intern!(py, "stop_advertising"))
                 .and_then(|coroutine| into_future(coroutine.as_ref(py)))
         })?
         .await
-        .map(|_| ())
+        .map(|_| ())?;
+
+        if self.advertising_status == AdvertisingStatus::AdvertisingLegacy {
```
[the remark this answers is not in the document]
whitevegagabriel: I agree.

The Python legacy `start_advertising` no-ops on duplicate calls to start advertising, so I followed suit here, but I wasn't sure.

Do you think it would be better to return an error?

rust/src/wrapper/device/mod.rs
whitevegagabriel: Same as above.

The Python legacy stop_advertising no-ops on duplicate calls to stop advertising.

Do you think it would be better to return an error?

rust/examples/broadcast.rs
nabilwadih: ah gotcha, perhaps `set_advertising_data` should be renamed to `set_legacy_advertising_data` then?

rust/src/wrapper/device/mod.rs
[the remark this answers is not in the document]
nabilwadih: Personally, I think an error is more clear, but if this is already the behavior in python than I think it makes sense to follow suite, since these are just bindings to the python

rust/src/wrapper/device/mod.rs
nabilwadih: Same as above, makes sense to keep behavior the same as the python

nabilwadih (approved):

rust/src/wrapper/hci.rs
```diff
@@ -118,27 +147,28 @@ impl ToPyObject for Address {
...
-    fn op_code(&self) -> u16 {
-        self.0.get_op_code().into()
+impl TryInto<packets::Address> for Address {
+    type Error = ConversionError<InvalidAddress>;
+
+    fn try_into(self) -> Result<packets::Address, Self::Error> {
+        let addr_le_bytes = self.as_le_bytes().map_err(ConversionError::Python)?;
+
+        let mut buf = [0_u8; 8];
+        buf[0..6].copy_from_slice(&addr_le_bytes);
```
nandita-v, line 161: It's not very obvious what we're doing here exactly, I think it would help to add a comment explaining!

rust/examples/broadcast.rs
whitevegagabriel: This raises an interesting question.

`set_advertising_data` literally sets a new value for the property called `advertising_data`.

So, is it better to (1) rename it to `set_legacy_advertising_data` and document that it sets the property called `advertising_data`, or (2) to call it `set_advertising_data` and document that it gets used for legacy advertisements?

Since extended advertising is gated as an "unstable" feature, I'm inclined toward (2), but I want to know what you think.

rust/src/wrapper/hci.rs
whitevegagabriel: Done!

rust/src/wrapper/device/mod.rs
barbibulle: Looks good

rust/src/wrapper/device/mod.rs
```diff
@@ -230,35 +263,159 @@ impl Device {
...
+            AdvertisingStatus::AdvertisingLegacy => return Err(PyErr::new::<PyException, _>("Already advertising in legacy mode. Stop the existing legacy advertisement to start an extended advertisement.")),
+            // Stop the current extended advertisement before advertising with new data.
+            // We could just issue an LeSetExtendedAdvertisingData command, but this approach
+            // allows better future flexibility if `start_advertising_extended` were to change.
+            AdvertisingStatus::AdvertisingExtended => self.stop_advertising_extended().await?,
+            _ => {}
+        }
+
+        // if you change this, don't forget to change the same handle in `stop_advertising_extended`
+        let advertising_handle = 0x00;
```
barbibulle: This will work temporarily for testing, but for real use cases, we'll need to support multiple advertising sets. For a typical device, you want to have one advertising set for legacy payloads (so that older peer devices can still find you) and one or more extended advertisers. Which does complicate the API somewhat (have to manage advertising instances, deal with running out of advertisers, etc.), which is why there's no official Bumble API for this yet (but working on it). The Android API is a good example for what an API could look like (even though it has its own gaps)

whitevegagabriel: Agreed that we will want to support multiple advertising sets in the future.

Once you've figured out how to support extended advertising in the core API, my plan is to completely drop this and adopt what you've implemented (which is why I've marked these additions as "unstable").

But in the meantime, we can at least use this for testing purposes.

bumble/controller.py
```diff
@@ -1000,12 +1000,16 @@ def on_hci_le_set_scan_parameters_command(self, command):
         '''
         See Bluetooth spec Vol 4, Part E - 7.8.10 LE Set Scan Parameters Command
         '''
-        self.le_scan_type = command.le_scan_type
-        self.le_scan_interval = command.le_scan_interval
-        self.le_scan_window = command.le_scan_window
-        self.le_scan_own_address_type = command.own_address_type
-        self.le_scanning_filter_policy = command.scanning_filter_policy
-        return bytes([HCI_SUCCESS])
+        ret = HCI_SUCCESS
```
marshallpierce, line 1003: The "mutable ret" pattern is sometimes needed, but I think in this case it's not worth its cost as a potential bug source in future modifications. Can just return directly at the bottom of the `if`, then return unconditionally (no else needed).

rust/examples/broadcast.rs
marshallpierce: Really, both have their merits. 1 is a better API overall, but 2 will cause less cognitive overhead for people switching from using the Python API.

rust/Cargo.toml
```diff
@@ -86,4 +91,9 @@ pyo3-asyncio-attributes = ["pyo3-asyncio/attributes"]
 dev-tools = ["dep:anyhow", "dep:clap", "dep:file-header", "dep:globset"]
 # separate feature for CLI so that dependencies don't spend time building these
 bumble-tools = ["dep:clap", "anyhow", "dep:anyhow", "dep:directories", "pyo3-asyncio-attributes", "dep:owo-colors", "dep:reqwest", "dep:rusb", "dep:log", "dep:env_logger", "dep:futures"]
+
+# all the unstable features
+unstable = ["unstable_extended_adv"]
+unstable_extended_adv = []
```
marshallpierce, line 97: should add CI build / test runs with and without the features

rust/src/wrapper/device/mod.rs
```diff
@@ -94,7 +120,10 @@ impl Device {
             PyModule::import(py, intern!(py, "bumble.device"))?
                 .getattr(intern!(py, "Device"))?
                 .call((), Some(kwargs))
-                .map(|any| Self(any.into()))
+                .map(|any| Self {
+                    obj: any.into(),
+                    advertising_status: AdvertisingStatus::NotAdvertising,
```
marshallpierce, line 125: what if the device is a real dongle and it's actually already advertising?

rust/src/wrapper/device/mod.rs
```diff
@@ -230,35 +269,162 @@ impl Device {
...
+            scan_request_notification_enable: Enable::Disabled,
+            secondary_advertising_max_skip: 0,
+            secondary_advertising_phy: SecondaryPhyType::Le1m,
+        };
+        self.send_command(extended_advertising_params_cmd.into(), true)
+            .await?;
+
+        // set random address
+        let random_address: packets::Address =
+            self.random_address()?.try_into().map_err(|e| match e {
```
marshallpierce, line 356: we should probably return a struct with interesting metadata like the generated random address instead of just ()

rust/src/wrapper/device/mod.rs
marshallpierce: I think reflecting what Python does is fine, but let's sprinkle in some docs on these methods explaining the state transitions and expected return values.

rust/src/wrapper/device/mod.rs
```diff
@@ -295,6 +461,15 @@ impl Device {
         })?;
         Ok(())
     }
+
+    /// Gets the Device's `random_address` property
+    pub fn random_address(&self) -> PyResult<Address> {
+        Python::with_gil(|py| {
+            self.obj
+                .getattr(py, intern!(py, "random_address"))
```
marshallpierce, line 469: Is this actually random? From some brief but possibly flawed perusal of the code, it looks like this would end up as all zeros?

rust/src/wrapper/device/mod.rs
whitevegagabriel: The device could pretty much be in any state (beyond just advertising/not advertising) until we send HCI_Reset during `power_on`.

We could (a) prevent some operations if `power_on` has not been called (b) reset the controller during `__init__` to ensure a clean slate (c) leave as-is

I wrote this to match how its written in `device.py`, though I do think (b) would be better (with some parameter to disable reset if desired).

(a) could be fine, but could result in more invalid states by accident if not careful

(b) would ideally be implemented in `device.py` since I don't think it makes sense to fix the bug in Rust only, but am curious to hear thoughts from @barbibulle

rust/src/wrapper/device/mod.rs
whitevegagabriel: `random_address` is set to random bytes retrieved from the Controller during `power_on`
```
if self.le_enabled:
    # Set the controller address
    if self.random_address == Address.ANY_RANDOM:
        # Try to use an address generated at random by the controller
        if self.host.supports_command(HCI_LE_RAND_COMMAND):
            # Get 8 random bytes
            response = await self.send_command(
                HCI_LE_Rand_Command(), check_result=True  # type: ignore[call-arg]
            )

            # Ensure the address bytes can be a static random address
            address_bytes = response.return_parameters.random_number[
                :5
            ] + bytes([response.return_parameters.random_number[5] | 0xC0])

            # Create a static random address from the random bytes
            self.random_address = Address(address_bytes)
```

rust/Cargo.toml
whitevegagabriel: Done, using a tool `cargo-all-features`.

rust/examples/broadcast.rs
whitevegagabriel: Unless @barbibulle has some opinion on it, I'm probably going to stick with how it is currently. It may be confusing to some Rust users, but only if you opt-in for the unstable extended API, in which case you are expected to know what you're signing up for.

Plus, when extended is eventually added to `bumble` core, this API may disappear anyway.

marshallpierce (approved):

whitevegagabriel:
@barbibulle are there any other changes I should make to this PR to make it ready for merge?

barbibulle:
> @barbibulle are there any other changes I should make to this PR to make it ready for merge?

Sorry for the delay. Looks good to go. I'll merge it now.

barbibulle (approved):

whitevegagabriel:
> Sorry for the delay. Looks good to go. I'll merge it now.

Not a problem, thank you!

PR #320, opened 2023-10-13
[Pandora] Implement L2CAP service

zxzxwu:
Avatar: https://github.com/google/avatar/pull/57

PR #322, opened 2023-10-25
pandora: implement L2CAP service

uael:


uael:
The issue I'm trying to solve is that we want waiting for incoming connections to be reproducible, while right now it depends on good timing between the link connections and the channel connection request. From a DUT point of view, it's quite hard to make those assumptions and it's almost impossible to implement the Wait API for Fluoride, but from the REF perspective we need maximum control and reproducibility.

Other solutions could be to:
1. make the `l2cap.WaitConnection` scoped at the device level instead of connection level and add a `l2cap.OwnedBy(channel, connection)` API to match channels and connections.
2. add a `l2cap.Register(spec) -> ServerToken` API, then modify `WaitConnection` to take a server token instead of a spec.

Both solutions work but add complexity to the APIs instead of the implementation

zxzxwu:
Can we simply add an `auto_accept=True` argument to `create_server()` method, and `accept()` method to channels? Thus, most flow can keep the same.

uael:
This is a proposed alternative to #320.
It offer more control over the bumble L2CAP layer, but require more structural changes and I'm not sure which path is preferred. Although I think deferred connection accept is a good addition to bumble

zxzxwu:
> This is a proposed alternative to #320. It offer more control over the bumble L2CAP layer, but require more structural changes and I'm not sure which path is preferred. Although I think deferred connection accept is a good addition to bumble

These features make Bumble over complicated. In fact, if some scenarios require a very complicated control like timing, overwriting the entire L2CAP layer inside the usage is simpler since in Python it's very easy to achieve this.

barbibulle:
closing as obsolete

PR #326, opened 2023-11-07
Android benchmarking app

barbibulle:
Android counterpart to `bench.py`. Only RFComm Server, RFComm Client, L2CAP Server and L2CAP Client are supported in this version.
Also, BT addresses must be entered manually, scanning isn't built-in.


extras/android/BtBench/app/src/main/java/com/github/google/bumble/btbench/Receiver.kt
```diff
@@ -0,0 +1,58 @@
...
+    override fun onSequencePacket(packet: SequencePacket) {
+        val received = packet.payload.size + 6
+        bytesReceived += received
+        val now = TimeSource.Monotonic.markNow()
+        lastPacketTime = now
+        viewModel.packetsReceived += 1
+        if (packet.flags and Packet.LAST_FLAG != 0) {
+            Log.info("received last packet")
+            val elapsed = now - startTime
+            viewModel.throughput = (bytesReceived / elapsed.toDouble(DurationUnit.SECONDS)).toInt()
```
mogenson, line 54: Could we add a log message with the throughput value that we can grep for in logcat?

extras/android/BtBench/app/src/main/java/com/github/google/bumble/btbench/Model.kt
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        with(preferences!!.edit()) {
+            putInt(SENDER_PACKET_SIZE_PREF_KEY, senderPacketSize)
+            apply()
+        }
+    }
+
+    fun updateSenderPacketSize() {
+        if (senderPacketSizeSlider < 0.2) {
```
mogenson, line 140: During use, the values for the snap points on the slider were not consistent. For example, two adjacent snap pointes set the packet size to 512. I think this is because of the floating point comparison of setting the slider to a value (eg 0.6F), then checking if the slider is less than that floating point value.

Giving the slider value a larger margin fixed this for me:
```
        if (senderPacketSizeSlider < 0.1F) {
            senderPacketSize = 1
        } else if (senderPacketSizeSlider < 0.3F) {
            senderPacketSize = 256
        } else if (senderPacketSizeSlider < 0.5F) {
            senderPacketSize = 512
        } else if (senderPacketSizeSlider < 0.7F) {
            senderPacketSize = 1024
        } else if (senderPacketSizeSlider < 0.9F) {
            senderPacketSize = 2048
        } else {
            senderPacketSize = 4096
        }
```

extras/android/BtBench/app/src/main/java/com/github/google/bumble/btbench/Receiver.kt
barbibulle: Yes, will do.

barbibulle:
> I've noticed that the app can be suspended when throughput data transmission takes a few seconds:
> 
> ```
> 11-07 16:14:04.508  5080  5242 I btbench.socket-server: got connection
> 11-07 16:14:14.252  1310  1750 D ActivityManager: freezing 5080 com.github.google.bumble.btbench
> ```
> 
> Is it possible to keep this app from being frozen by the Activity Manager? https://source.android.com/docs/core/perf/cached-apps-freezer Maybe keep the display on to keep the app in the foreground?

I usually have a developer setting that keeps the screen on while on USB adb. That does help prevent freezing the app. The other method is to do to the Developer settings and changing "Suspend execution for cached apps" to "disabled".

A cleaner long term solution would be to create a background service in this app, so that it can continue working without beeing frozen, even if you switch to a different app, but that's a task for another PR :-)

mogenson:
@barbibulle when going from phone (RFCOMM client) to USB dongle (RFCOMM server) I've noticed that even though we're calling create insecure RFCOMM socket, I get a pairing request popup on Android. This makes a test hard to automate using the ADB shell launcher commands.

Do you know what maybe causing the bonding attempt?

barbibulle:
> @barbibulle when going from phone (RFCOMM client) to USB dongle (RFCOMM server) I've noticed that even though we're calling create insecure RFCOMM socket, I get a pairing request popup on Android. This makes a test hard to automate using the ADB shell launcher commands.
> 
> Do you know what maybe causing the bonding attempt?

The "insecure" part of setting up an RFComm socket is a bit misleading. It's just about whether you want MITM protection or not, but not whether you want any form of security or not. Android will always require the devices to have a pairing and to encrypt the channel. If you don't bond, it will re-pair every time you setup the socket. To avoid the dialog coming up everytime, what you need to do is to bond the devices. You can do that from either side. From the Bumble side, you can use the `pair` app to initiate pairing and bonding. Or you can use the nrfConnect app on the phone to do that as well.

mogenson (approved):
Thanks for the info and context. Looking forward to having the BtBench app as part of the Bumble project.

PR #327, opened 2023-11-07
Update setup.cfg

Queera:


barbibulle:
It isn't clear what the intent of this PR is. Could you add some note to explain what you're trying to achieve?

PR #328, opened 2023-11-10
Fix pytype error access oneof protobuf field.

yehtsohsin:
request.target would cause attribute error which needs to specify which field it represents.


uael:
@yehtsohsin since the root cause had been fixed I'm closing this one :+1:

PR #330, opened 2023-11-11
Do not exit after pairing is finished

benquike:
Android performs additional service
discovery during pairing, otherwise
pairing fails.

### DIFF
--- a/apps/pair.py
+++ b/apps/pair.py
@@ -291,6 +291,7 @@ async def pair(
     mitm,
     bond,
     ctkd,
+    linger,
     io,
     oob,
     prompt,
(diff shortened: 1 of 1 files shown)


apps/pair.py
```diff
@@ -360,10 +360,8 @@ async def pair(
                         await connection.pair()
                     else:
                         await connection.authenticate()
-                    return
```
barbibulle, line 363: I'd prefer adding a `--linger` option (which could default to `True`) to tell the app to stay connected even after the pairing complete. This way, if a user wants to pair and have the app exit once done, they can still have that behavior.

apps/pair.py
```diff
@@ -360,10 +360,8 @@ async def pair(
...
                 except ProtocolError as error:
                     print(color(f'Pairing failed: {error}', 'red'))
-                    return
```
barbibulle, line 366: Do we still expect the interaction to continue after an error?

barbibulle (approved):

barbibulle:
@benquike There's a small merge conflict. Can you take a look and update the PR so I can merge it?

benquike:
Done, could you please merge it?

PR #332, opened 2023-11-15
Enumify: PhysicalTransport, Role, AddressType

zxzxwu:
As they are widely used as parameters, making them enums can significantly improve the static check.

### DIFF
--- a/bumble/controller.py
+++ b/bumble/controller.py
@@ -25,8 +25,6 @@ import random
 import struct
 from bumble.colors import color
 from bumble.core import (
-    BT_CENTRAL_ROLE,
-    BT_PERIPHERAL_ROLE,
     BT_LE_TRANSPORT,
     BT_BR_EDR_TRANSPORT,
 )
@@ -47,6 +45,7 @@ from bumble.hci import (
     HCI_REMOTE_USER_TERMINATED_CONNECTION_ERROR,
     HCI_VERSION_BLUETOOTH_CORE_5_0,
     Address,
+    Role,
     HCI_AclDataPacket,
     HCI_AclDataPacketAssembler,
     HCI_Command_Complete_Event,
@@ -98,7 +97,7 @@ class CisLink:
 class Connection:
     controller: Controller
     handle: int
-    role: int
+    role: Role
     peer_address: Address
     link: Any
     transport: int
@@ -390,7 +389,7 @@ class Controller:
             connection = Connection(
                 controller=self,
                 handle=connection_handle,
-                role=BT_PERIPHERAL_ROLE,
+                role=Role.PERIPHERAL,
                 peer_address=peer_address,
                 link=self.link,
                 transport=BT_LE_TRANSPORT,
@@ -450,7 +449,7 @@ class Controller:
                 connection = Connection(
                     controller=self,
                     handle=connection_handle,
-                    role=BT_CENTRAL_ROLE,
+                    role=Role.CENTRAL,
                     peer_address=peer_address,
                     link=self.link,
                     transport=BT_LE_TRANSPORT,
@@ -469,7 +468,7 @@ class Controller:
             HCI_LE_Connection_Complete_Event(
                 status=status,
                 connection_handle=connection.handle if connection else 0,
-                role=BT_CENTRAL_ROLE,
+                role=Role.CENTRAL,
                 peer_address_type=le_create_connection_command.peer_address_type,
                 peer_address=le_create_connection_command.peer_address,
                 connection_interval=le_create_connection_command.connection_interval_min,
@@ -693,7 +692,7 @@ class Controller:
                     controller=self,
                     handle=connection_handle,
                     # Role doesn't matter in Classic because they are managed by HCI_Role_Change and HCI_Role_Discovery
-                    role=BT_CENTRAL_ROLE,
+                    role=Role.CENTRAL,
                     peer_address=peer_address,
                     link=self.link,
                     transport=BT_BR_EDR_TRANSPORT,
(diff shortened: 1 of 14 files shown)


bumble/core.py
```diff
@@ -27,11 +28,19 @@
...
+class Role(enum.IntEnum):
+    CENTRAL    = 0
+    PERIPHERAL = 1
 
-BT_BR_EDR_TRANSPORT = 0
-BT_LE_TRANSPORT     = 1
+BT_CENTRAL_ROLE    = Role.CENTRAL
+BT_PERIPHERAL_ROLE = Role.PERIPHERAL
+
+class Transport(enum.IntEnum):
```
barbibulle, line 38: I'd recommend naming this `TransportType`, not to confuse with the `Transport` object from `bumble.transport.common`

zxzxwu: How about `BtTransport`?

uael: Mode ?

barbibulle: Maybe `PhysicalTransport`, since that's the term used in the BT spec?

bumble/core.py
```diff
@@ -28,11 +28,19 @@
...
-BT_LE_TRANSPORT     = 1
+class Role(enum.IntEnum):
+    CENTRAL    = 0
+    PERIPHERAL = 1
+
+BT_CENTRAL_ROLE    = Role.CENTRAL
+BT_PERIPHERAL_ROLE = Role.PERIPHERAL
+
+BT_BR_EDR_TRANSPORT = PhysicalTransport.BR_EDR
+BT_LE_TRANSPORT     = PhysicalTransport.LE
```
AlanRosenthal, line 44: what is the reason for these defines? could they be removed?

zxzxwu: That's for backward compatibility. There have been an significant usage of these primitive constants, so removing them here will be irresponsible.

bumble/core.py
zxzxwu: Done

zxzxwu:
I decide to leave legacy constant references, but redirect them to new stubs - they will be considered enum instead of int, which is still valid for linter.

bumble/controller.py
```diff
@@ -25,8 +25,6 @@
 import struct
 from bumble.colors import color
 from bumble.core import (
-    BT_CENTRAL_ROLE,
-    BT_PERIPHERAL_ROLE,
     BT_LE_TRANSPORT,
```
barbibulle, line 28: Should those be remove as well and import `PhysicalTransport` instead?

bumble/hci.py
```diff
@@ -701,29 +703,22 @@ def phy_list_to_bits(phys: Optional[Iterable[int]]) -> int:
 
 
 # ACL
```
barbibulle, line 705: This comment line should be moved now to line 712

bumble/hci.py
```diff
@@ -732,19 +727,13 @@ def phy_list_to_bits(phys: Optional[Iterable[int]]) -> int:
 
 HCI_LE_PHY_BIT_NAMES = ['LE_1M_PHY', 'LE_2M_PHY', 'LE_CODED_PHY']
 
-HCI_LE_PHY_TYPE_TO_BIT = {
-    HCI_LE_1M_PHY:    HCI_LE_1M_PHY_BIT,
-    HCI_LE_2M_PHY:    HCI_LE_2M_PHY_BIT,
-    HCI_LE_CODED_PHY: HCI_LE_CODED_PHY_BIT
+HCI_LE_PHY_TYPE_TO_BIT: dict[int, int] = {
```
barbibulle, line 730: should this be `dict[Phy, int]`?

bumble/hci.py
```diff
@@ -732,19 +727,13 @@ def phy_list_to_bits(phys: Optional[Iterable[int]]) -> int:
...
+    Phy.LE_1M:    1,
```
barbibulle, line 731: This is not the same value as before (supposed to be the bit offset for PHY bit lists), it will break other methods.

bumble/controller.py
zxzxwu: My initial idea is to keep this PR simple, focused on the major and necessary changes, and have another PR to cleanup them. But it's fine to cleanup now if we agree with the new design.

bumble/hci.py
zxzxwu: Done. This will requires some device API changes as well.

bumble/hci.py
zxzxwu: Done.

bumble/hci.py
zxzxwu: Done.

bumble/controller.py
barbibulle: Sounds good.

bumble/hci.py
```diff
@@ -889,10 +890,13 @@ class CsSubeventAbortReason(OpenIntEnum):
...
-HCI_RANDOM_DEVICE_ADDRESS_TYPE   = 0x01
-HCI_PUBLIC_IDENTITY_ADDRESS_TYPE = 0x02
-HCI_RANDOM_IDENTITY_ADDRESS_TYPE = 0x03
+class AddressType(enum.IntEnum):
+    PUBLIC_DEVICE   = 0x00
+    RANDOM_DEVICE   = 0x01
+    PUBLIC_IDENTITY = 0x02
+    RANDOM_IDENTITY = 0x03
+    # No address.
+    ANONYMOUS       = 0xFF
```
barbibulle, line 899: This doesn't seem to be used anywhere.

barbibulle (approved):

PR #334, opened 2023-11-16
Add support for LE Extended Advertising

zxzxwu:


zxzxwu:
Need to figure out what's the interops between legacy and extended adv.

bumble/device.py
```diff
@@ -1532,6 +1540,126 @@ async def stop_advertising(self) -> None:
             self.advertising = False
             self.auto_restart_advertising = False
 
+    async def start_extended_advertising(
```
barbibulle, line 1543: Thanks for adding this, it's nice to have a simple API to start using extended advertising.
I would recommend marking this as an "interim/experimental" API, because I think we'll need to update it soon (I started working on a (slightly more complicated) model that will allow an object-based approach that also unifies with the legacy advertising API. Users can use this API for now until the new one is ready.

bumble/device.py
```diff
@@ -1532,6 +1540,126 @@ async def stop_advertising(self) -> None:
...
+          target: Directed advertising target. Directed property should be set in advertising_properties arg.
+          own_address_type: own address type to use in the advertising.
+          scan_response: raw scan response. When a non-none value is set, HCI_LE_Set_Extended_Scan_Response_Data_Command will be sent.
+          advertising_data: raw advertising data. When a non-none value is set, HCI_LE_Set_Advertising_Set_Random_Address_Command will be sent.
+
+        Returns:
+          Handle of the new advertising set.
+        """
+
+        adv_handle = (
```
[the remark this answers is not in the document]
zxzxwu: Done.

bumble/device.py
zxzxwu: Add a new decorator to handle this kind of situation.

bumble/device.py
```diff
@@ -1536,6 +1547,135 @@ async def stop_advertising(self) -> None:
...
+                secondary_advertising_phy=1,  # LE 1M
+                advertising_sid=0,
+                scan_request_notification_enable=0,
+            ),  # type: ignore[call-arg]
+            check_result=True,
+        )
+
+        # Set the advertising data if present
+        if advertising_data is not None:
+            await self.send_command(
```
barbibulle, line 1614: If this command is rejected (invalid parameter, if for example the data is too big, or if there are not enough advertising sets allowed in the controller), this will leak the handle. 
The same issue may happen with the following two commands (they could fail, and in that case not only would the handle leak, but also a partial state may be set in the controller)
It would be good to have a way to clean up any intermediate state if any command fails here, before raising an error to the caller.

zxzxwu: Nice catch. It also reminds me that we have to remove the advertising set on stop.

bumble/utils.py
```diff
@@ -444,3 +444,19 @@ def inner(*args, **kwargs):
         return inner
 
     return wrapper
+
+
+def experimental(msg: str):
+    """
+    Throw deprecation warning before execution.
```
barbibulle, line 451: Did you mean "future warning"?

barbibulle (approved):

bumble/utils.py
zxzxwu: Done.

zxzxwu:
Note here: According to Core Spec Vol.4 Part E, Section 3.1.1 - Legacy and extended advertising, Host should not issue Legacy Advertising commands when the Controller supports Extended Advertising, so there is no interops issue.

barbibulle (approved):

PR #339, opened 2023-11-22
Slightly refactor and fix CTKD

zxzxwu:
It seems sample input data provided in the spec is big-endian (just like other AES-CMAC-based functions), but all keys are in little-endian( HCI standard), so they need to be reverse before and after applying AES-CMAC.

### DIFF
--- a/bumble/crypto.py
+++ b/bumble/crypto.py
@@ -21,6 +21,8 @@
 # -----------------------------------------------------------------------------
 # Imports
 # -----------------------------------------------------------------------------
+from __future__ import annotations
+
 import logging
 import operator
 
@@ -29,11 +31,13 @@ from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
 from cryptography.hazmat.primitives.asymmetric.ec import (
     generate_private_key,
     ECDH,
+    EllipticCurvePrivateKey,
     EllipticCurvePublicNumbers,
     EllipticCurvePrivateNumbers,
     SECP256R1,
 )
 from cryptography.hazmat.primitives import cmac
+from typing import Tuple
 
 
 # -----------------------------------------------------------------------------
@@ -46,16 +50,18 @@ logger = logging.getLogger(__name__)
 # Classes
 # -----------------------------------------------------------------------------
 class EccKey:
-    def __init__(self, private_key):
+    def __init__(self, private_key: EllipticCurvePrivateKey) -> None:
         self.private_key = private_key
 
     @classmethod
-    def generate(cls):
+    def generate(cls) -> EccKey:
         private_key = generate_private_key(SECP256R1())
         return cls(private_key)
 
     @classmethod
-    def from_private_key_bytes(cls, d_bytes, x_bytes, y_bytes):
+    def from_private_key_bytes(
+        cls, d_bytes: bytes, x_bytes: bytes, y_bytes: bytes
+    ) -> EccKey:
         d = int.from_bytes(d_bytes, byteorder='big', signed=False)
         x = int.from_bytes(x_bytes, byteorder='big', signed=False)
         y = int.from_bytes(y_bytes, byteorder='big', signed=False)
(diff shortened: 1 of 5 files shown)


bumble/smp.py
```diff
@@ -1098,15 +1098,47 @@ def start_encryption(self, key: bytes) -> None:
...
+            ct2: whether ct2 is supported on both devices.
+        Returns:
+            LE Long Tern Key bytes in little-endian.
+        '''
+        link_key_big_endian = bytes(reversed(link_key))
         ilk = (
-            crypto.h7(salt=SMP_CTKD_H7_BRLE_SALT, w=link_key)
-            if self.ct2
-            else crypto.h6(link_key, b'tmp2')
+            crypto.h7(salt=SMP_CTKD_H7_BRLE_SALT, w=link_key_big_endian)
```
barbibulle, line 1113: I would prefer to update `h6` and `h7` to be implemented the same way the other bt crypto functions are done in `crypto.py`, where the inputs and outputs are in little-endian, and the byte reversing is done internally, as appropriate, when calling internal crypto primitives that are not BT specific (like `aes_cmac`). See `crypto.l6` for example. 
Also, I realize that `crypto.py` uses the expression `bytes(reversed(xyz))` a lot. Since you're editing that part of the code, it might be a good opportunity to use `xxx[::-1]` as a faster way to reverse a `bytes` value, or maybe even make that a local function `def reverse(x: bytes) -> bytes: return x[::-1]` in `crypto.py` for clarity.

zxzxwu: One problem here is that the spec provides a sample data for the derive procedure and I would like to test it. However, it would be strange if we just reimplement the same code in test.

barbibulle: The other tests in `smp_tests.py` also use the test vectors from the spec, exactly as they are given (i.e big-endian), and reverses the bytes before calling the crypto functions (for example, look at `test_f6`, which uses the test vectors from `D.4 f6 LE SC CHECK VALUE GENERATION FUNCTION` in the spec. By having the `crypto.xxx` functions take their inputs and outputs in little-endian, it makes the code in `smp.py` simpler because it doesn't have to do the byte reversing each time before and after calling the crypto functions (see `smp.py` line 1711 and 1712 for example where `crypto.f6` is called.

zxzxwu: Okay, so should we keep a standalone `derive_[ltk/lk]` method? Also shall we move them to `crypto`?

bumble/crypto.py
```diff
@@ -93,35 +99,44 @@
...
     AES-128 ECB, expecting byte-swapped inputs and producing a byte-swapped output.
 
     See Bluetooth spec Vol 3, Part H - 2.2.1 Security function e
     '''
 
-    cipher = Cipher(algorithms.AES(bytes(reversed(key))), modes.ECB())
+    cipher = Cipher(algorithms.AES(reverse(key)), modes.ECB())
     encryptor = cipher.encryptor()
-    return bytes(reversed(encryptor.update(bytes(reversed(data)))))
+    return reverse(encryptor.update(reverse(data)))
```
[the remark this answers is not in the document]
zxzxwu: We may also change it to `modes.CBC(bytes(16))` as the data is too small to stream.

bumble/smp.py
```diff
@@ -1098,15 +1098,45 @@ def start_encryption(self, key: bytes) -> None:
...
+            crypto.h7(salt=SMP_CTKD_H7_LEBR_SALT, w=ltk)
+            if ct2
+            else crypto.h6(ltk, b'tmp1')
+        )
+        return crypto.h6(ilk, b'lebr')
+
+    async def get_link_key_and_derive_ltk(self) -> None:
+        '''Retrieves BR/EDR Link Key from storage and derive it to LE LTK.'''
+        link_key = await self.manager.device.get_link_key(self.connection.peer_address)
+        assert link_key is not None
```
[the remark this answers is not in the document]
zxzxwu: Then it needs to be judged before start/reply pairing request, because here it's an async subtask so we don't have a proper way to propagate the failure to the main task.

barbibulle (approved):

PR #342, opened 2023-11-27
Typing helper

zxzxwu:


zxzxwu:
This PR is not hurry. Reset to draft unless someone needs it.

bumble/helpers.py
```diff
@@ -122,30 +137,33 @@ def on_acl_pdu(self, pdu):
                         if assembler:
                             assembler.on_pdu(l2cap_pdu.payload)
                     else:
-                        psm_string = name_or_number(PSM_NAMES, psm)
                         self.analyzer.emit(
                             f'{color("L2CAP", "green")} [CID={l2cap_pdu.cid}, '
-                            f'PSM={psm_string}]: {l2cap_pdu.payload.hex()}'
+                            f'PSM={PSM_NAMES.get(psm)}]: {l2cap_pdu.payload.hex()}'
```
barbibulle, line 142: This won't raise, but it still doesn't do what you need when encountering a PSM that's not in the name map. In the original code, the function `name_or_number` returns either a name, if found in the map, or a string with the hex number representation of the value. We need to keep support for PSM values not in the map here.

barbibulle (approved):

PR #350, opened 2023-11-28
Add Coordinated Set Identification Service(CSIS)

zxzxwu:
SIRK is only implemented in plaintext mode. As we are not going to support LEA grouping soon, RSI generation is not included.


tests/csip_test.py
```diff
@@ -0,0 +1,74 @@
+# Copyright 2021-2023 Google LLC
```
zxzxwu, line 1: I put `csip_test.py` under `test/` because `test_utils.py` is not accessible from a subdirectory.
Might be a better solution?

barbibulle: That sounds fine to me.

barbibulle (approved):

barbibulle (approved):

PR #356, opened 2023-11-30
Add Published Audio Capabilities Service

zxzxwu:


barbibulle (approved):

bumble/profiles/bap.py
```diff
@@ -0,0 +1,490 @@
...
+        source_pac: Sequence[PacRecord] = [],
+        source_audio_locations: Optional[AudioLocation] = None,
+    ) -> None:
+        characteristics = []
+
+        self.supported_audio_contexts = gatt.Characteristic(
+            uuid=gatt.GATT_SUPPORTED_AUDIO_CONTEXTS_CHARACTERISTIC,
+            properties=gatt.Characteristic.Properties.READ,
+            permissions=gatt.Characteristic.Permissions.READABLE,
+            value=struct.pack('<HH', supported_sink_context, supported_source_context),
```
barbibulle, line 397: NOTE (for this and other characteristics in general): if the value could potentially be changed while the characteristic is published by a server, you could use a `PackedCharacteristicAdapter` with a `pack_format` (or other type of adapter), to do the ->bytes encoding on the fly as the characteristic is read.

bumble/profiles/bap.py
```diff
@@ -0,0 +1,490 @@
...
+    sink_audio_locations: Optional[gatt_client.CharacteristicProxy] = None
+    source_pac: Optional[gatt_client.CharacteristicProxy] = None
+    source_audio_locations: Optional[gatt_client.CharacteristicProxy] = None
+    available_audio_contexts: gatt_client.CharacteristicProxy
+    supported_audio_contexts: gatt_client.CharacteristicProxy
+
+    def __init__(self, service_proxy: gatt_client.ServiceProxy):
+        self.service_proxy = service_proxy
+
+        self.available_audio_contexts = service_proxy.get_characteristics_by_uuid(
```
barbibulle, line 465: A future enhancement here would be to provide an adapter to convert the raw bytes into the actual fields (in this case here, I think `available_audio_contexts` get split into `available_sink_context` and `available_source_context` fields. Either a `PackedCharacteristicAdapter` or `MappedCharacteristicAdapter` or some other type of adapter would do that automatically (see `battery_service.py` or `header_rate_service.py` for examples).

barbibulle (approved):

PR #359, opened 2023-12-01
Audio Stream Control Service

zxzxwu:
Based on https://github.com/google/bumble/pull/356

Status: (Sink only) Play, pause, decoding are verified.

### DIFF
--- a/bumble/profiles/bap.py
+++ b/bumble/profiles/bap.py
@@ -34,6 +34,11 @@ from bumble import gatt_client
 
 
 # -----------------------------------------------------------------------------
+# Logging
+# -----------------------------------------------------------------------------
+logger = logging.getLogger(__name__)
+
+# -----------------------------------------------------------------------------
 # Constants
 # -----------------------------------------------------------------------------
 
@@ -976,7 +981,7 @@ class AseStateMachine(gatt.Characteristic):
         return (AseResponseCode.SUCCESS, AseReasonCode.NONE)
 
     def on_release(self) -> Tuple[AseResponseCode, AseReasonCode]:
-        if self.state != AseStateMachine.State.IDLE:
+        if self.state == AseStateMachine.State.IDLE:
             return (
                 AseResponseCode.INVALID_ASE_STATE_MACHINE_TRANSITION,
                 AseReasonCode.NONE,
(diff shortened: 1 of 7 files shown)


zxzxwu:
In run_unicast_server.py, `dlc3` from https://github.com/google/liblc3 is used to decode the LC3 pdu. Seems like it needs a relatively large buffer to decode, so the latency is pretty high.

It would be easier to use if we can pack a liblc3 interface into Python, but I am not familiar with this part.

bumble/profiles/bap.py
```diff
@@ -452,6 +759,417 @@ def __init__(
...
+
+    def __init__(
+        self,
+        role: AudioRole,
+        ase_id: int,
+        service: AudioStreamControlService,
+    ) -> None:
+        self.service = service
+        self.ase_id = ase_id
+        self.state = AseStateMachine.State.IDLE
```
barbibulle, line 810: suggestion: `state` could be a property with a setter, so that when you change it it would log `State change from <x> -> <y>`, to make it easy to see state changes in the log trace

bumble/profiles/bap.py
```diff
@@ -452,6 +759,417 @@ def __init__(
         super().__init__(characteristics)
 
 
+class AseStateMachine(gatt.Characteristic):
```
barbibulle, line 762: Would it make sense to have the characteristic as a property of `AseStateMachine` rather than directly inheriting from `gatt.Characteristic`? The characteristic value (read and write) can easily be delegated to this object.

bumble/profiles/bap.py
```diff
@@ -452,6 +759,417 @@ def __init__(
...
+                for id in sink_ase_id
+            },
+            **{
+                id: AseStateMachine(role=AudioRole.SOURCE, ase_id=id, service=self)
+                for id in source_ase_id
+            },
+        }  # ASE state machines, by ASE ID
+
+        for ase in self.ase_state_machines.values():
+            print(ase.ase_id)
```
barbibulle, line 1089: use logger instead of print.

barbibulle (approved):

bumble/profiles/bap.py
zxzxwu: Done

bumble/profiles/bap.py
zxzxwu: Removed.

bumble/profiles/bap.py
zxzxwu: Maybe in another PR? Currently the design can avoid ASCS main service accessing members of characteristics.

PR #365, opened 2023-12-05
driver: Implement driver for Intel Ax210

SilverBzH:


bumble/drivers/intel.py
```diff
@@ -0,0 +1,343 @@
...
+
+@dataclass
+class IntelTLV:
+    type: int
+    lenght: int
+    value: bytearray
+
+
+@dataclass
+class IntelVersionTlv:
```
barbibulle, line 107: To be consistent, either this class should be named `IntelVersionTLV`, or `IntelTLV` should be `IntelTlv`

bumble/drivers/intel.py
```diff
@@ -0,0 +1,343 @@
...
+    cnvi_top: Optional[IntelTLV]
+    cnvr_top: Optional[IntelTLV]
+    cnvi_bt: Optional[IntelTLV]
+    img_type: Optional[IntelTLV]
+    sbe_type: Optional[IntelTLV]
+
+
+def parse_intel_version_tlvs(data: bytearray) -> IntelVersionTlv:
+    tlvs = {
```
barbibulle, line 116: Maybe using a subclass of `enum.IntEnum` would be easier for this.

bumble/drivers/intel.py
```diff
@@ -0,0 +1,343 @@
...
+                if tlv_type == type_code:
+                    parsed_tlvs[name] = IntelTLV(tlv_type, tlv_length, tlv_value)
+                    break
+
+            i += 2 + tlv_length
+        except IndexError:
+            logging.error("TLV parse error")
+            continue
+
+    return IntelVersionTlv(
```
barbibulle, line 142: suggestion: you could have all the fields default to None. Then in the while loop, just populate a dict for the properties you find, then invoke the constructor as `IntelVersionTlv(**tlvs)`

bumble/drivers/intel.py
```diff
@@ -0,0 +1,343 @@
...
+        cnvi_bt=parsed_tlvs.get('cnvi_bt'),
+        img_type=parsed_tlvs.get('img_type'),
+        sbe_type=parsed_tlvs.get('sbe_type'),
+    )
+
+
+async def secure_send(host, fragment_type: int, plen: int, param: bytes):
+    while plen > 0:
+        fragment_len = MAX_FRAGMENT_PAYLOAD if plen > MAX_FRAGMENT_PAYLOAD else plen
+        cmd_param = bytearray(fragment_len + 1)
```
barbibulle, line 154: suggest: `bytearray` isn't needed here, you can just construct: `cmd_param = bytes([fragment_type]) + param[:fragment_len]`

bumble/drivers/intel.py
```diff
@@ -0,0 +1,343 @@
...
+        cmd_param[0] = fragment_type
+        cmd_param[1 : 1 + fragment_len] = param[:fragment_len]
+
+        # await host.send_command(Hci_Intel_Secure_Send_Command(param=cmd_param))  # type: ignore
+        host.send_hci_packet(Hci_Intel_Secure_Send_Command(param=cmd_param))  # type: ignore
+        await asyncio.sleep(0.002)
```
barbibulle, line 160: `0.002` should probably be a named constant.

bumble/drivers/intel.py
```diff
@@ -0,0 +1,343 @@
...
+    cnvr_top_type = cnvr_top & 0x00000FFF
+
+    upper_name = intel_cnvx_top_pack_swab(cnvi_top_type, cnvi_top_step)
+    lower_name = intel_cnvx_top_pack_swab(cnvr_top_type, cnvr_top_step)
+
+    return f"ibt-{upper_name:04x}-{lower_name:04x}.sfi"
+
+
+def prepare_firmware(version: IntelVersionTlv) -> bytes:
+    if not sys.platform.startswith("linux"):
```
barbibulle, line 290: It would be nice to be able to load FW images from other places than the default Linux location, like for the Realtek driver (see rtk.py `Driver.find_binary_path`)

bumble/drivers/intel.py
SilverBzH: This is for debugging purpose here. I think David would like to implement the credit on HCI to get this faster, otherwise the update is really long

PR #366, opened 2023-12-06
Add advertiser classes and handle adv set terminated events

zxzxwu:
* Convert hci.OwnAddressType to enum
* Add LegacyAdvertiser and ExtendedAdvertiser classes
* Rename start/stop_advertising() => start/stop_legacy_advertising()
* Handle HCI_Advertising_Set_Terminated
* Properly restart advertisement on disconnection


zxzxwu:
Maybe we should merge this request after we have a way to test extended advertising?

bumble/device.py
```diff
@@ -1679,14 +1733,18 @@ async def start_extended_advertising(
...
+            )
 
         adv_handle = -1
         # Find a free handle
         for i in range(
             DEVICE_MIN_EXTENDED_ADVERTISING_SET_HANDLE,
             DEVICE_MAX_EXTENDED_ADVERTISING_SET_HANDLE + 1,
         ):
-            if i not in self.extended_advertising_handles:
+            if i not in self.extended_advertisers.keys():
```
barbibulle, line 1747: nit: `if i not in self.extended_advertisers` should be sufficient.

bumble/host.py
```diff
@@ -721,6 +721,14 @@ def on_hci_le_advertising_report_event(self, event):
     def on_hci_le_extended_advertising_report_event(self, event):
         self.on_hci_le_advertising_report_event(event)
 
+    def on_hci_le_advertising_set_terminated_event(self, event):
```
barbibulle, line 724: It would be nice to make the advertiser itself an event emitter, and then have the advertiser emit a 'termination' event.

bumble/host.py
```diff
@@ -721,6 +721,14 @@ def on_hci_le_advertising_report_event(self, event):
...
+        self.emit(
+            'advertising_set_terminated',
```
barbibulle, line 726: To be consistent, the pattern is that event names are nouns, following the same convention as for Web events, so for example we use 'disconnection' rather than 'disconnected', so here this would be 'advertisting_set_termination'.

barbibulle (approved):

bumble/host.py
zxzxwu: Done.

bumble/device.py
zxzxwu: I think this could provide a better readability?

barbibulle: Either way is fine, but in general I think it is more readable to use the standard/idiomatic way of using collections.
For looking up a key in a dict, the recommended way in most style guides is to just use `key in dict` (see https://google.github.io/styleguide/pyguide.html#284-decision for example).

PR #368, opened 2023-12-09
support drivers that can't use reset directly.

barbibulle:
This should allow a driver to interact with a controller, via the `host`, before any command, including RESET is issued.
See the example with the Realtek driver. 
Note that if a driver needs to send a command and receive a response before RESET is performed, it must set `host.ready = True` to allow the inbound packet processing to accept the packet even if no RESET has been done (by default, the host ignores packets from the controller that are received before a RESET).

Also included is a generic way to add HCI metadata to an HCI transport source. Previously, only certain transport sources would set metdata (USB would set `vendor_id` and `product_id`, which the Realtek driver uses to determine if it should try
to load a FW image). Now, a driver can be "forced" for a specific HCI source, when no relevant metadata is available (for 
example, when the controller is only visible via a transport that does not set metadata, like a serial port or TCP socket). 
Any number of metadata `key=value` pairs can be added to a transport name, in square brackets, after the transport type.
Example: `usb:[driver=rtk]0` would "inject" a "driver" metadata entry for the usb transport, which the Realtek driver will see as a request that this driver be activated, even if the `vendor_id` and `product_id` are not present, or don't match the internal allow-list.

### DIFF
--- a/bumble/drivers/__init__.py
+++ b/bumble/drivers/__init__.py
@@ -60,12 +60,23 @@ class Driver(abc.ABC):
 # Functions
 # -----------------------------------------------------------------------------
 async def get_driver_for_host(host):
-    """Probe all known diver classes until one returns a valid instance for a host,
-    or none is found.
+    """Probe diver classes until one returns a valid instance for a host, or none is
+    found.
+    If a "driver" HCI metadata entry is present, only that driver class will be probed.
     """
-    if driver := await rtk.Driver.for_host(host):
-        logger.debug("Instantiated RTK driver")
-        return driver
+    driver_classes = {"rtk": rtk.Driver}
+    if driver_name := host.hci_metadata.get("driver"):
+        # Only probe a single driver
+        probe_list = [driver_name]
+    else:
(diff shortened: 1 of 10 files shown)


SilverBzH (approved):

bumble/drivers/__init__.py
```diff
@@ -60,12 +60,23 @@ async def init_controller(self):
...
+    if driver_name := host.hci_metadata.get("driver"):
+        # Only probe a single driver
+        probe_list = [driver_name]
+    else:
+        # Probe all drivers
+        probe_list = driver_classes.keys()
+
+    for driver_name in probe_list:
+        logger.debug(f"Probing {driver_name} driver class")
+        if driver := await rtk.Driver.for_host(host):
```
uael, line 77: `driver_classes[driver_name].for_host(..)` ?

uael (approved):

bumble/drivers/__init__.py
barbibulle: Yes! (oops)

PR #369, opened 2023-12-11
Pandora host: support advertising interval in advertise

SilverBzH:


bumble/pandora/host.py
```diff
@@ -285,8 +285,9 @@ async def Advertise(
             raise NotImplementedError(
                 "TODO: add support for extended advertising in Bumble"
             )
-        if request.interval:
-            raise NotImplementedError("TODO: add support for `request.interval`")
+        if advertising_interval := request.interval:
+            self.device.config.advertising_interval_min = advertising_interval
+            self.device.config.advertising_interval_max = advertising_interval
```
barbibulle, line 290: Is `interval_max` maybe computable from `interval` and `interval_range`?

barbibulle (approved):

bumble/pandora/host.py
SilverBzH: Yes you right I'll add the logic

PR #371, opened 2023-12-13
driver: allow send hci_packet for drivers

SilverBzH:
Allow to send hci_packet for drivers that don't support the HCI_RESET before the firmware update

Correct a typo in the doc

### DIFF
--- a/bumble/host.py
+++ b/bumble/host.py
@@ -139,6 +139,7 @@ class Host(AbortableEventEmitter):
         super().__init__()
 
         self.hci_metadata = {}
+        self.controller_init_required = False
         self.ready = False  # True when we can accept incoming packets
         self.connections = {}  # Connections, by connection handle
         self.pending_command = None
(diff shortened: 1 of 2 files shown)


bumble/host.py
```diff
@@ -139,6 +139,7 @@ def __init__(
         super().__init__()
 
         self.hci_metadata = {}
+        self.controller_init_required = False
```
barbibulle, line 142: It's not clear what the purpose of this new property is that couldn't be done with the `ready` property already. For drivers that have to send commands and receive responses before a reset has been done, they should set the `host.ready` property to `True` to allow incoming packets to be processed. Drivers are responsible for leaving the controller in an initialized and reset state when returning from `driver.init_controller()`, with `host.ready` therefore set to `True` as well.

SilverBzH: Ok sound good I'll do the appropriate modifications then

PR #374, opened 2023-12-14
Complete CSIP and CAP

zxzxwu:
Changes:
* RSI generation
* SIRK encryption
* Delegated SIRK reader
* Prand and random address generation

TODO:
* Encrypted SIRK. Requiring a change: https://github.com/google/bumble/pull/374#issuecomment-1857170345, so it would be implemented in another CL.

Test:
* csip_test, cap_test
* Manually pairing with Android CSIS client

### DIFF
--- a/bumble/crypto.py
+++ b/bumble/crypto.py
@@ -101,6 +101,16 @@ class EccKey:
 
 
 # -----------------------------------------------------------------------------
+def generate_prand() -> bytes:
+    '''Generates random 3 bytes, with the 2 most significant bits of 0b01.
+
+    See Bluetooth spec, Vol 6, Part E - Table 1.2.
+    '''
+    prand_bytes = secrets.token_bytes(6)
+    return prand_bytes[:2] + bytes([(prand_bytes[2] & 0b01111111) | 0b01000000])
+
+
+# -----------------------------------------------------------------------------
 def xor(x: bytes, y: bytes) -> bytes:
     assert len(x) == len(y)
     return bytes(map(operator.xor, x, y))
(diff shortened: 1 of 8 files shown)


zxzxwu:
One question: LTK and LK must be retrieved asynchronously, but GATT read must be synchronous. How to resolve this?

barbibulle:
> One question: LTK and LK must be retrieved asynchronously, but GATT read must be synchronous. How to resolve this?I 

I have a WIP local branch that implements support for async characteristic values (both read and write). I'll polish it off and submit a PR shortly, that should solve your use case here.

bumble/crypto.py
```diff
@@ -100,6 +100,49 @@ def dh(self, public_key_x: bytes, public_key_y: bytes) -> bytes:
...
+    '''Generates random 3 bytes, with the 2 most significant bits of 0b01.
+
+    See Bluetooth spec, Vol 6, Part E - Table 1.2.
+    '''
+    prand_bytes = secrets.token_bytes(6)
+    return prand_bytes[:2] + bytes([(prand_bytes[2] & 0b01111111) | 0b01000000])
+
+
+# -----------------------------------------------------------------------------
+def generate_random_static_address() -> bytes:
```
barbibulle, line 114: Maybe this function should return an `Address`, with the right address type? (Also, this is a function that *uses* crypto, but isn't a crypto function itself, so maybe it should live somewhere else?

barbibulle (approved):

bumble/crypto.py
zxzxwu: I think rand / prand is also considered crypto somehow? But yes, they're better to be classmethods of `Address`. The only exception is prand - hosting it in hci/Address is not very proper, but there isn't a module more proper than crypto(from my point of view).

zxzxwu: Moving address methods to https://github.com/google/bumble/pull/379

PR #376, opened 2023-12-15
Manage lifecycle of CIS and SCO links in host

zxzxwu:


bumble/device.py
```diff
@@ -3071,26 +3071,12 @@ async def create_cis(self, cis_acl_pairs: List[Tuple[int, int]]) -> List[CisLink
...
-            raise HCI_StatusError(result)
-
-        pending_cis_establishments: Dict[int, asyncio.Future[CisLink]] = {}
-        for cis_handle, _ in cis_acl_pairs:
-            pending_cis_establishments[
-                cis_handle
-            ] = asyncio.get_running_loop().create_future()
-
         with closing(EventWatcher()) as watcher:
+            pending_cis_establishments: Dict[int, asyncio.Future[CisLink]] = {}
```
barbibulle, line 3075: suggest: consider using a dict comprenhension to instance the dict as a single statement.

bumble/device.py
```diff
@@ -3071,26 +3071,12 @@ async def create_cis(self, cis_acl_pairs: List[Tuple[int, int]]) -> List[CisLink
...
+            for cis_handle, _ in cis_acl_pairs:
+                pending_cis_establishments[
+                    cis_handle
+                ] = asyncio.get_running_loop().create_future()
 
             @watcher.on(self, 'cis_establishment')
             def on_cis_establishment(cis_link: CisLink) -> None:
```
barbibulle, line 3082: nit: the `None` default isn't necessary for a `get`

bumble/host.py
```diff
@@ -665,18 +684,23 @@ def on_hci_connection_complete_event(self, event):
...
+            ) is None:
+                logger.warning('!!! DISCONNECTION COMPLETE: unknown handle')
+                return
+
             logger.debug(
                 f'### DISCONNECTION: [0x{event.connection_handle:04X}] '
                 f'{connection.peer_address} '
                 f'reason={event.reason}'
             )
-            del self.connections[event.connection_handle]
```
barbibulle, line 679: Why not delete the entry here?

bumble/host.py
```diff
@@ -665,18 +684,23 @@ def on_hci_connection_complete_event(self, event):
             )
 
     def on_hci_disconnection_complete_event(self, event):
-        # Find the connection
```
barbibulle, line 668: Wouldn't it be an error to receive a disconnection event for an unknown handle? Is there a case where we would expect that to happen and would want to still emit an event?

zxzxwu:
I will work on a local emulation of CIS and SCO. Will merge this PR later.

barbibulle (approved):

endrik-exe:
Hi @zxzxwu 
i don't know if this a propper place for asking question or not.

I have question with sco link.
i am testing HFP implementation and SCO Connection using Bumble.

my question is is the SCO Stream already implemented or not in bumble?

because when i try to register event `scoLink.on('pdu')` it is still not receiving any event.
how to receive or send audio data via the sco link?

zxzxwu:
> Hi @zxzxwu i don't know if this a propper place for asking question or not.
> 
> I have question with sco link. i am testing HFP implementation and SCO Connection using Bumble.
> 
> my question is is the SCO Stream already implemented or not in bumble?
> 
> because when i try to register event `scoLink.on('pdu')` it is still not receiving any event. how to receive or send audio data via the sco link?

Not all controller support SCO over HCI. Unlike A2DP which is mandatorily required to be transmitted over ACL, SCO is quite customized. Many chip vendors integrate Bluetooth controller with their own audio DSP and disables the HCI path because it's quite complicated and in a bad performance usually. You may try with different SCO parameters to see if that work, but I guess no :/

PR #377, opened 2023-12-16
Add functions/tool for gen/verifying BLE IRK/RPA

benquike:


zxzxwu:
I am wondering whether this kind of use case should be in Bumble? You can implement the tool in your own codebase like AOSP, since it doesn't involve Bumble's internal lifecycle and functionality.

zxzxwu:
But still Bumble has many "tools"? Maybe we need a guideline about what should be in Bumble, and what is not.

benquike:
> I am wondering whether this kind of use case should be in Bumble? You can implement the tool in your own codebase like AOSP, since it doesn't involve Bumble's internal lifecycle and functionality.

Do you plan to add BLE privacy < 1.2 with random address resolution in host?

zxzxwu:
> Do you plan to add BLE privacy < 1.2 with random address resolution in host?

Yes, Bumble already supports Host-side resolution.
https://github.com/google/bumble/blob/main/bumble/device.py#L3228

But it doesn't handle advertising reports(only in connection), so that's something we can enhance Bumble.

barbibulle:
> But still Bumble has many "tools"? Maybe we need a guideline about what should be in Bumble, and what is not.

Since Bumble is a general purpose project for developers working on Bluetooth-related things, I think it's good to host a collection of tools that relate to that, including very narrow tools like this that are useful for people working on the internals of the stack(s).

barbibulle:
> > Do you plan to add BLE privacy < 1.2 with random address resolution in host?
> 
> Yes, Bumble already supports Host-side resolution. https://github.com/google/bumble/blob/main/bumble/device.py#L3228
> 
> But it doesn't handle advertising reports(only in connection), so that's something we can enhance Bumble.

There is support for address resolution for advertising reports in some places, like the `scan.py` tool (ex: https://github.com/google/bumble/blob/main/apps/scan.py#L60 ). 
But we could definitely make this easier by doing automatic address resolution in a shared way, so that apps that want to scan don't have to instantiate their own resolver. I'll add support for that in my WIP branch in which I'm adding better support for RPA.

bumble/helpers.py
```diff
@@ -232,3 +233,26 @@ def __init__(
         )
         self.host_to_controller_analyzer.peer = self.controller_to_host_analyzer
         self.controller_to_host_analyzer.peer = self.host_to_controller_analyzer
+
+
+def generate_irk() -> bytes:
+    return crypto.r()
+
+
+def gen_rpa_with_irk(irk: bytes) -> bytes:
```
barbibulle, line 242: suggest: naming this `generare_rpa_with_irk` to be consistent with `generate_irk`.

apps/ble_rpa_tool.py
```diff
@@ -0,0 +1,76 @@
...
+    irk_bytes = bytes.fromhex(irk)
+    rpa = gen_rpa_with_irk(irk_bytes)
+    print(addr_bytes_to_str(rpa))
+
+
+@click.command()
+@click.argument("irk", type=str)
+@click.argument("rpa", type=str)
+def verify_rpa(irk: str, rpa: str) -> None:
+    rpa_bytes = addr_str_to_bytes(rpa)
```
barbibulle, line 63: You can use the `Address` class for this: `bytes(hci.Address(rpa))`

apps/ble_rpa_tool.py
```diff
@@ -0,0 +1,76 @@
...
+def gen_irk() -> None:
+    print(generate_irk().hex())
+
+
+@click.command()
+@click.argument("irk", type=str)
+def gen_rpa(irk: str) -> None:
+    irk_bytes = bytes.fromhex(irk)
+    rpa = gen_rpa_with_irk(irk_bytes)
+    print(addr_bytes_to_str(rpa))
```
barbibulle, line 56: You can use the `hci.Address` class for this: `hci.Address(rpa).to_string(with_type_qualifier=False)`.

bumble/helpers.py
barbibulle, line 242: Maybe this should return an `hci.Address` object.

benquike: Done

apps/ble_rpa_tool.py
benquike: Done

apps/ble_rpa_tool.py
benquike: Done

bumble/helpers.py
benquike: Done

barbibulle (approved):

benquike:
OK to merge?

barbibulle:
> OK to merge?

PR #379 was merged earlier, it includes functions for generating private addresses. Would be good to update this PR by removing `gen_rpa_with_irk` and using the new `Address` methods instead.

benquike:
> > OK to merge?
> 
> PR #379 was merged earlier, it includes functions for generating private addresses. Would be good to update this PR by removing `gen_rpa_with_irk` and using the new `Address` methods instead.

Done

PR #378, opened 2023-12-22
Improve the linger option of bumble-pair

benquike:
No matter pairing fails or not, make linger effective


barbibulle (approved):

benquike:
OK to merge?

PR #379, opened 2023-12-22
Add random address generation methods

zxzxwu:


bumble/hci.py
```diff
@@ -1881,6 +1883,45 @@ def parse_address_preceded_by_type(data, offset):
         address_type = data[offset - 1]
         return Address.parse_address_with_type(data, offset, address_type)
 
+    @classmethod
+    def generate_static(cls) -> Address:
```
barbibulle, line 1887: We should try to be consistent in the naming of these 3 methods.
To be very clear, even if a bit long, the full names would be:
`generate_static_address`, `generate_non_resolvable_private_address`, and `generate_resolvable_private_address`. (we can omit the term `random` here I think since we only generate random addresses).
An alternative would be to drop the `_address` suffix. But in any case, these methods should either all have the `_address` suffix, or none of them should have it.

barbibulle (approved):

barbibulle (approved):

PR #380, opened 2023-12-22
support per-transport ACL queues

barbibulle:
This was a TODO item: support separate HCI ACL packet queues for LE and Classic transports.
This also increases the buffer size for USB, since some controllers support large buffers for Classic (> 2600 bytes!).


bumble/host.py
```diff
@@ -553,26 +583,25 @@ def on_hci_command_complete_event(self, event):
...
-        total_packets = sum(event.num_completed_packets)
-        if total_packets <= self.acl_packets_in_flight:
-            self.acl_packets_in_flight -= total_packets
-            self.check_acl_packet_queue()
-        else:
-            logger.warning(
-                color(
-                    '!!! {total_packets} completed but only '
-                    f'{self.acl_packets_in_flight} in flight'
+        for i, connection_handle in enumerate(event.connection_handles):
```
zxzxwu, line 594:
```suggestion
        for num_completed_packet, connection_handle in zip(event.num_completed_packets, event.connection_handles):
```

barbibulle: Done

zxzxwu (approved):

PR #381, opened 2023-12-24
Support non-directed address generation offload

zxzxwu:
Address generation offloading uses peer address as an entry of IRK, even in non-directed usage. By adding an empty entry containing local IRK to controller resolving list, controllers are able to generate RPA without pairing.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -21,6 +21,7 @@ import functools
 import json
 import asyncio
 import logging
+import secrets
 from contextlib import asynccontextmanager, AsyncExitStack, closing
 from dataclasses import dataclass
 from collections.abc import Iterable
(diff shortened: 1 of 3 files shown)


zxzxwu:
@benquike @uael @hchataing It seems rootcanal doesn't handle address resolution offloading properly?

zxzxwu:
Maybe the rootcanal is fine, but Bumble tries to add resolving list when there is active connection...

zxzxwu:
```
root-canal D 12-26 15:45:46.731 link_layer_controller.cc:3221         7  Legacy advertising ignored by initiator because the advertising address da:4c:10:de:17:01[PUBLIC_IDENTITY_ADDRESS] does not match the peer address da:4c:10:de:17:01[PUBLIC_DEVICE_ADDRESS]
```

hchataing:
I just updated v1.6.0 of RootCanal to pypi.

barbibulle (approved):

PR #382, opened 2023-12-26
extended advertising v2

barbibulle:
This is an enhanced API for extended advertising which allows:
1/ using extended advertising commands for legacy advertising (keeping the original advertising API as-is)
2/ creating advertising set objects, which can be independently started/stopped/updated, using a new API.
The new/extended API also allows creating advertisements with legacy PDUs (very typical when you want to be compatible with older devices/phones) in addition to newer ones.

The unit tests aren't complete yet, because they will be easier to write once we merge the PR with basic support for extended advertising in the local controller class. Using Host mocks isn't a great way to test, because you need the full logic of a Controller to ensure that the sequence of events (for example, you need responses for some commands, not just mocking accepting a command and mocking some events), as well as the declared/supported properties, behave correctly.
A future PR will also include a way to test with Root Canal, when available (only on some platforms), but that's out of scope for this PR.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -3632,7 +3632,7 @@ class Device(CompositeEventEmitter):
         self_address = None
         if role == HCI_CENTRAL_ROLE:
             own_address_type = self.connect_own_address_type
-            assert own_address_type
+            assert own_address_type is not None
         else:
             if self.supports_le_extended_advertising:
                 # We'll know the address when the advertising set terminates,
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -664,6 +664,7 @@ class AdvertisingSet(EventEmitter):
         self.advertising_data = advertising_data
 
     async def set_scan_response_data(self, scan_response_data: bytes) -> None:
+        # pylint: disable=line-too-long
         if (
             scan_response_data
             and not self.advertising_parameters.advertising_event_properties.is_scannable
@@ -674,7 +675,6 @@ class AdvertisingSet(EventEmitter):
             )
             return
 
-        # pylint: disable=line-too-long
         await self.device.send_command(
             HCI_LE_Set_Extended_Scan_Response_Data_Command(
                 advertising_handle=self.advertising_handle,
@@ -2016,6 +2016,7 @@ class Device(CompositeEventEmitter):
         if self.supports_le_extended_advertising:
             # Use extended advertising commands with legacy PDUs.
             self.legacy_advertising_set = await self.create_advertising_set(
+                auto_start=True,
                 auto_restart=auto_restart,
                 random_address=self.random_address,
                 advertising_parameters=AdvertisingParameters(
@@ -2036,8 +2037,6 @@ class Device(CompositeEventEmitter):
                     self.scan_response_data if advertising_type.is_scannable else b''
                 ),
             )
-
-            await self.legacy_advertising_set.start()
         else:
             # Use legacy commands.
             self.legacy_advertiser = LegacyAdvertiser(
@@ -2068,8 +2067,37 @@ class Device(CompositeEventEmitter):
         scan_response_data: bytes = b'',
         periodic_advertising_parameters: Optional[PeriodicAdvertisingParameters] = None,
         periodic_advertising_data: bytes = b'',
+        auto_start: bool = True,
         auto_restart: bool = False,
     ) -> AdvertisingSet:
+        """
+        Create an advertising set.
+
+        This method allows the creation of advertising sets for controllers that
+        support extended advertising.
+
+        Args:
+          advertising_parameters:
+            The parameters to use for this set. If None, default parameters are used.
+          random_address:
+            The random address to use (only relevant when the parameters specify that
+            own_address_type is random).
+          advertising_data:
+            Initial value for the set's advertising data.
+          scan_response_data:
+            Initial value for the set's scan response data.
+          periodic_advertising_parameters:
+            The parameters to use for periodic advertising (if needed).
+          periodic_advertising_data:
+            Initial value for the set's periodic advertising data.
+          auto_start:
+            True if the set should be automatically started upon creation.
+          auto_restart:
+            True if the set should be automatically restated after a disconnection.
+
+        Returns:
+          An AdvertisingSet instance.
+        """
         # Allocate a new handle
         try:
             advertising_handle = next(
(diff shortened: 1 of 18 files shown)


bumble/device.py
```diff
@@ -3225,6 +3514,73 @@ async def indicate_subscriber(self, connection, attribute, value=None, force=Fal
...
+        # Setup auto-restart of the advertising set if needed.
+        if advertising_set.auto_restart:
+            connection.once(
+                'disconnection',
+                lambda _: self.abort_on('flush', advertising_set.start()),
+            )
+
+        self.emit_le_connection(connection)
+
+    def emit_le_connection(self, connection: Connection) -> None:
```
zxzxwu, line 3570: Should this be a private method?

bumble/device.py
```diff
@@ -1616,290 +1938,259 @@ def supports_le_phy(self, phy):
...
-            ):
-                await self.send_command(
-                    HCI_LE_Set_Advertising_Set_Random_Address_Command(
-                        advertising_handle=adv_handle,
-                        random_address=self.random_address,
-                    ),
-                    check_result=True,
-                )
+            if periodic_advertising_parameters:
+                # TODO: call LE Set Periodic Advertising Parameters command
```
zxzxwu, line 2152: nit: raise NotImplementedError here

bumble/device.py
```diff
@@ -433,37 +440,344 @@ def is_directed(self):
...
+        await self.device.send_command(
+            HCI_LE_Set_Advertising_Set_Random_Address_Command(
+                advertising_handle=self.advertising_handle,
+                random_address=(random_address or self.device.random_address),
+            ),
+            check_result=True,
+        )
+
+    async def start(
+        self, duration: float = 0.0, max_advertising_events: int = 0
```
zxzxwu, line 721: nit: Need some docstring to describe units and behaviors.

bumble/device.py
```diff
@@ -3225,6 +3514,73 @@ async def indicate_subscriber(self, connection, attribute, value=None, force=Fal
...
+    @host_event_handler
+    def on_advertising_set_termination(
+        self,
+        status,
+        advertising_handle,
+        connection_handle,
+        number_of_completed_extended_advertising_events,
+    ):
+        if not (
+            advertising_set := self.extended_advertising_sets.get(advertising_handle)
```
zxzxwu, line 3535: The terminated set might be the lagacy one?

bumble/device.py
```diff
@@ -1616,290 +1941,265 @@ def supports_le_phy(self, phy):
...
-        return self.legacy_advertiser or self.extended_advertisers
+        if self.legacy_advertiser:
+            return True
+
+        if self.legacy_advertising_set and self.legacy_advertising_set.enabled:
+            return True
+
+        for advertising_set in self.extended_advertising_sets.values():
+            if advertising_set.enabled:
+                return True
```
zxzxwu, line 2200: nit
```suggestion
        any(advertising_set.enabled for advertising_set in self.extended_advertising_sets.values())
```

bumble/device.py
```diff
@@ -433,37 +443,344 @@ def is_directed(self):
...
+        # Sanity check
+        if (
+            not advertising_parameters.advertising_event_properties.is_legacy
+            and advertising_parameters.advertising_event_properties.is_connectable
+            and advertising_parameters.advertising_event_properties.is_scannable
+        ):
+            raise ValueError(
+                "non-legacy extended advertising event properties may not be both "
+                "connectable and scannable"
+            )
```
zxzxwu, line 624: I prefer to raise warning instead of error here? Also, before issuing a configuration command, we may also raise sanity check warnings.

BTW, the word "sanity" is not considered inclusive in Google. Might be better to call it like compliance?

bumble/host.py
```diff
@@ -38,13 +38,97 @@
...
+    HCI_LE_READ_MAXIMUM_ADVERTISING_DATA_LENGTH_COMMAND,
     HCI_READ_BUFFER_SIZE_COMMAND,
     HCI_READ_LOCAL_VERSION_INFORMATION_COMMAND,
     HCI_RESET_COMMAND,
     HCI_SUCCESS,
     HCI_SUPPORTED_COMMANDS_FLAGS,
     HCI_SYNCHRONOUS_DATA_PACKET,
     HCI_VERSION_BLUETOOTH_CORE_4_0,
+    HCI_INQUIRY_COMPLETE_EVENT,
+    HCI_INQUIRY_RESULT_EVENT,
```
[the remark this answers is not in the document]
barbibulle: You're right, the list is becoming too long. I'll make the change here to just import `hci` and use `hci.XXX` in this file. Eventually, I'll try to push a PR later on to do the same change in other files that also import too many symbols directly. It will make the lines a bit longer, but I think it's a cleaner model.

bumble/device.py
```diff
@@ -436,37 +441,344 @@ def is_directed(self):
...
+        if self.is_legacy:
+            properties |= properties.USE_LEGACY_ADVERTISING_PDUS
+        if self.is_anonymous:
+            properties |= properties.ANONYMOUS_ADVERTISING
+        if self.include_tx_power:
+            properties |= properties.INCLUDE_TX_POWER
+
+        return int(properties)
+
+    @staticmethod
```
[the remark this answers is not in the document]
barbibulle: done

tests/device_test.py
```diff
@@ -301,20 +301,14 @@ async def test_legacy_advertising_disconnection(auto_restart):
...
-            advertising_type=advertiser.advertising_type,
-            own_address_type=advertiser.own_address_type,
-            auto_restart=advertiser.auto_restart,
-            advertising_data=advertiser.advertising_data,
-            scan_response_data=advertiser.scan_response_data,
-        )
+        assert device.is_advertising
     else:
-        device.start_legacy_advertising.assert_not_called()
+        not device.is_advertising
```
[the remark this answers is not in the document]
barbibulle: fixed

examples/run_extended_advertiser_2.py
```diff
@@ -0,0 +1,99 @@
+# Copyright 2021-2022 Google LLC
```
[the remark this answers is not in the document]
barbibulle: fixed

bumble/device.py
barbibulle: done

bumble/device.py
barbibulle: done

bumble/device.py
barbibulle: done

bumble/device.py
barbibulle: fixed

bumble/device.py
barbibulle: done

bumble/device.py
barbibulle: Done

zxzxwu:
LGTM

zxzxwu (approved):
LGTM

PR #383, opened 2023-12-28
Controller: SCO implementation

zxzxwu:
This provides a capability to test SCO virtually. We will also apply the same strategy to CIS.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -32,6 +32,7 @@
         "dhkey",
         "diversifier",
         "endianness",
+        "ESCO",
         "Fitbit",
         "GATTLINK",
         "HANDSFREE",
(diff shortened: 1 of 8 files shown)


zxzxwu:
@barbibulle I suggest we should abandon the support of RemoteLink, so that the type annotation between controller and link would be simpler.

barbibulle:
> @barbibulle I suggest we should abandon the support of RemoteLink, so that the type annotation between controller and link would be simpler.

I agree. RemoteLink served its purpose in the past, but there are easier ways to achieve the same with a slightly different topology: instead of having a local controller that can exchange data with a remote controller via a remote link relay, the alternative is to have a relay that exposes HCI interfaces over a remotable transport (TCP, WebSocket) rather than "link" interfaces. We can now use RootCanal as such a remote HCI hub.

barbibulle (approved):

PR #385, opened 2023-12-28
request MTU change after connection

barbibulle:
Somehow, Android doesn't negotiate a DLE change unless an MTU change is requested...


AlanRosenthal (approved):
very interesting - is this an Android bug or just expected behavior?

barbibulle:
> very interesting - is this an Android bug or just expected behavior?
It's a bug as far as I'm concerned. But since there's no user-accessible API to know or control the Data Length of a connection, it's not technically breaking any contracts with developers, it just makes it tricky to know what Data Length you're going to get (and because this behavior is undocumented, it could also change at any point)

PR #387, opened 2023-12-28
support async read/write for characteristic values

barbibulle:
This allows the `read` and/or `write` functions of `CharacteristicValue` to be async (or be sync functions that return an awaitable), without requiring them to be. The server will automatically adapt to both types.


zxzxwu (approved):
Awesome!

PR #391, opened 2024-01-02
LE read remote features

zxzxwu:
I think we can share BR/EDR and LE behavior in the same function. However, BR/EDR type is a little bit complicated(3+1 pages), so returning them might not be so useful.


zxzxwu:
BTW, why we need this feature is because sending LE data without performing feature/version exchange is very easy to fail (on the first packet).

bumble/host.py
```diff
@@ -487,8 +488,8 @@ def supported_commands(self):
 
         return commands
 
-    def supports_le_feature(self, feature):
-        return (self.local_le_features & (1 << feature)) != 0
+    def supports_le_feature(self, feature: LeFeatureMask) -> bool:
```
barbibulle, line 491: Not necessarily a problem, but this is a different method compared to before: instead of asking for just one feature, this asks for an entire set of features.
We have to options here:
1/ Either we keep the old semantics, and just ask for one feature at a time, in which case we should have an `LeFeature` simple IntEnum (we could also have an `LeFeatureMask` as an IntFlags if we wanted an actual type for it), or
2/ rename this method to `supports_le_features`, taking a whole mask/set, but in that case it should return `self.local_le_features & features == features` (i.e "all features" are supported, not just "one of the features" is supported, which isn't terribly useful because you don't know which one)

bumble/device.py
```diff
@@ -906,6 +906,30 @@ async def get_phy(self):
     async def request_remote_name(self):
         return await self.device.request_remote_name(self)
 
+    async def read_remote_feature(self) -> None:
```
barbibulle, line 909: maybe this Connection method should be called `get_xxx` and delegate to a Device method, like `connection.get_phy` delegating to `device.get_connection_phy`, to follow the same pattern of locating the actual sending of HCI commands in the Device class.
In that case, the method could be `Connection.get_remote_le_features`, delegating to `Device.get_connection_remote_le_features`.

barbibulle, line 909: This method could return the feature mask, just like `get_phy` returns the PHY values ,or `get_rssi` returns the RSSI value.

bumble/host.py
zxzxwu: Agree. But I think it's better to define flags as they are, instead of their log position. So 2/ is the one I preffered.

bumble/device.py
zxzxwu: Sure, it makes sense to separate Classic and LE into two methods.

zxzxwu: I am think whether it's necessary to delegate methods(but I will do if you insist), since for most of the time, callers won't call via device. The benefit of this way I can come up is to make `Connection` class simpler.

barbibulle: The idea is to keep the Connection class very simple, without direct knowledge of HCI commands, mostly as a helper class that's useful for grouping connection-related functions together, and keep the bulk of the actual implementation in the Device class. This pattern also allows some of the device class methods to use other properties of the Device object, without requiring the Connection class to know about the internal of Device (even though we don't really enforce any sort of public/private distinction between methods currently, maybe we will eventually as the implementation becomes more stable).

zxzxwu: Okay, moved implementation to Device.

barbibulle (approved):

PR #396, opened 2024-01-08
Support loading HFP configuration from file (#1)

henry434343:


zxzxwu:
I think this doesn't need to be loaded in device? Passing a parameter to load in `examples/run_hfp_handsfree.py` is enough.

henry434343:
> I think this doesn't need to be loaded in device? Passing a parameter to load in `examples/run_hfp_handsfree.py` is enough.

Since we would like to use the device directly in our application, passing parameters to examples might not fit our requirements. 
Might I know how can we determine which functions can be added to the device? Thanks for the help.

barbibulle:
> > I think this doesn't need to be loaded in device? Passing a parameter to load in `examples/run_hfp_handsfree.py` is enough.
> 
> Since we would like to use the device directly in our application, passing parameters to examples might not fit our requirements. Do you have any suggestion? Thanks. Might I know how can we determine which functions can be added to the device? Thanks for the help.

I agree with @zxzxwu here. This config is an HFP-specific config, which the Device object doesn't use at all. It's only needed to instantiate a `HfProtocol` object (as is done in the HFP example app). If you want to pass a config file to `run_hfp_handsfree.py`, that code should be enhanced to allow parsing a config file (as you have done in this PR). If you're writing a different application, you would parse the config file from there. It may be useful to offer the JSON config file parsing function as a reusable helper, so that different apps could use the same method without having to re-implement that code. That function could therefore be located in `hfp.py`.

henry434343:
> > > I think this doesn't need to be loaded in device? Passing a parameter to load in `examples/run_hfp_handsfree.py` is enough.
> > 
> > 
> > Since we would like to use the device directly in our application, passing parameters to examples might not fit our requirements. Do you have any suggestion? Thanks. Might I know how can we determine which functions can be added to the device? Thanks for the help.
> 
> I agree with @zxzxwu here. This config is an HFP-specific config, which the Device object doesn't use at all. It's only needed to instantiate a `HfProtocol` object (as is done in the HFP example app). If you want to pass a config file to `run_hfp_handsfree.py`, that code should be enhanced to allow parsing a config file (as you have done in this PR). If you're writing a different application, you would parse the config file from there. It may be useful to offer the JSON config file parsing function as a reusable helper, so that different apps could use the same method without having to re-implement that code. That function could therefore be located in `hfp.py`.

Thanks for the clarification. 
Could I keep the HFP configuration to the same device configuration, or do I need to use a separate HFP-specific config file to indicate the HFP feature?
For instance 
```
def from_config_file(config): # in hfp.py
    ...

configuration = hfp.from_config_file(sys.argv[1]) # in our application
```

zxzxwu:
A question here is that, since HFP is not loaded by default, your application should have an entry for HFP, right?
If there is a customized entry point, could you modify there to take parameters?

henry434343:
> A question here is that, since HFP is not loaded by default, your application should have an entry for HFP, right? If there is a customized entry point, could you modify there to take parameters?

Thanks for the clarification, will pass parameter from our entry point.

PR #398, opened 2024-01-10
update credits even without a sink

barbibulle:
Even if the RFComm session has no sink (not very useful, but could happen in a test), we still need to update the credits.

### DIFF
--- a/bumble/rfcomm.py
+++ b/bumble/rfcomm.py
@@ -538,7 +538,7 @@ class DLC(EventEmitter):
             f'[{self.dlci}] {len(data)} bytes, '
             f'rx_credits={self.rx_credits}: {data.hex()}'
         )
-        if len(data):
+        if data:
             if self.sink:
                 self.sink(data)  # pylint: disable=not-callable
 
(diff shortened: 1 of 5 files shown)


bumble/rfcomm.py
```diff
@@ -538,8 +538,9 @@ def on_uih_frame(self, frame: RFCOMM_Frame) -> None:
             f'[{self.dlci}] {len(data)} bytes, '
             f'rx_credits={self.rx_credits}: {data.hex()}'
         )
-        if len(data) and self.sink:
-            self.sink(data)  # pylint: disable=not-callable
+        if data:
+            if self.sink:
+                self.sink(data)  # pylint: disable=not-callable
```
zxzxwu, line 543: Since we have annotation now, it is no longer needed.
```suggestion
                self.sink(data)
```

zxzxwu (approved):

bumble/rfcomm.py
barbibulle: I still get a pylint error without that comment:
`bumble/rfcomm.py:543:16: E1102: self.sink is not callable (not-callable)`

PR #415, opened 2024-01-30
HFP: State memory and event emission

zxzxwu:
* Store active codec in protocol object.
* Emit unsolicited AG responses as events.
* Implement enhanced call status (CLCC).
* Also rewrite docstrings.

### DIFF
--- a/bumble/hfp.py
+++ b/bumble/hfp.py
@@ -21,12 +21,11 @@ import asyncio
 import dataclasses
 import enum
 import traceback
-import warnings
-from typing import Dict, List, Union, Set, Any, TYPE_CHECKING
+import pyee
+from typing import Dict, List, Union, Set, Any, Optional, TYPE_CHECKING
 
-from . import at
-from . import rfcomm
-
+from bumble import at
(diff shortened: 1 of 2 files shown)


bumble/hfp.py
```diff
@@ -361,9 +444,12 @@ class HfIndicatorState:
     enabled: bool = False
 
 
-class HfProtocol:
-    """Implementation for the Hands-Free side of the Hands-Free profile.
-    Reference specification Hands-Free Profile v1.8"""
+class HfProtocol(pyee.EventEmitter):
```
hchataing, line 447: It might be useful to document in the class docstring the list of events emitted from HfProtocol

hchataing:
Other than this doc request, looks good !

bumble/hfp.py
zxzxwu: Done. That's also what I want to do in static check, but it will become a complicated mechanism.

hchataing (approved):

barbibulle (approved):

PR #419, opened 2024-02-01
Add local LMP feature reader

zxzxwu:


tests/host_test.py
```diff
@@ -0,0 +1,62 @@
+# Copyright 2021-2022 Google LLC
```
barbibulle, line 1: nit: 2024

barbibulle (approved):

tests/host_test.py
zxzxwu: Done

PR #420, opened 2024-02-02
Add RFCOMM and SDP context manager and search helper

zxzxwu:
* Add an RFCOMM (Client, Server) and SDP(Client only because Server is managed by device) context manager.
* Add an RFCOMM helper function to search channel from SDP.
* And some test cases.

### DIFF
--- a/bumble/rfcomm.py
+++ b/bumble/rfcomm.py
@@ -22,10 +22,13 @@ import asyncio
 import dataclasses
 import enum
 from typing import Callable, Dict, List, Optional, Tuple, Union, TYPE_CHECKING
+from typing_extensions import Self
 
 from pyee import EventEmitter
 
-from . import core, l2cap
+from bumble import core
+from bumble import l2cap
+from bumble import sdp
 from .colors import color
 from .core import (
     UUID,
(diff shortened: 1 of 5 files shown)


bumble/rfcomm.py
```diff
@@ -154,15 +152,54 @@ def make_service_sdp_records(
...
+                sdp.SDP_SERVICE_CLASS_ID_LIST_ATTRIBUTE_ID,
+                sdp.DataElement.sequence([sdp.DataElement.uuid(uuid)]),
             )
         )
 
     return records
 
 
+# -----------------------------------------------------------------------------
+async def search_channels_from_sdp(connection: Connection) -> Dict[int, List[UUID]]:
```
barbibulle, line 165: suggest: maybe this could be named `find_rfcomm_channels ` (see other comment for alignement with similar functionality currently in `bench.py`)

barbibulle, line 165: There's a similar helper in `bench.py`. The other one doesn't return a list, but looks just for one specific UUID. I think both are useful. 
Maybe the single-uuid version from `bench.py` could be re-written by using this one, and moved to `rfcomm.py`.
We'd end up with:

- `rfcomm.find_rfcomm_channels` (this one, just different name) and 
- `rfcomm.find_rfcomm_channel_with_uuid` (the refactored one moved over from `bench.py`)

zxzxwu: Done

zxzxwu: I was wondering whether it should be a directional search or general search, but it seems not bad to have both.

bumble/rfcomm.py
```diff
@@ -154,15 +152,81 @@ def make_service_sdp_records(
...
+                    ]
+            if not service_classes or not channel:
+                logger.warning(f"Bad result {attribute_lists}.")
+            else:
+                results[channel] = service_classes
+    return results
+
+
+# -----------------------------------------------------------------------------
+async def find_rfcomm_channel_with_uuid(
```
barbibulle, line 204: That function could be implemented by calling the generic find functions, then looking for the UUID in the resulting dict (less duplicated code)

zxzxwu: Done

barbibulle (approved):

PR #424, opened 2024-02-04
Fix double-disable legacy advertising set

zxzxwu:
When legacy advertising set is disabled passively(by set termination), the legacy advertising set won't be released, and the next `stop_advertising()` call will try to disable it again and cause an error.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -2065,7 +2065,9 @@ class Device(CompositeEventEmitter):
         """Stop legacy advertising."""
         # Disable advertising
         if self.legacy_advertising_set:
-            await self.legacy_advertising_set.stop()
+            if self.legacy_advertising_set.enabled:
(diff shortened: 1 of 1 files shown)


zxzxwu:
The error is due to https://github.com/google/bumble/pull/423.

bumble/device.py
```diff
@@ -3541,11 +3543,9 @@ def on_advertising_set_termination(
         connection_handle,
         number_of_completed_extended_advertising_events,
     ):
+        # Legacy advertising set is also one of extended advertising sets.
```
barbibulle, line 3546: Yes, indeed! I think we could also remove the test on line 2212 for the same reason.

barbibulle (approved):

bumble/device.py
zxzxwu: Done

PR #425, opened 2024-02-04
Refactor command supporting list

zxzxwu:


bumble/host.py
```diff
@@ -583,31 +595,18 @@ def send_l2cap_pdu(self, connection_handle: int, cid: int, pdu: bytes) -> None:
...
-                if value == command:
-                    # Check if the flag is set
-                    if octet < len(self.local_supported_commands) and flag_position < 8:
-                        return (
-                            self.local_supported_commands[octet] & (1 << flag_position)
-                        ) != 0
-
-        return False
+    def supports_command(self, op_code: int) -> bool:
+        return (
```
barbibulle, line 599: I recommend using `HCI_SUPPORTED_COMMANDS_MASKS.get()` instead of `HCI_SUPPORTED_COMMANDS_MASKS[]` to avoid an exception if the `op_code` isn't in the map (ex: a caller asking about an unknown/new/vendor op_code).

barbibulle (approved):

bumble/host.py
zxzxwu: Done

PR #434, opened 2024-02-28
show timestamps from snoop logs

barbibulle:
Show UTC timestamps when printing logs from a snoop log.

### DIFF
--- a/apps/show.py
+++ b/apps/show.py
@@ -130,7 +130,7 @@ class Printer:
 
     def print(self, message: str) -> None:
         self.index += 1
-        print(f"[{self.index}]{message}")
+        print(f"[{self.index:8}]{message}")
 
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


apps/show.py
```diff
@@ -89,7 +120,17 @@ def next_packet(self):
...
+
+
+# -----------------------------------------------------------------------------
+class Printer:
+    def __init__(self):
+        self.index = 0
+
+    def print(self, message: str) -> None:
+        self.index += 1
+        print(f"[{self.index}]{message}")
```
zxzxwu, line 133: nit: Could align here to make it prettier 😃 
```suggestion
        print(f"[{self.index:8}]{message}")
```

zxzxwu (approved):

apps/show.py
barbibulle: done

PR #435, opened 2024-03-03
open_tcp_server_transport: allow explicit sock as input.

jeru:
When a user doesn't need an exact port, but cares more about getting SOME unused port, they can do:
* Create a socket outside with port=None or port=0.
* Use socket.getsockname()[1] to get the allocated port and pass to the TCP client somehow.
* Use the created socket to create a TCP server transport.

Use-case: unit-testing embedded software that implements a BLE host. The controller will be a Bumble controller, connected to the host via a TCP channel.
* The host will have a TCP-client HCI transport for testing.
* The pytest setup code will allocate the TCP server and pass the port number to the host.

Also add some unittests with python mock.


jeru:
Thanks for the suggestion. Indeed that makes the interface cleaner. Refactored accordingly.

And sorry for the noob mistakes on the lint and the extra trouble on re-approving the workflows... Was using an outdated "black" which touched too many files, so didn't even notice "invoke project.pre-commit" has multiple stages.

tests/transport_tcp_server_test.py
```diff
@@ -0,0 +1,64 @@
...
+Not hermetic. Should only run manually with
+  $ PYTEST_NOSKIP=1 pytest tests
+''',
+)
+def test_open_with_real_socket():
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+        sock.bind(('localhost', 0))
+        port = sock.getsockname()[1]
+        assert port != 0
+        asyncio.run(open_tcp_server_transport(sock=sock))
```
barbibulle, line 64: this should be `open_tcp_server_transport_with_socket`

jeru: Thanks. Now manually tested as well.

jeru:
Any further comments?

barbibulle:
> Any further comments?

Sorry about that, I forgot the approval was pending. Good to go!

barbibulle (approved):

jeru:
Thanks!

PR #437, opened 2024-03-04
pandora_server: Parse FLAGS into advertising data

SilverBzH:
Bug: 328089785

### DIFF
--- a/bumble/pandora/host.py
+++ b/bumble/pandora/host.py
@@ -50,6 +50,8 @@ from pandora.host_grpc_aio import HostServicer
 from pandora.host_pb2 import (
     NOT_CONNECTABLE,
     NOT_DISCOVERABLE,
+    DISCOVERABLE_LIMITED,
+    DISCOVERABLE_GENERAL,
     PRIMARY_1M,
     PRIMARY_CODED,
     SECONDARY_1M,
@@ -65,6 +67,7 @@ from pandora.host_pb2 import (
     ConnectResponse,
     DataTypes,
     DisconnectRequest,
+    DiscoverabilityMode,
     InquiryResponse,
     PrimaryPhy,
     ReadLocalAddressResponse,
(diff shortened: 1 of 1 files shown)


SilverBzH:
Test with a BumbleBluetoothTests:
Screenshot 2024-03-05 at 10 18 10 AM

bumble/pandora/host.py
```diff
@@ -735,6 +738,18 @@ def uuid16_from_str(uuid: str) -> bytes:
                 )
             )
 
+        flag_map = {
+            NOT_DISCOVERABLE: AdvertisingData.BR_EDR_NOT_SUPPORTED_FLAG,
```
DeltaEvo, line 742: This seems wrong, the Flags data type is supposed to be a bit set, so when the `le_discoverability_mode` is NOT_DISCOVERABLE, I would expect to have a flag set to `0`

bumble/pandora/host.py
```diff
@@ -735,6 +738,18 @@ def uuid16_from_str(uuid: str) -> bytes:
...
+            DISCOVERABLE_LIMITED: AdvertisingData.LE_LIMITED_DISCOVERABLE_MODE_FLAG,
+            DISCOVERABLE_GENERAL: AdvertisingData.LE_GENERAL_DISCOVERABLE_MODE_FLAG,
+        }
+
+        if dt.le_discoverability_mode:
```
DeltaEvo, line 747: This one is a bit special, should we always set it?

I don't remember how protobuf works but I think here you are exclusing the `NOT_DISCOVERABLE` value

bumble/pandora/host.py
```diff
@@ -735,6 +738,18 @@ def uuid16_from_str(uuid: str) -> bytes:
...
+            flags = flag_map.get(
+                dt.le_discoverability_mode, AdvertisingData.BR_EDR_NOT_SUPPORTED_FLAG
+            )
```
DeltaEvo, line 750:
```suggestion
            flags = flag_map[dt.le_discoverability_mode]
```

bumble/pandora/host.py
DeltaEvo: Oh, per the spec it seems that we can exclude the structure when all flags are 0

```
The Flags data type shall be included when any of the Flag bits are non-zero and the advertising packet is connectable, otherwise the Flags data type may be omitted.
```

DeltaEvo (changes requested):

bumble/pandora/host.py
SilverBzH: Just updated it

DeltaEvo (approved):

PR #439, opened 2024-03-07
Media Control Service Client implementation

zxzxwu:
We may also drop new Characteristic adapters, because MCS characteristics seldom get read.

### DIFF
--- a/bumble/profiles/mcp.py
+++ b/bumble/profiles/mcp.py
@@ -1,0 +1,448 @@
+# Copyright 2021-2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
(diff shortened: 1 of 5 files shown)


bumble/profiles/mcp.py
```diff
@@ -0,0 +1,450 @@
...
+# -----------------------------------------------------------------------------
+# Classes
+# -----------------------------------------------------------------------------
+
+
+class ObjectId(int):
+    '''See Media Control Service 4.4.2. Object ID field.'''
+
+    @classmethod
+    def parse_from(cls: Type[Self], data: bytes) -> Self:
```
barbibulle, line 170: for factory methods that take bytes as input, the pattern in the rest of the code is that the method should be named just `from_bytes`.
A few classes in the code base have a `parse_from_bytes`, which is used by the automated parsing code, takes an offset and bytes and returns not just an instance, but a (offset, instance) tuple.
I see that there's one exception, where `hfp.py` use `parse_from` as a factory method, from bytes, in a couple places, that's something we can fix later.

barbibulle (approved):

bumble/profiles/mcp.py
```diff
@@ -0,0 +1,450 @@
...
+
+from typing import Type, Optional, ClassVar, Dict, TYPE_CHECKING
+from typing_extensions import Self
+
+# -----------------------------------------------------------------------------
+# Constants
+# -----------------------------------------------------------------------------
+
+
+class PlayingOrder(enum.IntEnum):
```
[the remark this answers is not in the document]
zxzxwu: Done

bumble/profiles/mcp.py
zxzxwu: Done

PR #454, opened 2024-03-25
GATT server support for Rust

marshallpierce:
With a whole lot of support to get there:

- Device::add_service(), with some complexity around how to represent the mutable Python `Service` being used under the covers
- AttributePermission, CharacteristicProperty, and other supporting GATT-related stuff
- Ability to handle the different data types used by various Python `Characteristic` subclasses
- Support for running tests in the context of a background `rootcanal` process
- `TryToPy` and `TryFromPy` traits to help make the lifecycle more clear for structs that have a Python equivalent, but are also meaningful standalone
- Add example with a minimal battery service
- Rearrange known service IDs to make it possible to refer to a service UUID in a meaningful way
- Make Address hold its own state to avoid propagating PyResult so much, and adopt the new TryToPy and TryFromPy traits
- `AdvertisingDataValue` trait so users don't have to remember as many ADU encoding details at a call site


bumble/device.py
```diff
@@ -3533,7 +3533,11 @@ def add_default_services(self, generic_access_service=True):
     async def notify_subscriber(self, connection, attribute, value=None, force=False):
         await self.gatt_server.notify_subscriber(connection, attribute, value, force)
 
-    async def notify_subscribers(self, attribute, value=None, force=False):
+    async def notify_subscribers(self,
+        attribute: Attribute,
+        value: Optional[bytes] = None,
+        force: bool = False,
+    ) -> None:
```
marshallpierce, line 3541: Added type annotations based on poking around at the callers, but since of course the type info isn't pervasive I'm not sure these are right

rust/Cargo.toml
```diff
@@ -17,51 +17,57 @@ rust-version = "1.76.0"
...
 futures = "0.3.28"
+async-trait = "0.1.77"
+uuid = "1.8.0"
 
 # Dev tools
 file-header = { version = "0.1.2", optional = true }
 globset = { version = "0.4.13", optional = true }
 
 # CLI
-anyhow = { version = "1.0.71", optional = true }
```
marshallpierce, line 43: No longer optional due to wanting a "don't care" error for callbacks, etc, that wasn't the Python-specific `PyError`

rust/examples/battery_service.rs
```diff
@@ -0,0 +1,187 @@
...
+        CharacteristicProperty::Read | CharacteristicProperty::Notify,
+        AttributePermission::Readable.into(),
+        CharacteristicValueHandler::new(Box::new(BatteryRead), Box::new(NoOpWrite)),
+    );
+    let service = Service::new(uuid.into(), vec![battery_level]);
+    let service_handle = device.add_service(&service)?;
+
+    let mut builder = AdvertisementDataBuilder::new();
+    builder.append(CommonDataType::CompleteLocalName, "Bumble Battery")?;
+    builder.append(
```
marshallpierce, line 131: This is fairly manual, but given the use cases for Bumble that might want to do somewhat exciting things vis-a-vis spec compliance, I'm not sure it makes sense to make a more OS-style API that doesn't require you to assemble the advertisement and prevents you from doing things wrong...?

rust/pytests/rootcanal.rs
```diff
@@ -0,0 +1,295 @@
...
+///
+/// Sets up `env_logger`, if it wasn't already configured, so that debug logging with rootcanal
+/// output can be enabled with `RUST_LOG=debug` env var.
+///
+/// If the correct rootcanal binary isn't already cached, it will be downloaded from GitHub
+/// Releases if a binary is available for the current OS & architecture.
+///
+/// This is a stop-gap until rootcanal's build is improved to the point that a rootcanal-sys crate,
+/// and using rootcanal as a library rather than a separate process, becomes feasible.
+pub(crate) async fn run_with_rootcanal<O, F>(closure: F) -> anyhow::Result<()>
```
marshallpierce, line 41: @hchataing forgive me for these rootcanal crimes. Let me know if there is a less bad way to do this while I try to get the rootcanal build working on more systems.

rust/pytests/rootcanal.rs
```diff
@@ -0,0 +1,295 @@
...
+
+        Ok(rootcanal_bin)
+    })
+    .await
+}
+
+/// Execute `closure` with a simple directory lock held.
+///
+/// Assumes that directory creation is atomic, which it is on most OS's.
+async fn with_dir_lock<T>(
```
marshallpierce, line 269: Needed since tests run concurrently, and I don't want to sacrifice that as integration-level ones are rather slow

rust/src/wrapper/att.rs
```diff
@@ -0,0 +1,62 @@
...
+pub use crate::internal::att::{
+    AttributePermission, AttributePermissions, AttributeRead, AttributeUuid, AttributeWrite,
+    MutexRead, NoOpWrite,
+};
+use crate::wrapper::core::TryToPy;
+use pyo3::prelude::PyModule;
+use pyo3::types::PyBytes;
+use pyo3::{intern, PyAny, PyResult, Python};
+
+impl TryToPy for AttributeUuid {
```
marshallpierce, line 26: Not quite sure how it'll work out, but I'm experimenting with having `Try[To|From]Py` in the `wrapper` module distinct from the relatively pure Rust structs in `crate::internal`

rust/src/wrapper/device/mod.rs
```diff
@@ -112,51 +116,73 @@ impl Device {
...
             kwargs.set_opt_item("generic_access_service", generic_access_service)?;
 
-            PyModule::import(py, intern!(py, "bumble.device"))?
-                .getattr(intern!(py, "Device"))?
+            let device_ctor = PyModule::import(py, intern!(py, "bumble.device"))?
+                .getattr(intern!(py, "Device"))?;
+
+            // Needed for Python 3.8-3.9, in which the Semaphore object, when constructed, calls
+            // `get_event_loop`.
+            wrap_python_async(py, device_ctor)?
```
marshallpierce, line 129: Had to sprinkle this on Device and Peer invocations that end up instantiating an `asyncio.Semaphore`

rust/pytests/rootcanal.rs
```diff
@@ -0,0 +1,295 @@
...
+                        None => break,
+                        Some(line) => debug!("{line}"),
+                    }
+                }
+            }
+        }
+    })
+}
+
+/// Returns the path to a rootcanal binary, downloading it from GitHub if necessary.
```
nabilwadih, line 217: It seems a little odd to me that this is downloading artifacts at runtime, would it make sense to move this step into a build.rs and setting this up at build time?

rust/src/internal/core.rs
```diff
@@ -0,0 +1,195 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::internal::adv::AdvertisementDataValue;
+use lazy_static::lazy_static;
+use nom::combinator;
+use std::fmt;
+
+lazy_static! {
+    static ref BASE_UUID: [u8; 16] = hex::decode("0000000000001000800000805F9B34FB")
```
nabilwadih, line 21: where does this come from?

rust/src/wrapper/assigned_numbers/services.rs
```diff
@@ -14,69 +14,298 @@
...
+        (0x184C, GENERIC_TELEPHONE_BEARER),
+        (0x184D, MICROPHONE_CONTROL),
     ]
     .into_iter()
     .map(|(num, name)| (Uuid16::from_le_bytes(num.to_le_bytes()), name))
     .collect();
 }
+
+pub const GENERIC_ACCESS: NamedUuid = NamedUuid::new(
+    Uuid16::from_be_bytes(0x1800_u16.to_be_bytes()),
```
nabilwadih, line 87: nit: would it make sense to pull these values out of the hashmap, so there is one source of truth?

nabilwadih (approved):

rust/src/internal/core.rs
marshallpierce: It's the bluetooth base UUID: https://stackoverflow.com/questions/36212020/how-can-i-convert-a-bluetooth-16-bit-service-uuid-into-a-128-bit-uuid

rust/src/wrapper/assigned_numbers/services.rs
marshallpierce: Good point; that's an awkward vestige from the former structure. I think it will be cleaner to construct the hashmap from the NamedUuid instances, but either way one source of truth.

rust/examples/battery_service.rs
```diff
@@ -0,0 +1,188 @@
+// Copyright 2023 Google LLC
```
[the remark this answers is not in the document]
marshallpierce: Fixed 'em all.

rust/pytests/rootcanal.rs
marshallpierce: I don't enjoy the current approach, but I don't see a great way of doing it with `build.rs` either that would work nicely with doing dirty tracking. I hope either way it's short lived, though -- with `n` additional hours I think I can get `rootcanal` building as a library, at which point a `rootcanal-sys` crate becomes reasonable, and this whole spawn-a-process thing can go away.

bumble/device.py
barbibulle: This is good. We want to eventually have everything type, incrementally.

rust/examples/battery_service.rs
barbibulle: Having a set of reusable AdvertismementData pre-set builders would be nice, for users who what just basic common stuff. That's something I've considered adding to the Python lib but haven't found the time to yet.

whitevegagabriel (approved):

whitevegagabriel:
Also, although we've not quite kept up-to-date with the Changelog up to this point, could we at least make note of these changes in the Changelog now, so they're there for the next release?

rust/examples/battery_service.rs
```diff
@@ -0,0 +1,188 @@
...
+    )?;
+    device.set_advertising_data(builder)?;
+    device.power_on().await?;
+    device.start_advertising(true).await?;
+
+    let char_handle = service_handle
+        .characteristic_handle(battery_level_uuid)
+        .expect("Battery level should be present");
+
+    loop {
```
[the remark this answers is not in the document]
marshallpierce: It could be done with a cancellation token and a `select!` in the loop, but would it really change anything? Either way they both just die when ^C'd. There's no meaningful teardown to be done.

rust/src/internal/core.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    /// The UUID in little-endian bytes form
+    pub fn as_le_bytes(&self) -> [u8; 2] {
+        let mut uuid = self.be_bytes;
+        uuid.reverse();
+        uuid
+    }
+
+    pub(crate) fn parse_le(input: &[u8]) -> nom::IResult<&[u8], Self> {
+        combinator::map_res(nom::bytes::complete::take(2_usize), |b: &[u8]| {
```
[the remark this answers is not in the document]
marshallpierce: These are intentional to show a hint of where the functions come from -- though the `nom` could certainly be lopped off, I think.

rust/src/wrapper/hci.rs
```diff
@@ -73,100 +73,159 @@ impl IntoPy<PyObject> for HciCommand {
...
     }
 
     /// True if the address is resolvable
-    pub fn is_resolvable(&self) -> PyResult<bool> {
-        Python::with_gil(|py| {
-            self.0
-                .getattr(py, intern!(py, "is_resolvable"))?
-                .extract::<bool>(py)
-        })
+    pub fn is_resolvable(&self) -> bool {
```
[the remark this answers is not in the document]
marshallpierce: Because I am a doofus and copied the impl of `is_resolved()`, not `is_resolvable()`. Good eye!

rust/src/wrapper/hci.rs
```diff
@@ -73,100 +73,159 @@ impl IntoPy<PyObject> for HciCommand {
 }
 
 /// A Bluetooth address
-#[derive(Clone)]
-pub struct Address(pub(crate) PyObject);
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
```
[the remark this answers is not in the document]
marshallpierce: Yep!

rust/src/wrapper/mod.rs
```diff
@@ -36,13 +36,14 @@ pub mod controller;
 pub mod core;
 pub mod device;
 pub mod drivers;
-pub mod gatt_client;
+
```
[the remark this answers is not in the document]
marshallpierce: nope!

rust/pytests/rootcanal.rs
```diff
@@ -0,0 +1,295 @@
...
+    let version = "1.10.0";
+    // from https://github.com/google/rootcanal/releases
+    let (url, zip_bin_path) = match (env::consts::OS, env::consts::ARCH) {
+        ("linux", "x86_64") => ("https://github.com/google/rootcanal/releases/download/v1.10.0/rootcanal-1.10.0-linux-x86_64.zip", "rootcanal-linux-x86_64/bin/rootcanal"),
+        ("macos", "x86_64")  => ("https://github.com/google/rootcanal/releases/download/v1.10.0/rootcanal-1.10.0-macos-x86_64.zip", "rootcanal-macos-x86_64/bin/rootcanal"),
+        _ => {
+            return Err(anyhow!("No Rootcanal binary available for {} {}, sorry", env::consts::OS, env::consts::ARCH));
+        }
+    };
+    let rootcanal_dir = directories::ProjectDirs::from("com", "google", "bumble")
```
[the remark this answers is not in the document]
marshallpierce: Done

rust/pytests/rootcanal.rs
```diff
@@ -0,0 +1,295 @@
...
+async fn with_dir_lock<T>(
+    dir: &path::Path,
+    closure: impl Future<Output = anyhow::Result<T>>,
+) -> anyhow::Result<T> {
+    // wait until we can create the dir
+    loop {
+        match fs::create_dir(dir) {
+            Ok(_) => break,
+            Err(e) => {
+                if e.kind() == io::ErrorKind::AlreadyExists {
```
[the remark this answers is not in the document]
marshallpierce: Done

rust/examples/battery_service.rs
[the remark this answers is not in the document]
whitevegagabriel: In the past (not sure if the behavior is still the same), if you force-quit without calling "stop advertising", the dongle will continue advertising.

That said, in the past, there were issues with the `asyncio` library inadvertently setting up a `KeyboardInterrupt` signal handler when it imported the `signal` library, so even if you gracefully await ctrl-c, it exits before "stop advertising" gets called.

Though at the end of the day, this is just an example. So if you want to keep it simple, I can vibe with that.

rust/src/internal/core.rs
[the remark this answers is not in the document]
whitevegagabriel: I personally would find this easier to read without the long namespaces. But I've also done far less parsing than you have, so knowing when its important to include these namespaces goes a little over my head (aside from when needing to disambiguate, like if you were using `streaming::take` and `complete::take` in this same module).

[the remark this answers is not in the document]
marshallpierce: Yep, that's pretty much it: if you just `take`, are you taking bytes? bits? 

In general I've found it helpful for reader context to leave a bit of the module path present for things outside the current crate, or in unusual circumstances, the current module.

rust/examples/battery_service.rs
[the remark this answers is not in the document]
marshallpierce: Hm, I suppose it would be better for readers of the example to be able to see best practices in all of them. I'll add in ctrl-c handling.

rust/src/internal/core.rs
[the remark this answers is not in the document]
marshallpierce: Oh, now I remember... I can't chop off `nom` because we already have the `bytes` crate in scope, and introducing different names at use-time seems even more confusing.

PR #467, opened 2024-04-11
Make A2DP constant enums

zxzxwu:


web/speaker/speaker.py
barbibulle, line 1: This change doesn't work. The speaker.py app that's under apps/speaker is similar but different from the one in web/speaker. This one, under web/speaker, is a limited subset, which is designed to run entirely in a browser page (so, no use of stdio, command line, subprocesses, etc.). The one under apps/speaker is a command line tool that can take advantage of the local filesystem, subprocesses, etc.
It should be possible to refactor some of the functionality between the two into a shared library, maybe in a separate PR. But the web/speaker/speaker.py app must remain purely a browser-only app.

bumble/a2dp.py
```diff
@@ -55,19 +56,12 @@
...
-A2DP_NON_A2DP_CODEC_TYPE       = 0xFF
-
-A2DP_CODEC_TYPE_NAMES = {
-    A2DP_SBC_CODEC_TYPE:            'A2DP_SBC_CODEC_TYPE',
-    A2DP_MPEG_1_2_AUDIO_CODEC_TYPE: 'A2DP_MPEG_1_2_AUDIO_CODEC_TYPE',
-    A2DP_MPEG_2_4_AAC_CODEC_TYPE:   'A2DP_MPEG_2_4_AAC_CODEC_TYPE',
-    A2DP_ATRAC_FAMILY_CODEC_TYPE:   'A2DP_ATRAC_FAMILY_CODEC_TYPE',
-    A2DP_NON_A2DP_CODEC_TYPE:       'A2DP_NON_A2DP_CODEC_TYPE'
-}
+class CodecType(enum.IntEnum):
```
barbibulle, line 59: This enum is open-ended (new codec values can be defined over time), so it should subclass `utils.OpenIntEnum`.
The same applies to a few other enums in this PR

PR #468, opened 2024-04-12
pyusb: powercycle the dongle

SilverBzH:


SilverBzH:
This PR allows power cycling of PyUSB devices when connected to a hub with port-level power control. Power cycling will be skipped if a compatible hub is not found.

barbibulle:
> This PR allows power cycling of PyUSB devices when connected to a hub with port-level power control. Power cycling will be skipped if a compatible hub is not found.

the `pyusb` transport is mostly a legacy implementation. The current/maintained USB transport is is `transport/usb.py`, implementation based on `libusb1` instead of `pyusb`. Would it make sense to implement this behavior in `usb.py` instead (or in addition)?

bumble/transport/pyusb.py
```diff
@@ -245,6 +258,13 @@ def device_path(device):
         raise ValueError('device not found')
     logger.debug(f'USB Device: {device}')
 
+    # Power Cycle the device
+    try:
+        device = _power_cycle(device)  # type: ignore
```
barbibulle, line 263: Power-cycling should be optional (controlled with a transport property for example) rather than the default. For USB dongles where firmware is loaded dynamically, it's disruptive to have to re-load the firmware each time you open the device.
Ideally, USB devices are well-behaved, and an HCI RESET should put them back in a clean state.

bumble/transport/pyusb.py
```diff
@@ -308,3 +328,73 @@ def device_path(device):
...
+    This involves temporarily disabling its port on the hub and then re-enabling it.
+    """
+    device_path = f'{device.bus}-{".".join(map(str, device.port_numbers))}'  # type: ignore
+    hub = _find_hub_by_device_path(device_path)
+
+    if hub:
+        try:
+            device_port = device.port_numbers[-1]  # type: ignore
+            _set_port_status(hub, device_port, False)
+            time.sleep(POWER_CYCLE_DELAY)
```
barbibulle, line 345: Blocking sleeps aren't good for async code. This function should become an async function, and use `asyncio.sleep()` instead.

bumble/transport/pyusb.py
SilverBzH: I agree that it should be optional, and adding a specific spec to the transport would make more sense, any specific character you would like to add to the transport specification ?

The primary purpose of this PR is to address the issue that we couldn't return the device to a clean state using an HCI_RESET when it was crashing. We had to resort to manually unplugging and reconnecting the device.

However, oddly enough, the firmware did not reload into the dongles after performing a power cycle through this interface.

pyusb_power_cycle

SilverBzH:
> the `pyusb` transport is mostly a legacy implementation. The current/maintained USB transport is is `transport/usb.py`, implementation based on `libusb1` instead of `pyusb`. Would it make sense to implement this behavior in `usb.py` instead (or in addition)?

We switched to using the `pyusb` implementation because we encountered issues with the `usb` one during our testing. We didn't investigate the `usb` implementation thoroughly before moving to `pyusb`, as it is performing well in our test environment.

bumble/transport/pyusb.py
SilverBzH: Agree, done.

bumble/transport/pyusb.py
barbibulle: A simple way to explicitly ask that the transport try a power cycle at open could be: prefix the name with `!`.
So: `pyusb:0` --> no power-cycle, `pyusb:!0` --> power-cycle.

SilverBzH: I've added this logic in the last commit. Let me know if it is okay with you

barbibulle (approved):

PR #477, opened 2024-04-29
Fix HFP query call status

zxzxwu:
* Fix misplaced arguments
* Fix tests


henry434343 (approved):
Thanks for the quick response and fix.

barbibulle (approved):

PR #478, opened 2024-04-29
Make DeviceConfiguration dataclass

zxzxwu:


bumble/device.py
```diff
@@ -1252,75 +1254,47 @@ def __str__(self):
...
     def load_from_dict(self, config: Dict[str, Any]) -> None:
+        config = copy.deepcopy(config)
+
         # Load simple properties
-        self.name = config.get('name', self.name)
-        if address := config.get('address', None):
+        if address := config.pop('address', None):
             self.address = Address(address)
-        self.class_of_device = config.get('class_of_device', self.class_of_device)
-        self.advertising_interval_min = config.get(
```
barbibulle, line 1291: Some configs use just `advertising_interval` (without `_min` or `_max`) to mean `advertisting_interval_min=advertising_interval_max=advertising_interval`). Maybe we could keep backward compatibility with those configs: if `advertising_interval` is set, automatically use it to set `_min` and `_max`?

zxzxwu: Oh, I didn't notice that. Added the original logic back.

barbibulle (approved):

PR #480, opened 2024-05-03
HFP: Add AG example and fix errors

zxzxwu:
* Fix CLIP handler error
* Fix CMER handler error
* Implement AG +VGS +VGM +CCWA handler
* Fix wrong AG UUID (the original one is for HSP)
* Fix find_hf_sdp_record() may find AG record
* Implement an example UI for AG

### DIFF
--- a/bumble/hfp.py
+++ b/bumble/hfp.py
@@ -50,7 +50,7 @@ from bumble.core import (
     ProtocolError,
     BT_GENERIC_AUDIO_SERVICE,
     BT_HANDSFREE_SERVICE,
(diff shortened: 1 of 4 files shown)


barbibulle (approved):

zxzxwu:
I just found my controller can support SCO over HCI very well - based on this, I will add a browser play example here.

zxzxwu:
A potential issue: When remote send RFCOMM PDU right after a DLC is established, `open_dlc()` might still be in progress, so there is no sink handler set to receive HFP protocol SDU at the moment. Will fix this in another PR.

PR #481, opened 2024-05-04
allow checking results for HCI_Command_Status_Event

barbibulle:
Support `check_status == True` when calling `send_command` for commands that return an `HCI_Command_Status_Event`. (not sure why this has not come up before...)


zxzxwu (approved):

zxzxwu:
I think we can also enable some checks in device.py to verify it? Like create connection commands.

PR #483, opened 2024-05-07
RFCOMM: Handle packets received before DLC sink set

zxzxwu:
This race condition is not common since usually the initiator (no matter it's an L2CAP or RFCOMM channel) should send the first packet. However, in HFP, if it is AG opening the channel, AG cannot send the first packet, and HF may send AT+BRSF too quick that `open_dlc()` haven't returned yet.

### DIFF
--- a/bumble/rfcomm.py
+++ b/bumble/rfcomm.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 
 import logging
 import asyncio
+import collections
 import dataclasses
 import enum
 from typing import Callable, Dict, List, Optional, Tuple, Union, TYPE_CHECKING
@@ -54,6 +55,7 @@ logger = logging.getLogger(__name__)
 # fmt: off
 
 RFCOMM_PSM = 0x0003
+DEFAULT_RX_QUEUE_SIZE = 32
 
 class FrameType(enum.IntEnum):
     SABM = 0x2F  # Control field [1,1,1,1,_,1,0,0] LSB-first
@@ -445,7 +447,8 @@ class DLC(EventEmitter):
         RESET = 0x05
 
     connection_result: Optional[asyncio.Future]
-    sink: Optional[Callable[[bytes], None]]
+    _sink: Optional[Callable[[bytes], None]]
+    _enqueued_rx_packets: collections.deque[bytes]
 
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


bumble/rfcomm.py
```diff
@@ -549,8 +554,7 @@ def on_uih_frame(self, frame: RFCOMM_Frame) -> None:
             f'rx_credits={self.rx_credits}: {data.hex()}'
         )
         if data:
-            if self.sink:
-                self.sink(data)  # pylint: disable=not-callable
+            asyncio.create_task(self._dispatch_data(data))
```
barbibulle, line 557: Creating a task for each packet may be a bit heavy. An alternative is to use an async queue, and a single task that waits for `opened`, then takes packets from the queue and forwards them to the sink.

zxzxwu: Waiting for opened as a task, but keeping other packets synchronously delivered might not (?) promise the order of packets. How about making a queue as a default sink, and let sink become a property - when it's set, dump all queued packets sink to the handler?

barbibulle: I mean delivering packets synchornously to a queue, and has the async task pull from that queue. That guarantees the order.
Something like this: https://github.com/google/bumble/blob/22eb405fde31628c64ff1a1a7adc584bf140fe42/apps/speaker/speaker.py#L150
(in your case, the `pump_packets` task would first wait for `opened` before the `While`).

bumble/rfcomm.py
```diff
@@ -466,17 +468,30 @@ def __init__(
         self.state = DLC.State.INIT
         self.role = multiplexer.role
         self.c_r = 1 if self.role == Multiplexer.Role.INITIATOR else 0
-        self.sink = None
         self.connection_result = None
         self.drained = asyncio.Event()
         self.drained.set()
+        # Queued packets when sink is not set.
+        self._packet_queue = asyncio.Queue(maxsize=32)
```
barbibulle, line 475: It doesn't seem that you need an async queue here, since no one if awaiting its elements. A simple `dequeue` would work just fine.

bumble/rfcomm.py
barbibulle: The other model you implemented, with a queue to hold the packets until a sink is set, works fine too. It does require the user to be careful to not set the sink too early, but since this is for internal use, we an easily guarantee that.

barbibulle (approved):

bumble/rfcomm.py
zxzxwu: Done

PR #489, opened 2024-05-21
basic auracast app

barbibulle:
This app is intended to support auracast features. To start with, only scanning for broadcasts is implemented. More functionality, including sending and receiving broadcasts, will be added in future PRs.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -16,22 +16,21 @@
 # Imports
 # -----------------------------------------------------------------------------
 from __future__ import annotations
-from enum import IntEnum
-import copy
-import functools
-import json
 import asyncio
-import logging
-import secrets
-import sys
+from collections.abc import Iterable
 from contextlib import (
     asynccontextmanager,
     AsyncExitStack,
     closing,
-    AbstractAsyncContextManager,
 )
+import copy
 from dataclasses import dataclass, field
-from collections.abc import Iterable
+from enum import Enum, IntEnum
+import functools
+import json
+import logging
+import secrets
+import sys
 from typing import (
     Any,
     Callable,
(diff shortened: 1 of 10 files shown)


bumble/pbp.py
```diff
@@ -0,0 +1,54 @@
...
+    @classmethod
+    def from_bytes(cls, data: bytes) -> Self:
+        features = cls.Features(data[0])
+        metadata_length = data[1]
+        metadata_ltv = data[1 : 1 + metadata_length]
+        return cls(
+            features=features, metadata=le_audio.Metadata.from_bytes(metadata_ltv)
+        )
+
+    def __str__(self) -> str:
```
zxzxwu, line 48: Shouldn't dataclass provide default __str__?

barbibulle: Yes. This is a leftover from a previous version of the class. I'll remove it.

bumble/bap.py
```diff
@@ -0,0 +1,109 @@
...
+
+from bumble import hci
+from bumble import le_audio
+
+
+# -----------------------------------------------------------------------------
+# Classes
+# -----------------------------------------------------------------------------
+@dataclasses.dataclass
+class BroadcastAudioAnnouncement:
```
zxzxwu, line 31: Can we host BAP libraries together in `profiles/bap.py`?

barbibulle: Yes, of course! It should be there. I'll make the change.

bumble/device.py
```diff
@@ -1706,6 +1988,18 @@ def find_connection_by_bd_addr(
 
         return None
 
+    def lookup_periodic_advertising_sync(
```
zxzxwu, line 1991: This could be unnecessary if PA table is a dict mapped from handle?

bumble/device.py
```diff
@@ -795,6 +866,201 @@ def on_termination(self, status: int) -> None:
...
+                advertising_sid=self.sid,
+                advertiser_address_type=self.advertiser_address.address_type,
+                advertiser_address=self.advertiser_address,
+                skip=self.skip,
+                sync_timeout=int(self.sync_timeout * 100),
+                sync_cte_type=0,
+            )
+        )
+        if response.status != HCI_Command_Status_Event.PENDING:
+            raise HCI_StatusError(response)
```
zxzxwu, line 948: Could be replaced with `send_command(..., check_result=True)` (HCI_SUCCESS and HCI_PENDING are both 0)

barbibulle: Good point! I now realize that there's a bunch of places in the code base where the same idea applies. I'll make a separate PR to unify that.

bumble/device.py
barbibulle: That wouldn't work, because the `periodic_advertising_syncs` list contains entries that don't have a handle yet (the handle is only assigned later when the controller sends a sync establishment event)

zxzxwu (approved):

PR #490, opened 2024-05-21
Fix: Preserve transport metadata

SilverBzH:
Preserve transport metadata when wrapping with SnoopingTransport

### DIFF
--- a/bumble/transport/common.py
+++ b/bumble/transport/common.py
@@ -425,6 +425,10 @@ class SnoopingTransport(Transport):
     class Source:
         sink: TransportSink
 
+        @property
+        def metadata(self) -> dict[str, Any]:
+            return getattr(self.source, 'metadata', {})
+
         def __init__(self, source: TransportSource, snooper: Snooper):
             self.source = source
             self.snooper = snooper


bumble/transport/__init__.py
```diff
@@ -40,9 +40,16 @@ def _wrap_transport(transport: Transport) -> Transport:
     # If BUMBLE_SNOOPER is set, try to automatically create a snooper.
     if snooper_spec := os.getenv('BUMBLE_SNOOPER'):
         try:
-            return SnoopingTransport.create_with(
+            wrapped_transport = SnoopingTransport.create_with(
                 transport, create_snooper(snooper_spec)
             )
+
+            # Transfer and extend metadata
+            if hasattr(transport, "source"):
```
barbibulle, line 48: It may be simpler to just add a `metadata`  `@property` on the `SnoopingTransport.Source` class that returns the wrapped source's metadata (no copy needed, it's Ok to pass the wrapped object's properties directly up).

SilverBzH: Agree, that's better. Just done it.

barbibulle (approved):

Issue #476, opened 2024-04-26
Duplicate HCI Reset When Tearing Down and Re-Creating Transport


hkpeprah:
## Summary
I am seeing a weird issue where if I tear down a transport, then re-create it, that the device fails to power on. The following example reproduces it. When the second `power_on()` is invoked, an error is raised (given below), and the attempt fails.

```
import asyncio
import logging

import bumble.device
import bumble.transport


async def open_transport(transport_name):
    link = await bumble.transport.open_transport_or_link(transport_name)
    dev = bumble.device.Device.with_hci("bumble", "F0:F1:F2:F3:F4:F5", link.source, link.sink)
    await dev.power_on()
    return link, dev


async def close_transport(link, dev):
    await dev.power_off()
    await dev.host.reset()
    await link.close()


async def main():
    logging.basicConfig(level=logging.DEBUG)
    for _ in range(2):
        link, dev = await open_transport("usb:0a12:0001")
        asyncio.sleep(1)
        await close_transport(link, dev)
    print("Done")


if __name__ == "__main__":
    asyncio.run(main())
```

## Error
```
DEBUG:bumble.transport.usb:USB Device: Bus 001 Device 048: ID 0a12:0001
DEBUG:bumble.transport.usb:selected endpoints: configuration=1, interface=0, setting=0, acl_in=0x82, acl_out=0x02, events_in=0x81,
DEBUG:bumble.transport.usb:auto-detaching kernel driver
DEBUG:bumble.transport.usb:current configuration = 1
DEBUG:bumble.transport.usb:starting USB event loop
DEBUG:bumble.drivers:Probing driver class: rtk
DEBUG:bumble.drivers.rtk:USB device (0A12, 0001) not in known list
DEBUG:bumble.drivers:Probing driver class: intel
DEBUG:bumble.drivers.intel:USB device (0A12, 0001) not in known list
DEBUG:bumble.host:### HOST -> CONTROLLER: HCI_RESET_COMMAND
DEBUG:bumble.host:### CONTROLLER -> HOST: HCI_COMMAND_COMPLETE_EVENT:
  num_hci_command_packets: 1
  command_opcode:          HCI_RESET_COMMAND
  return_parameters:       HCI_SUCCESS
DEBUG:bumble.host:### HOST -> CONTROLLER: HCI_READ_LOCAL_SUPPORTED_COMMANDS_COMMAND
[...]

PR #497, opened 2024-05-28
[Bug] Edimax BLE Dongle Fails After Teardown and Re-Instantiation

hkpeprah:
## Overview
This PR addresses an issue where the some RTK BLE dongles fail to perform an HCI reset after the transport is torn down and re-instantiated. To address that, we prevent crashing the background threads when invalid data comes in, and handle resetting the parser. A retry for the HCI reset has been implemented for the RTK dongles where errant data comes in after being torn down. This is related to #476.

## Setup
1. Dongle: [Edimax BT-8500 Nano BT Adapter](https://www.amazon.com/dp/B08M1VJHVD)
2. OS: `Ubuntu 20.04.6 LTS`
3. Firmware: [`BT-8500 Linux Bluetooth Driver 1.0.0.3`](https://www.edimax.com/edimax/download/download/data/edimax/global/download/bluetooth/bt-8500) (Also tried the [Linux Kernel FW](https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/plain/rtl_bt/))

### Issue 1
I observed that when closing and re-opening a transport to the Edimax BT Dongle (which has a RealTek chipset), that an HCI Reset would sometimes result in a second HCI reset response on the next command, requiring two HCI resets to be performed in the driver specific code.

### Issue 2
I observed that when the first HCI reset was performed after closing and re-opening a transport to the Edimax BT Dongle that sometimes the call would time out. Digging into the packet data, I observed that the data was sometimes invalid (examples follow):

```
b'\x04\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x0e\x04\x03\x03\x0c\x00'
```

```
b'\x04\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```

These would cause the background threads to crash, either from parsing the packet data, failing to handle the packet in `on_hci_packet()`, or failing to feed the data due to an invalid packet type.

## Proposed Changes
1. When performing an HCI reset in RTK's `driver_info_for_host`, perform the reset twice.
2. Expose the `hci_source` from the `Host`, and if the HCI reset fails in RTK's `driver_info_for_host`, clear the parser; this addresses the issue where we get errant packet data that causes the reset we get from the second command to fail to parse.
3. Add optional `timeout` to `send_command()` to allow for timing out the command.
4. In `host.on_packet()` capture when `hci.HCI_Packet.from_bytes()` fails.
5. Changed `raise ValueError` to `reset` and `break` in `ParserSource.feed_data()` to prevent crashing background thread when invalid packet type is found; allows the parser to continue parsing with the next bit of (hopefully good) data.

### DIFF
--- a/bumble/host.py
+++ b/bumble/host.py
@@ -152,6 +152,7 @@ class Host(AbortableEventEmitter):
     acl_packet_queue: Optional[AclPacketQueue] = None
     le_acl_packet_queue: Optional[AclPacketQueue] = None
     hci_sink: Optional[TransportSink] = None
+    hci_source: Optional[TransportSource] = None
     hci_metadata: Dict[str, Any]
     long_term_key_provider: Optional[
         Callable[[int, bytes, int], Awaitable[Optional[bytes]]]
@@ -506,6 +507,7 @@ class Host(AbortableEventEmitter):
     def set_packet_source(self, source: TransportSource) -> None:
         source.set_packet_sink(self)
         self.hci_metadata = getattr(source, 'metadata', self.hci_metadata)
+        self.hci_source = source
 
     def send_hci_packet(self, packet: hci.HCI_Packet) -> None:
         logger.debug(f'{color("### HOST -> CONTROLLER", "blue")}: {packet}')
@@ -514,7 +516,7 @@ class Host(AbortableEventEmitter):
         if self.hci_sink:
             self.hci_sink.on_packet(bytes(packet))
 
-    async def send_command(self, command, check_result=False):
+    async def send_command(self, command, check_result=False, timeout=None):
         # Wait until we can send (only one pending command at a time)
         async with self.command_semaphore:
             assert self.pending_command is None
@@ -526,7 +528,8 @@ class Host(AbortableEventEmitter):
 
             try:
                 self.send_hci_packet(command)
-                response = await self.pending_response
+                await asyncio.wait_for(self.pending_response, timeout=timeout)
+                response = self.pending_response.result()
 
                 # Check the return parameters if required
                 if check_result:
@@ -625,14 +628,19 @@ class Host(AbortableEventEmitter):
 
     # Packet Sink protocol (packets coming from the controller via HCI)
     def on_packet(self, packet: bytes) -> None:
-        hci_packet = hci.HCI_Packet.from_bytes(packet)
+        try:
+            hci_packet = hci.HCI_Packet.from_bytes(packet)
+        except Exception as error:
+            logger.warning(f'!!! error parsing packet from bytes: {error}')
+            return
+
         if self.ready or (
             isinstance(hci_packet, hci.HCI_Command_Complete_Event)
             and hci_packet.command_opcode == hci.HCI_RESET_COMMAND
         ):
             self.on_hci_packet(hci_packet)
         else:
-            logger.debug('reset not done, ignoring packet from controller')
+            logger.debug(f'reset not done, ignoring packet from controller: {hci_packet}')
 
     def on_transport_lost(self):
         # Called by the source when the transport has been lost.
--- a/bumble/host.py
+++ b/bumble/host.py
@@ -516,7 +516,7 @@ class Host(AbortableEventEmitter):
         if self.hci_sink:
             self.hci_sink.on_packet(bytes(packet))
 
-    async def send_command(self, command, check_result=False, timeout=None):
+    async def send_command(self, command, check_result=False, response_timeout: Optional[int] = None):
         # Wait until we can send (only one pending command at a time)
         async with self.command_semaphore:
             assert self.pending_command is None
@@ -528,7 +528,7 @@ class Host(AbortableEventEmitter):
 
             try:
                 self.send_hci_packet(command)
-                await asyncio.wait_for(self.pending_response, timeout=timeout)
+                await asyncio.wait_for(self.pending_response, timeout=response_timeout)
                 response = self.pending_response.result()
 
                 # Check the return parameters if required
--- a/bumble/host.py
+++ b/bumble/host.py
@@ -640,7 +640,9 @@ class Host(AbortableEventEmitter):
         ):
             self.on_hci_packet(hci_packet)
         else:
-            logger.debug(f'reset not done, ignoring packet from controller: {hci_packet}')
+            logger.debug(
+                f'reset not done, ignoring packet from controller: {hci_packet}'
+            )
 
     def on_transport_lost(self):
         # Called by the source when the transport has been lost.
--- a/bumble/transport/common.py
+++ b/bumble/transport/common.py
@@ -132,7 +132,9 @@ class PacketParser:
                         packet_type
                     ) or self.extended_packet_info.get(packet_type)
                     if self.packet_info is None:
-                        raise ValueError(f'invalid packet type {packet_type}')
+                        logger.warning(f'ignoring invalid packet type: {packet_type}')
+                        self.reset()
+                        break
                     self.state = PacketParser.NEED_LENGTH
                     self.bytes_needed = self.packet_info[0] + self.packet_info[1]
                 elif self.state == PacketParser.NEED_LENGTH:
--- a/tests/transport_test.py
+++ b/tests/transport_test.py
@@ -62,11 +62,10 @@ def test_parser_extensions():
     except ValueError:
         exception_thrown = True
 
-    assert exception_thrown
+    assert not exception_thrown
 
     # Now add a custom info
     parser.extended_packet_info[0x77] = (1, 1, 'B')
-    parser.reset()
     parser.feed_data(bytes([0x77, 0x00, 0x02, 0x01, 0x02]))
     assert len(sink.packets) == 1
 
--- a/tests/transport_test.py
+++ b/tests/transport_test.py
@@ -55,7 +55,8 @@ def test_parser_extensions():
     sink = Sink()
     parser = PacketParser(sink)
 
-    # Check that an exception is thrown for an unknown type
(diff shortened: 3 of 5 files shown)


bumble/host.py
```diff
@@ -514,7 +516,7 @@ def send_hci_packet(self, packet: hci.HCI_Packet) -> None:
         if self.hci_sink:
             self.hci_sink.on_packet(bytes(packet))
 
-    async def send_command(self, command, check_result=False):
+    async def send_command(self, command, check_result=False, timeout=None):
```
barbibulle, line 519: `send_command` is async, so we don't need to add a `timeout` parameter to it in order to wait at most some amount of time: we can leave that up to the caller, which can invoke `await asyncio.wait_for(host.send_command(...), timeout)`.

tests/transport_test.py
```diff
@@ -62,11 +62,10 @@ def test_parser_extensions():
     except ValueError:
         exception_thrown = True
 
-    assert exception_thrown
+    assert not exception_thrown
```
barbibulle, line 65: This of course makes the statement on line 58 no longer valid, so it should be updated.

hkpeprah: Whoops. Missed that.
- [ ] Update comment.

bumble/host.py
hkpeprah: Sorry, I'm a bit new to AsyncIO. If the calling thread times out on the co-routine, while the co-routine terminate and yield the lock so that another `send_command()` can run?

barbibulle: with `wait_for`, if the timeout expires before the coroutine returns, the coroutine is canceled, which means that here the `await self.pending_response` with raise a TimeoutError, which will exit out of the `async with self.command_semaphore`, releasing the semaphore, so if another task was waiting on it, it can now proceed.

The only small different between putting the `wait_for` on the outside, rather than just around `asyncio.wait_for(self.pending_response,...)` is that the timeout encompasses not only waiting for the response, but also waiting for the semaphore before that.
So, maybe there's a case to be made for specifying an inner timeout, like you have it, which only applies to waiting for the response. I can see how this could make sense, even if I don't have an actual use case in mind. So maybe we leave things the way you have them here, but rename the `timeout` parameter to `response_timeout`, to make it clear that it's not a timeout on the whole method, but just on the waiting for a response inside that method.

bumble/transport/common.py
```diff
@@ -132,7 +132,9 @@ def feed_data(self, data: bytes) -> None:
                         packet_type
                     ) or self.extended_packet_info.get(packet_type)
                     if self.packet_info is None:
-                        raise ValueError(f'invalid packet type {packet_type}')
+                        logger.warning(f'ignoring invalid packet type: {packet_type}')
```
barbibulle, line 135: the tradeoff here is that if there's a data source with corrupted data, the parser will go on forever without ever throwing an exception. I think that's Ok. If we encounter cases where that happens, we can always enhance the parser with a property that configures how much  invalid data it is willing to ignore.

bumble/host.py
hkpeprah: I can do that!
- [ ] Rename `timeout -> response_timeout`

bumble/transport/common.py
hkpeprah: I think I found that when I got invalid packet type data in back-to-back packets, eventually I got a valid packet, which allowed me to continue. With the exception, the background thread crashed and my application code wasn't aware of it without an `wait asyncio.wait_for()` wrap-around the `.power_on()` method. I think it's possible with this change that the rest of the packet data could contain a start of a valid packet after the invalid data, and since we don't throw the exception, now we miss it, and end up in the same situation of waiting indefinitely, but I think in that case, the behaviour is the same.

bumble/host.py
```diff
@@ -152,6 +152,7 @@ class Host(AbortableEventEmitter):
     acl_packet_queue: Optional[AclPacketQueue] = None
     le_acl_packet_queue: Optional[AclPacketQueue] = None
     hci_sink: Optional[TransportSink] = None
+    hci_source: Optional[TransportSource] = None
```
barbibulle, line 155: This isn't needed. The host doesn't need to keep a reference to the source that's feeding it. It does need to keep a reference to the sink that it feeds, but the source only comes into the picture when calling `set_packet_source`, which is a convenience method for telling the source that its sink is the host object.

hkpeprah: I used that in the RTK driver when we get invalid packet data, as I noticed sometimes I got a large packet that had a payload of `254`, but never got the remaining packet data for that packet:
```
            if host.hci_source and hasattr(host.hci_source, "parser"):
                host.hci_source.parser.reset()
```
Is there another way we can reset the parser from the driver? The packet I was getting from the dongle had the string `RTK_5` or something like that.

hkpeprah: Maybe a `parser` property or some other `reset` exposed from the `host()`?

hkpeprah: @barbibulle, any thoughts on a way of working around this?

barbibulle: Sorry for the delay. There are a couple issues to work through here:
1/ The source/sink relationship is a push model, and source is a purely abstract concept of "some object that pushes data to a sink". A sink is just an object that has a `on_packet` method. This means that an object like `Host`, which receives HCI data from a "source" doesn't necessarily know what the source is, it just receives packets. The `Host.set_packet_source` method is purely a convenience function that tells the source where to push data (and looks up a `metadata` property on the source, but that isn't a requirement. In theory, one could attach a source to a `Host` without calling that method. This means that looking for an `host.hci_source` may or may not work depending on how objects were instantiated.

2/ Not all sources have a `parser` property. It is common to see it, because most sources are `Transport` implementations, which often need to extract HCI packets from a stream, and they usually use a `PacketParser` for that. But you could very well have a source that doesn't need to parse an HCI stream (for example, a transport where HCI packets arrive already delimited). So calling something like `host.hci_source.parser.reset()` only works for a transport that happens to use a `PacketParser` and happens to store that parser in a property named `parser`. That could break at any time, since that's not a requirement of transport implementations.

But in this case, the problem we need to solve is narrow, so we should be able to address it differently.
I'll try a few things out over the next week to see if I can think of a maintainable way to achieve what's needed here (i.e without having to put too many constraints on what types of objects and classes can be used as HCI data sources).

hkpeprah: I could break out the changes into two separate PRs:

1. One for the RTK / HCI Source
2. The other for the parser fixes.

Would that be better?

barbibulle: Sorry but took so long to follow up. Yes, breaking up in two PRs would help. For the parser changes, I’ll try to find some time over the weekend to come up with an alternative.

barbibulle: Update: I though about a maintainable way to allow signaling of errors back to an HCI source so that it can reset itself, without creating an explicit reference from a sink to a source (i.e no storing of `hci_source` in the `Host` object): a source can listen for events from its sink (first testing if the sink is an EventEmitter, since not all are required to be), then, in the RTK driver case here, the driver could have the host emit an error (upon encountering the timeout), which the usb source would get called back for, and reset itself.  But... there's probably an even simpler way: it turns out that the usb source really doesn't need to use an HCI parser, since packets coming from USB are always delimited. The current implementation subclasses ParserSource by convenience, and uses `self.parser.feed_data(packet)` to emit packets to its sink, but that's really not needed. I'll push a PR where the USB source doesn't use a parser anymore, which should solve the issue here, since even after receiving corrupted data in one USB transfer, the next USB transfer should be a whole packet again.

hkpeprah:
Replaced by #534 .

bumble/host.py
hkpeprah: Created #534 . Thank you!

PR #499, opened 2024-06-04
rfcomm bridge app

barbibulle:
Simple app that offers a TCP<->RFCOMM bridge. A client mode and a server more are supported.
In client mode, the bridge waits for an inbound TCP connection, then initiates an outbound RFCOMM connection, then bridges the two bi-directionally.
In server mode, the bridge waits for an inbound RFCOMM connection, then initiates an outbound TCP connection, then bridges the two bi-directionally.

### DIFF
--- a/bumble/rfcomm.py
+++ b/bumble/rfcomm.py
@@ -446,10 +446,6 @@ class DLC(EventEmitter):
         DISCONNECTED = 0x04
         RESET = 0x05
 
-    connection_result: Optional[asyncio.Future]
-    _sink: Optional[Callable[[bytes], None]]
-    _enqueued_rx_packets: collections.deque[bytes]
-
     def __init__(
         self,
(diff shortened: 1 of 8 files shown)


bumble/rfcomm.py
```diff
@@ -609,6 +629,19 @@ def connect(self) -> None:
         self.connection_result = asyncio.get_running_loop().create_future()
         self.send_frame(RFCOMM_Frame.sabm(c_r=self.c_r, dlci=self.dlci))
 
+    async def disconnect(self) -> None:
```
zxzxwu, line 632: Shall we cover new functions in rfcomm_test?

zxzxwu (approved):

bumble/rfcomm.py
barbibulle: Yes. Done.

PR #501, opened 2024-06-04
Reorganize exceptions

zxzxwu:
Replace builtin errors so they could be handled more properly. 

* Add BaseBumbleException as a "real" root error
* Add several core error classes and properly replace builtin errors with them
* Add several error classes for specific modules (transport, device)

Errors raised in Apps, Examples, Tests and Pandora are left because they are unlikely to be handled.

### DIFF
--- a/bumble/core.py
+++ b/bumble/core.py
@@ -76,7 +76,13 @@ def get_dict_key_by_value(dictionary, value):
 # -----------------------------------------------------------------------------
 # Exceptions
 # -----------------------------------------------------------------------------
-class BaseError(Exception):
+
+
+class BaseBumbleError(Exception):
+    """Base Error raised by Bumble."""
+
+
+class BaseError(BaseBumbleError):
     """Base class for errors with an error code, error name and namespace"""
 
     def __init__(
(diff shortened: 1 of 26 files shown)


bumble/core.py
```diff
@@ -115,18 +121,38 @@ class ProtocolError(BaseError):
...
+
+class InvalidPacketError(BaseBumbleError, ValueError):
+    """Invalid Packet Error"""
+
+
+class InvalidOperationError(BaseBumbleError, ValueError):
+    """Invalid Operation Error"""
+
+
+class OutOfResourcesError(BaseBumbleError, ValueError):
```
barbibulle, line 148: Should be a subclass of `RuntimeError`

bumble/core.py
```diff
@@ -115,18 +121,38 @@ class ProtocolError(BaseError):
...
 
+class InvalidArgumentError(BaseBumbleError, ValueError):
+    """Invalid Argument Error"""
+
+
+class InvalidPacketError(BaseBumbleError, ValueError):
+    """Invalid Packet Error"""
+
+
+class InvalidOperationError(BaseBumbleError, ValueError):
```
barbibulle, line 144: should be a subclass of `RuntimeError`

bumble/core.py
```diff
@@ -115,18 +121,38 @@ class ProtocolError(BaseError):
...
+    """Invalid Operation Error"""
+
+
+class OutOfResourcesError(BaseBumbleError, ValueError):
+    """Out of Resources Error"""
+
+
+class UnreachableError(BaseBumbleError, ValueError):
```
barbibulle, line 152: Maybe `ValueError` isn't needed here as a base class.

bumble/transport/common.py
```diff
@@ -49,10 +50,16 @@
...
-    """
+class TransportLostError(core.BaseBumbleError):
+    """The Transport has been lost/disconnected."""
+
+
+class TransportInitError(core.BaseBumbleError):
+    """Error raised when the transport cannot be initialized."""
+
+
+class TransportSpecError(core.BaseBumbleError):
```
barbibulle, line 61: Should subclass `ValueError` as well.

bumble/core.py
zxzxwu: Done

bumble/core.py
zxzxwu: Done

bumble/core.py
zxzxwu: Done

bumble/transport/common.py
zxzxwu: Done

barbibulle (approved):

PR #502, opened 2024-06-05
support out of order advertising set termination / connection events

barbibulle:
Some controllers (like the ASUS BT500) send the advertising set termination event *before* the connection complete event. At that time, the Connection object for connection_handle in the event isn't yet known. When that happens, we need to support remembering the address information until the connection complete event arrives.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -1574,6 +1574,7 @@ class Device(CompositeEventEmitter):
 
         # Extended advertising.
         self.extended_advertising_sets: Dict[int, AdvertisingSet] = {}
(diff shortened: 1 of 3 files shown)


zxzxwu (approved):
Would be nice to have a test for it
like https://github.com/google/bumble/blob/f910a696ad5d286201d40480398393b1e9bf5237/tests/device_test.py#L362C11-L362C47

barbibulle:
> Would be nice to have a test for it like https://github.com/google/bumble/blob/f910a696ad5d286201d40480398393b1e9bf5237/tests/device_test.py#L362C11-L362C47

done

PR #503, opened 2024-06-06
'override' from typing_extensions has no effect.

Gopi-SB:
Hi Team .

'override' does not support in Python < 3.12. and 'override' from typing_extensions has no effect.
 so removing this library import


zxzxwu:
No, that's why typing_extensions is needed.

`@override` is introduced to typing in 3.12, but since typing_extensions is not a built-in module, it could be used in any release still under maintainance.

See https://typing-extensions.readthedocs.io/en/latest/

PR #504, opened 2024-06-10
bench role change

barbibulle:
Allow requesting a specific role switch after connection in the Bench app.

### DIFF
--- a/apps/bench.py
+++ b/apps/bench.py
@@ -40,6 +40,8 @@ from bumble.hci import (
     HCI_LE_1M_PHY,
     HCI_LE_2M_PHY,
     HCI_LE_CODED_PHY,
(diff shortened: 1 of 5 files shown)


apps/bench.py
```diff
@@ -1000,6 +1008,8 @@ def __init__(
         self.max_credits = max_credits
         self.credits_threshold = credits_threshold
         self.dlc = None
+        self.max_credits = max_credits
+        self.credits_threshold = credits_threshold
```
zxzxwu, line 1012: Repeated?

zxzxwu (approved):

apps/bench.py
barbibulle: Fixed.

PR #505, opened 2024-06-12
usb_probe.py: Add interactive mode for finding a specific device

AlanRosenthal:
````
❯ python apps/usb_probe.py --interactive
Unplug your device and press the enter key!
Plug in your device and press the enter key!

ID 2FE3:000B
  Bumble Transport Names: usb:0 or usb:2FE3:000B
  Bus/Device:             001/031
  Class:                  Wireless Controller
  Subclass/Protocol:      1/1 [Bluetooth]
````

### DIFF
--- a/apps/usb_probe.py
+++ b/apps/usb_probe.py
@@ -168,7 +168,13 @@ def is_bluetooth_hci(device):
 # -----------------------------------------------------------------------------
 @click.command()
 @click.option('--verbose', is_flag=True, default=False, help='Print more details')
-def main(verbose):
+@click.option(
+    '--interactive',
+    is_flag=True,
+    default=False,
+    help='Detech a device by unplugging / replugging it in',
+)
+def main(verbose, interactive):
     logging.basicConfig(level=os.environ.get('BUMBLE_LOGLEVEL', 'WARNING').upper())
 
     load_libusb()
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: apps/usb_probe.py [hunk 1]
```diff
@@ -168,15 +168,34 @@ def is_bluetooth_hci(device):
...
     logging.basicConfig(level=os.environ.get('BUMBLE_LOGLEVEL', 'WARNING').upper())
 
     load_libusb()
     with usb1.USBContext() as context:
         bluetooth_device_count = 0
         devices = {}
 
-        for device in context.getDeviceIterator(skip_on_error=True):
+        if interactive:
+            input("Unplug your device and press the enter key!")
```
barbibulle, line 186: It would be nice to also support USB hotplug events to re-scan automatically when a device is removed/added, so you don't have to type anything on the keyboard.

apps/usb_probe.py
```diff
@@ -168,15 +168,34 @@ def is_bluetooth_hci(device):
 # -----------------------------------------------------------------------------
 @click.command()
 @click.option('--verbose', is_flag=True, default=False, help='Print more details')
-def main(verbose):
+@click.option(
+    '--interactive',
+    is_flag=True,
+    default=False,
+    help='Detech a device by unplugging / replugging it in',
```
barbibulle, line 175: Detech -> Detect

barbibulle (approved):

PR #506, opened 2024-07-03
a2dp: emit delay_report

barbibulle:
Delay report commands were not handled. This handles them by emitting an event with the delay value.

### DIFF
--- a/bumble/avdtp.py
+++ b/bumble/avdtp.py
@@ -2162,7 +2162,7 @@ class LocalStreamEndPoint(StreamEndPoint, EventEmitter):
     def on_abort_command(self):
         self.emit('abort')
 
-    def on_delayreport_command(self, delay):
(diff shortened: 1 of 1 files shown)


bumble/avdtp.py
```diff
@@ -2162,6 +2162,9 @@ def on_close_command(self):
     def on_abort_command(self):
         self.emit('abort')
 
+    def on_delayreport_command(self, delay):
+        self.emit('delay_report', delay)
```
zxzxwu, line 2166: Though this is accessed by reflection, could you add type annotation to help user understand the type of emitted value?

zxzxwu (approved):

bumble/avdtp.py
barbibulle: done.

PR #507, opened 2024-07-09
Create `inv web.build`

AlanRosenthal:
This command will build a wheel, copy it in the web directory, and create a file `packageFile` with the name of the wheel. If the correct override param is given, bumble.js will read `packageFile` and load that package.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -20,7 +20,10 @@ Invoke tasks
 # Imports
 # -----------------------------------------------------------------------------
 import os
-
+import glob
+import shutil
+import urllib
+from pathlib import Path
 from invoke import task, call, Collection
 from invoke.exceptions import Exit, UnexpectedExit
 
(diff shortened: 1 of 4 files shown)


tasks.py
```diff
@@ -205,5 +208,21 @@ def __init__(self, *args, **kwargs):
...
+# -----------------------------------------------------------------------------
+@task
+def web_build(ctx):
+    # Step 1: build the wheel
+    build(ctx)
+    # Step 2: Copy the wheel to the web folder, so the http server can access it
+    newest_wheel = Path(max(glob.glob('dist/*.whl'), key=lambda f: os.path.getmtime(f)))
+    shutil.copy(newest_wheel, Path('web/'))
+    # Step 3: Write wheel's name to web/packageFile
+    with open(Path('web', 'packageFile'), mode='w') as package_file:
```
barbibulle, line 220: It might be simpler to just copy the built `.whl` file to just `web/bumble.whl` rather than go through the indirection of a text file pointing to the built file.
Then the web page can just reference `?package=bumble.whl`.

AlanRosenthal: Unfortunately doing that results in this error message:
```
packaging.utils.InvalidWheelFilename: Invalid wheel filename (wrong number of parts): bumble
```

barbibulle: Bummer.

barbibulle (approved):

PR #510, opened 2024-07-10
device.py: Update get_characteristics_by_uuid()

AlanRosenthal:
`get_characteristics_by_uuid()` now allows a UUID to be passed to the service param. This allows for users to easily query for a service uuid and characteristic uuid with one API.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -27,6 +27,7 @@ import copy
 from dataclasses import dataclass, field
 from enum import Enum, IntEnum
 import functools
+import itertools
 import json
 import logging
 import secrets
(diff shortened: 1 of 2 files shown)


bumble/device.py
```diff
@@ -1172,8 +1172,17 @@ def get_services_by_uuid(self, uuid: core.UUID) -> List[gatt_client.ServiceProxy
         return self.gatt_client.get_services_by_uuid(uuid)
 
     def get_characteristics_by_uuid(
-        self, uuid: core.UUID, service: Optional[gatt_client.ServiceProxy] = None
+        self,
+        uuid: core.UUID,
+        service: Optional[gatt_client.ServiceProxy | core.UUID] = None,
```
barbibulle, line 1177: The '|' type syntax is only for Python >= 3.10, but we still support down to 3.8, so this should be using `Union`

bumble/device.py
```diff
@@ -1172,8 +1172,17 @@ def get_services_by_uuid(self, uuid: core.UUID) -> List[gatt_client.ServiceProxy
...
     ) -> List[gatt_client.CharacteristicProxy]:
+        if isinstance(service, core.UUID):
+            services = self.get_services_by_uuid(service)
+            characteristics = []
+            for service in services:
+                characteristics.extend(self.get_characteristics_by_uuid(uuid, service))
+            return characteristics
```
barbibulle, line 1184: suggestion: use a comprehension: `return [self.get_characteristics_by_uuid(uuid, s) for s in self.get_services_by_uuid(service)]`, or a map: `return list(self.get_services_by_uuid(service).map(functools.partial(self.get_characteristics_by_uuid(uuid))))`

barbibulle (approved):

PR #514, opened 2024-07-16
Rework some device and hci functions

zxzxwu:
* Add type hints
* Add wait_for_complete behavior for async commands

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -1493,17 +1493,18 @@ class Connection(CompositeEventEmitter):
         self.remove_listener('disconnection', abort.set_result)
         self.remove_listener('disconnection_failure', abort.set_exception)
 
-    async def set_data_length(self, tx_octets, tx_time) -> None:
+    async def set_data_length(self, tx_octets: int, tx_time: int) -> None:
         return await self.device.set_data_length(self, tx_octets, tx_time)
 
     async def update_parameters(
         self,
-        connection_interval_min,
-        connection_interval_max,
-        max_latency,
-        supervision_timeout,
-        use_l2cap=False,
-    ):
+        connection_interval_min: int,
+        connection_interval_max: int,
+        max_latency: int,
+        supervision_timeout: int,
+        use_l2cap: bool = False,
+        wait_for_complete: bool = False,
+    ) -> None:
         return await self.device.update_connection_parameters(
             self,
             connection_interval_min,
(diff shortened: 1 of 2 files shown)


bumble/device.py
```diff
@@ -1493,37 +1493,55 @@ async def sustain(self, timeout: Optional[float] = None) -> None:
...
-        max_latency,
-        supervision_timeout,
-        use_l2cap=False,
-    ):
+        connection_interval_min: int,
+        connection_interval_max: int,
+        max_latency: int,
+        supervision_timeout: int,
+        use_l2cap: bool = False,
+        wait_for_complete: bool = False,
```
barbibulle, line 1506: maybe `wait_for_completion` would be a better name for this parameter.

bumble/device.py
```diff
@@ -1493,37 +1493,55 @@ async def sustain(self, timeout: Optional[float] = None) -> None:
...
-    async def set_phy(self, tx_phys=None, rx_phys=None, phy_options=None):
-        return await self.device.set_connection_phy(self, tx_phys, rx_phys, phy_options)
+    async def set_phy(
+        self,
+        tx_phys: Union[Iterable[int], int, None] = None,
+        rx_phys: Union[Iterable[int], int, None] = None,
+        phy_options: int = 0,
+        wait_for_complete: bool = False,
+    ) -> None:
+        """Sets PHY preference of this connection.
```
barbibulle, line 1525: Since this docstring is essentially the same as the one for `device.set_connection_phy`, maybe there's a clever way to just say "See XYZ" and refer to that other location, rather than have a copy of the docstring, which is hard to maintain in both places.

bumble/device.py
```diff
@@ -3382,36 +3406,67 @@ async def update_connection_parameters(
...
-        result = await self.send_command(
-            HCI_LE_Connection_Update_Command(
-                connection_handle=connection.handle,
-                connection_interval_min=connection_interval_min,
-                connection_interval_max=connection_interval_max,
-                max_latency=max_latency,
-                supervision_timeout=supervision_timeout,
-                min_ce_length=min_ce_length,
-                max_ce_length=max_ce_length,
+        with closing(EventWatcher()) as watcher:
```
barbibulle, line 3409: suggestion: move the call to `self.send_command()` to a local function, which the code followibng `with closing(...) ...:` can invoke, but can also be invoked without the `with` close when `wait_for_completion` is `False`, so that we don't unnecessarily create the context and the future in that case.

PR #516, opened 2024-07-16
Add Treadmill webapp

AlanRosenthal:


web/treadmill/treadmill.py
```diff
@@ -0,0 +1,172 @@
...
+        connection.on('security_request', self.on_security_request)
+        connection.on('pairing_failure', self.on_pairing_failure)
+        connection.on('pairing', self.on_pairing)
+        self.peer = Peer(connection)
+        print(f'Connected to {self.peer}')
+        print("Starting service discovery...")
+        await self.peer.discover_all()
+        print("Service discovery complete!")
+
+        hr_measurement_characteristics = self.peer.get_characteristics_by_uuid(
```
barbibulle, line 124: You could use the `HeartRateServiceProxy` class from the profiles. Something like: `heart_rate_service = peer.create_service_proxy(HeartRateServiceProxy)`.
(see `examples/heart_rate_client.py` for an example)

web/treadmill/treadmill.py
```diff
@@ -0,0 +1,172 @@
...
+            uuid=GATT_HEART_RATE_MEASUREMENT_CHARACTERISTIC,
+            service=GATT_HEART_RATE_SERVICE,
+        )
+        hr_measurement_characteristic = hr_measurement_characteristics[0]
```
barbibulle, line 128: It would be safe to check that the service actually has that characteristic, otherwise that's going to raise.

web/treadmill/treadmill.py
```diff
@@ -0,0 +1,172 @@
...
+                    'peer_address': str(connection.peer_address),
+                    'is_encrypted': "Yes" if connection.is_encrypted else "No",
+                }
+            )
+
+    def emit_on_security_request(self):
+        if listener := self.listeners.get('on_security_request'):
+            listener()
+
+    def on(self, event_name, listener):
```
barbibulle, line 96: You can just subclass `pyee.EventEmitter` and get all the listener management and event emission implemented already for you.

web/treadmill/treadmill.py
```diff
@@ -0,0 +1,172 @@
...
+            return
+
+        self.scan_entries[advertisement.address] = self.ScanEntry(advertisement)
+        self.emit_scanning_update()
+
+    @AsyncRunner.run_in_task()
+    async def on_connection(self, connection: Connection):
+        self.emit_connection_updates(connection)
+        connection.listener = self
+        connection.on('security_request', self.on_security_request)
```
barbibulle, line 115: If you want to prompt the user to accept/reject pairing, you should probably use the `PairingDelegate` mechanism rather than intercepting security requests. See `apps/pair.py` for an example (see how the `--prompt` option is handled, it does exactly that: ask the user to accept pairing or not).

PR #520, opened 2024-07-25
Generate RPA when own address type is RESOLVABLE_OR_RANDOM

Gopi-SB:
Generate RPA when own address type is RESOLVABLE_OR_RANDOM

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -2365,12 +2365,18 @@ class Device(CompositeEventEmitter):
         assert self.legacy_advertiser is None
         assert self.legacy_advertising_set is None
 
+        # Generate RPA
+        if own_address_type == OwnAddressType.RESOLVABLE_OR_RANDOM:
+            random_address = Address.generate_private_address(self.irk)
+        else:
+            random_address=self.random_address
+
         if self.supports_le_extended_advertising:
             # Use extended advertising commands with legacy PDUs.
             self.legacy_advertising_set = await self.create_advertising_set(
                 auto_start=True,
                 auto_restart=auto_restart,
-                random_address=self.random_address,
+                random_address=random_address,
                 advertising_parameters=AdvertisingParameters(
                     advertising_event_properties=(
                         AdvertisingEventProperties.from_advertising_type(
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -2366,8 +2366,10 @@ class Device(CompositeEventEmitter):
         assert self.legacy_advertising_set is None
 
         # Generate RPA
-        if own_address_type == OwnAddressType.RESOLVABLE_OR_RANDOM:
-            random_address = Address.generate_private_address(self.irk)
+        if self.address_resolution_offload == False:
+            if own_address_type == OwnAddressType.RESOLVABLE_OR_RANDOM:
+                own_address_type = OwnAddressType.RANDOM
+                random_address = Address.generate_private_address(self.irk)
         else:
             random_address=self.random_address
 


zxzxwu:
-1. As spec mentions in the description of OwnAddressType, setting this field to _or_resolvable means

> Controller generates Resolvable Private Address based on the local IRK from the resolving list. If the resolving list contains no matching entry, use the public (or random) address.

image

Since we already implement RPA generation offload(https://github.com/google/bumble/blob/main/bumble/device.py#L2209), setting an RPA here will break the base identity address if there is a corresponding entry in the resolving list. 

An alternative way is when RPA generation offload is not enabled (though I think this is a quite rare case), change the own_address_type to RANDOM but generate an RPA for it.

barbibulle:
As @zxzxwu explains, this PR won’t work as is. I have a WIP change to support basic RPA. I’ll upload a PR shortly, @Gopi-SB I think it should give you what you need if you want to support RPA in simple scenarios.

Gopi-SB:
> As @zxzxwu explains, this PR won’t work as is. I have a WIP change to support basic RPA. I’ll upload a PR shortly, @Gopi-SB I think it should give you what you need if you want to support RPA in simple scenarios.

@barbibulle Thanks Please upload the PR

Gopi-SB:
Hi @zxzxwu ,  I have modified PR based on your inputs. Please check and let me know.

Hi @barbibulle , We have some test case dependency on these code changes . Can you consider this PR as temporary fix till you upload your basic RPA support related PR

barbibulle:
> Hi @zxzxwu , I have modified PR based on your inputs. Please check and let me know.
> 
> Hi @barbibulle , We have some test case dependency on these code changes . Can you consider this PR as temporary fix till you upload your basic RPA support related PR

This PR doesn't really do what you want (for example, it won't work with pairing, and will lead to incorrect values reported in the self address for the connection).
If what you need temporarily is to use an RPA, why not just set your device random address to an address generated by `Address.generate_private_address(self.irk)`? (that will also not completely work with pairing, because you don't have the identity address, but that would be essentially equivalent to what you have in this PR, but with no code change)

zxzxwu:
If your testing controller (no matter it's physical or virtual) supports offloaded address generation, you may just enable it and let controller do the magic. Or, you may use `create_advertising_set()` API, which allows you to set random address, unless your controller doesn't support extended advertising.

Gopi-SB:
Hi @barbibulle @zxzxwu ,

Thanks for the suggestions .
To provide more clarity, I wanted to share my RPA address rotation test case requirement here .

Setup:
DUT : Android cuttlefish
Ref Device : Bumble stack with virtual Bluetooth Controller (RootCanal)

steps:
1.Pair and connect to Bumble over LE using RPA
2.Disconnect and remove bonding on Android
3.Restart the Bumble to rotate its MAC address (RPA is different than the previous)
4.Again pair and connect to Bumble over LE using RPA
5.Verify that identity address is same in both First and second pairing attempts
6.Verify that RPA address is different in First and second pairing attempts

Let me know if you have any inputs to achieve this test criteria using bumble.

Suggestion provided in previous comments does not help to achieve this requirement.

Thanks,
Gopi

zxzxwu:
This could be achieved without changing Bumble:
```py
# Advertise
random_address = Address.generate_private_address(device.irk)
advertising_set = await device.create_advertising_set(
    advertising_parameters=AdvertisingParameters(
        # You may change advertising properties
        advertising_event_properties=AdvertisingEventProperties(),
        peer_address=peer_address,
        own_address_type=OwnAddressType.RANDOM,
    ),
    random_address=random_address,
)
# Your test
...
```

It's actually same as the logic in your PR.

barbibulle:
Please try #522 , which should do what you need.

Gopi-SB:
> Please try #522 , which should do what you need.

Hi @barbibulle , Thanks . #522 works fine for my test case requirement . But i think host generated RPA address not required if controller supports offloaded address generation.  and we have some test cases with fixed random address(i.e config.address) to verify it on android side .if we enable **le_privacy_enabled** config , generated random value is always different than config.address.

barbibulle:
> > Please try #522 , which should do what you need.
> 
> Hi @barbibulle , Thanks . #522 works fine for my test case requirement . But i think host generated RPA address not required if controller supports offloaded address generation. and we have some test cases with fixed random address(i.e config.address) to verify it on android side .if we enable **le_privacy_enabled** config , generated random value is always different than config.address.

I've pushed an update to this PR that supports controller-generated RPA (but I don't recommend using that mode, it doesn't work with many controllers).
When you set `le_privacy_enabled` in a config, it is normal that the generated random address is going to be different from `config.address`. `config.address` is the devices' static random address. Whenever a new RPA is generated, the `device.random_address` is updated. The fixed static random address is `device.static_address`.

PR #522, opened 2024-08-01
add basic RPA support

barbibulle:
This PR adds a device config to tell the Device object to automatically create a random private address at power on. Optionally, there's also a periodic task to update the RPA based on a configurable timeout.
For manual control, the `update_rpa` method can be called.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -182,6 +182,7 @@ from .core import (
     BaseBumbleError,
     ConnectionParameterUpdateError,
     CommandTimeoutError,
+    ConnectionParameters,
     ConnectionPHY,
     InvalidArgumentError,
     InvalidOperationError,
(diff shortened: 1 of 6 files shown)


zxzxwu:
I am thinking whether RPA rotation should be provided by Device itself? Providing attributes to support host-side security could be great, but handling them automatically is a little bit complicated (there are many Android issues due to the same behavior). Since Bumble is not a product, letting users rotate RPA manually should be enough.

barbibulle:
> I am thinking whether RPA rotation should be provided by Device itself? Providing attributes to support host-side security could be great, but handling them automatically is a little bit complicated (there are many Android issues due to the same behavior). Since Bumble is not a product, letting users rotate RPA manually should be enough.

I agree. This is why in this PR, automatic RPA update is off by default. Users can call Device.update_rpa() if they want to. The le_privacy_enabled is only a convenient way to have the device pick an RPA at power on, which can be useful if you need an RPA with some of the example code or apps but without writing code: just set the property in a device config.

zxzxwu (approved):

PR #528, opened 2024-08-09
Fix CTKD failure introduced by Host RPA generation

zxzxwu:
* Fix the bug which device doesn't generate a static address by default
* Properly distribute public or random address if no identity address type is set
* Add configurable identity address type and io capabilities

### DIFF
--- a/bumble/smp.py
+++ b/bumble/smp.py
@@ -1078,11 +1078,19 @@ class Session:
         )
 
     def send_identity_address_command(self) -> None:
-        identity_address = {
-            None: self.manager.device.static_address,
-            Address.PUBLIC_DEVICE_ADDRESS: self.manager.device.public_address,
-            Address.RANDOM_DEVICE_ADDRESS: self.manager.device.static_address,
-        }[self.pairing_config.identity_address_type]
+        if self.pairing_config.identity_address_type == Address.PUBLIC_DEVICE_ADDRESS:
+            identity_address = self.manager.device.public_address
+        elif self.pairing_config.identity_address_type == Address.RANDOM_DEVICE_ADDRESS:
+            identity_address = self.manager.device.static_address
+        else:
+            # No identity address type set. If the controller has a public address, it
+            # will be more responsible to be the identity address.
+            if self.manager.device.public_address != Address.ANY:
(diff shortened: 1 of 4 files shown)


bumble/smp.py
```diff
@@ -1079,7 +1079,7 @@ def send_pairing_dhkey_check_command(self) -> None:
 
     def send_identity_address_command(self) -> None:
         identity_address = {
-            None: self.manager.device.static_address,
+            None: self.connection.self_address,
```
barbibulle, line 1082: It doesn't seem right to send an identity address that's not the device's static random or public address.
The spec, BLUETOOTH CORE SPECIFICATION Version 5.4 | Vol 6, Part B Paragraph 1.3 does say "A device's Identity Address is a Public Device Address or Random Static Device Address that it uses in packets it transmits. If a device is using Resolvable Private Addresses, it shall also have an Identity Address.",
so if the connection address is an RPA, that can't be used as an identity address (and the peer would also get confused because they would then be resolving further RPAs from this device as resolving to that connection address.
Can you explain in which way you think this breaks CTKD?

zxzxwu: Consider the following enviroment:
- Bumble has Classic and CTKD over Classic enabled
- Bumble doesn't set preferred identity address type

In this case, Bumble will send its Static Random address as identity address, which is very confusing to remote device - but scenario commonly exists in most of Bumble's Classic-only examples.

One thing very difficult to handle is that Bumble allows device to have public and random static address simultaneously, so it's hard to tell which one should be considered as identity address unless user always sets the perferred identity address.

barbibulle: I see. So maybe the right thing here is that if the user doesn't choose static or public identity in the pairing config (default), then the choice should be:
  * the public address if it is set (not == ANY)
  * else the static random address

If the user wants the static random address even if the controller has a public address, then they should set it explicitly in the PairingConfig.

zxzxwu: Make sense, but not sure what's the impact to existing users.

zxzxwu: Looks good to them.

barbibulle (approved):

PR #529, opened 2024-08-10
basic broadcast assistant functionality

barbibulle:
Adds some basic broadcast assistant support to the auracast.py app.
This is still very basic, and might not completely work yet, but since it includes a small reorganization of some of the profile implementation files, I'd like to merge this soon so that the branch doesn't end up with conflicts later on if changes are made to the reorganized profiles.


bumble/profiles/ascs.py
```diff
@@ -0,0 +1,739 @@
...
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for
+
+"""LE Audio - Audio Stream Control Service"""
```
zxzxwu, line 14: If we are going to split Basic Audio Profile into different services, I think the directory name `profiles` will become not so precise? Though many other services have been there for some time.

bumble/profiles/tmap.py
```diff
@@ -0,0 +1,87 @@
...
+            GATT_TMAP_ROLE_CHARACTERISTIC
+        ):
+            self.role = DelegatedCharacteristicAdapter(
+                characteristics[0],
+                decode=lambda value: Role(
+                    struct.unpack_from('<H', value, 0)[0],
+                ),
+            )
+        else:
+            self.role = None
```
zxzxwu, line 87: Would it be better to raise an error here? (and make `role` not optional)

bumble/profiles/ascs.py
barbibulle: It's Ok for the `profiles` directory to contain files that represent entire profiles (ex: tmap.py or bap.py), just one service (ex: bass.py, battery_service.py) or even just multi-profile shared definitions (ex: le_audio.py). Some profiles share elements, so it's convenient to break those out, and some profiles are large, so splitting them into individual services also makes sense when needed. So the `profiles` directory is just a container for "implementation and definition files associated with different profiles".

bumble/profiles/tmap.py
barbibulle: done.

zxzxwu (approved):

PR #531, opened 2024-08-12
BtBench: Add Scan functionality

AlanRosenthal:
invoked via `adb shell am start -n com.github.google.bumble.btbench/.MainActivity --es autostart scan`


extras/android/BtBench/app/src/main/java/com/github/google/bumble/btbench/MainActivity.kt
```diff
@@ -166,6 +167,7 @@ class MainActivity : ComponentActivity() {
                 "rfcomm-server" -> runRfcommServer()
                 "l2cap-client" -> runL2capClient()
                 "l2cap-server" -> runL2capServer()
+                "scan" -> runScan()
```
barbibulle, line 170: It would be good to split this into start-scan and stop-scan, so that you can stop the scan without killing the app.

barbibulle (approved):

PR #532, opened 2024-08-12
Implement Hap support

wescande:
Hearing Access Service 1.0 implementation


bumble/profiles/hap.py
```diff
@@ -0,0 +1,556 @@
...
+        preset_record: PresetRecord
+
+        def __init__(self, idx, preset):
+            self.prev_index = idx
+            self.preset_record = preset
+
+        def to_bytes(self) -> bytes:
+            return bytes([self.prev_index]) + self.preset_record.to_bytes()
+
+    def __init__(self, id: ChangeId, params: Generic):
```
zxzxwu, line 114: Similarly, `PresetChangedOperation` could be a dataclass as well

bumble/profiles/hap.py
```diff
@@ -0,0 +1,556 @@
...
+        self.change_id = id
+        self.additional_parameters = params
+
+    change_id: ChangeId
+    additional_parameters: Generic | int
```
zxzxwu, line 119: As Bumble still need to support Python 3.8+, we need to use `Union[Generic, int]`

bumble/profiles/hap.py
```diff
@@ -0,0 +1,556 @@
...
+                [
+                    HearingAidPresetControlPointOpcode.PRESET_CHANGED,
+                    self.change_id,
+                    is_last,
+                ]
+            )
+            + self.additional_parameters_to_bytes()
+        )
+
+    @staticmethod
```
zxzxwu, line 139: classmethod is preferred over staticmethod

bumble/profiles/hap.py
```diff
@@ -0,0 +1,556 @@
...
+            CAN_BE_WRITTEN = 0b1
+
+        class IsAvailable(OpenIntEnum):
+            IS_UNAVAILABLE = 0b0
+            IS_AVAILABLE = 0b1
+
+        writable: Writable
+        is_available: IsAvailable
+
+        def to_bytes(self) -> bytes:
```
zxzxwu, line 179: Could be `__bytes__`, which will be used when `bytes(obj)` is invoked.

bumble/profiles/hap.py
```diff
@@ -0,0 +1,556 @@
...
+        await self.set_next_preset(connection, False)
+        # TODO (low priority) inform other server of the change
+
+    async def _on_set_previous_preset_synchronized_locally(
+        self, connection: Optional[device.Connection]
+    ):
+        assert (
+            self.hearing_aid_features.preset_synchronization_support
+            == PresetSynchronizationSupport.PRESET_SYNCHRONIZATION_IS_SUPPORTED
+        )
```
zxzxwu, line 535: Prefer to raise specific error instead of assertion (unless it's a non-null assertion which is just for type checking).

bumble/profiles/hap.py
wescande: classmethod is taking a class type as parameter that will be unused

Why should we prefer it over static ?

zxzxwu: https://google.github.io/styleguide/pyguide.html#217-function-and-method-decorators

wescande: Ok, sounds like I should avoid classmethod too.
I inlined it in the only place that was using it as this is just used to sort once and doesn't really need to be expose at module level

bumble/profiles/hap.py
```diff
@@ -0,0 +1,590 @@
...
+            @watcher.on(
+                device, 'connection'
+            )  # should it be on_characteristic_subscription ?
+            async def on_connection(connection: Connection) -> None:
+                # TODO if device is bonded && device is HAP ?
+                with contextlib.closing(EventWatcher()) as watcher:
+
+                    @watcher.on(connection, 'disconnection')
+                    def on_disconnection() -> None:
+                        self.currently_connected_clients.remove(connection)
```
wescande, line 250: I haven't see a lot of usage of watcher.  
Is this a correct usage of it ?

Should the service listen to connection ? or should it listen to both `on_connection_encryption_change` to know the bonded device and at the same time to `on_characteristic_subscription` to know the device that want to be notified with the hearing_aid_preset_control_point

zxzxwu: Watcher is a context helper to release listeners after some one-shot usage. Right after exiting the context, listeners will be all released. It seems your listeners need to work on a longer lifecycle so I guess they should be native listeners instead of context-managed listeners?

bumble/profiles/hap.py
```diff
@@ -0,0 +1,590 @@
...
+        self.hearing_aid_features = features
+        for p in presets:
+            self.preset_records[p.index] = p
+
+        with contextlib.closing(EventWatcher()) as watcher:
+
+            @watcher.on(
+                device, 'connection'
+            )  # should it be on_characteristic_subscription ?
+            async def on_connection(connection: Connection) -> None:
```
zxzxwu, line 244: Currently we don't introduce `AsyncIOEventEmitter` so I guess this won't work?

wescande: You mean the method the watcher call cannot be a async method ?

bumble/profiles/hap.py
wescande: You mean using calling watcher_on directly and not using the annotation ?

bumble/profiles/hap.py
```diff
@@ -0,0 +1,590 @@
...
+                    == None
+                ):
+                    self.preset_changed_operations_history_per_device[
+                        connection.device.public_address
+                    ] = []
+                    return
+                # Send all the PresetChangedOperation that occur when not connected
+                await self._preset_changed_operation(connection)
+                # Update the active preset index if needed
+                await self.notify_active_preset_for_connection(connection)
```
zxzxwu, line 266:
```suggestion
        @device.on('connection')
        def on_connection(connection: Connection) -> None:
            @connection.on('disconnection')
            def on_disconnection(_reason):
                self.currently_connected_clients.remove(connection)

            self.currently_connected_clients.add(connection)
            if connection.device.public_address not in self.preset_changed_operations_history_per_device:
                self.preset_changed_operations_history_per_device[
                    connection.device.public_address
                ] = []
                return
        
            async def on_connection_async():
                # Send all the PresetChangedOperation that occur when not connected
                await self._preset_changed_operation(connection)
                # Update the active preset index if needed
                await self.notify_active_preset_for_connection(connection)
            connection.abort_on('disconnection', on_connection_async())
```

bumble/profiles/hap.py
zxzxwu: I provide a way in the suggestion.

bumble/profiles/hap.py
```diff
@@ -0,0 +1,590 @@
...
+    # TODO this need to be triggered when device is unbonded
+    def on_forget(self, addr: Address):
+        self.preset_changed_operations_history_per_device.pop(addr)
+
+    def _on_write_hearing_aid_preset_control_point(
+        self, connection: Optional[Connection], value: bytes
+    ):
+        opcode = HearingAidPresetControlPointOpcode(value[0])
+        handler = getattr(self, '_on_' + opcode.name.lower())
+        handler(connection, value)
```
zxzxwu, line 312: Since your handlers are asynchrounous, you need to use them with abort_on, otherwise they won't work.
```suggestion
        connection.abort_on('disconnection', handler(connection, value))
```

bumble/profiles/hap.py
zxzxwu: Please check the suggestion. Also I suggest you to write an example like https://github.com/google/bumble/blob/main/examples/run_unicast_server.py and make it work with Android. It should be helpful to understand how asyncio and event emitter work.

wescande:
In order to check if this is working, I am currently trying to implement an HAP test in avatar
See https://android-review.git.corp.google.com/q/topic:%22hap_for_charlie%22 that is still work in progress as of today.

I am struggling with having a connection as Android require LeAudio to activate Hap :/

zxzxwu:
> In order to check if this is working, I am currently trying to implement an HAP test in avatar See https://android-review.git.corp.google.com/q/topic:%22hap_for_charlie%22 that is still work in progress as of today.
> 
> I am struggling with having a connection as Android require LeAudio to activate Hap :/

wescande:
I have added a run_hap_server but I don't really understand yet how you are making use of these script.

bumble/profiles/hap.py
```diff
@@ -0,0 +1,599 @@
...
+    ] = {}
+
+    # Keep an updated list of connected client to send notification to
+    currently_connected_clients: Set[Connection] = set()
+
+    def __init__(
+        self, device: Device, features: HearingAidFeatures, presets: List[PresetRecord]
+    ) -> None:
+        # 2.5. GATT sub-procedure requirements
+        assert att.ATT_DEFAULT_MTU >= 49
```
zxzxwu, line 244: Though Python doesn't forbid overwrite constants, this does violate the convention. Is it possible to check the MTU in use from `device`, instead of checking and chaning the constant?

bumble/profiles/hap.py
```diff
@@ -0,0 +1,599 @@
...
+
+        self.device = device
+        self.server_features = features
+        for p in presets:
+            assert len(p.name.encode()) >= 1
+            assert len(p.name.encode()) <= 40
```
zxzxwu, line 250: Prefer to raise core.InvalidArgumentError

bumble/profiles/hap.py
```diff
@@ -0,0 +1,599 @@
...
+            self.preset_records[p.index] = p
+
+        @device.on('connection')
+        def on_connection(connection: Connection) -> None:
+            @connection.on('disconnection')
+            def on_disconnection(_reason):
+                self.currently_connected_clients.remove(connection)
+            # TODO Should we filter on device bonded && device is HAP ?
+            self.currently_connected_clients.add(connection)
+            if connection.device.public_address not in self.preset_changed_operations_history_per_device:
```
zxzxwu, line 260: Is the address must be public?

bumble/profiles/hap.py
```diff
@@ -0,0 +1,599 @@
...
+        del presets[num_presets:]
+        if len(presets) == 0:
+            raise att.ATT_Error(CommonErrorCode.OUT_OF_RANGE)
+
+        AsyncRunner.spawn(self._read_preset_response(connection, presets))
+
+    async def _read_preset_response(
+        self, connection: Connection, presets: List[PresetRecord]
+    ):
+        try:
```
zxzxwu, line 354: Is there any error raised here?

bumble/profiles/hap.py
```diff
@@ -0,0 +1,599 @@
...
+                    self.hearing_aid_preset_control_point,
+                    value=bytes(
+                        [
+                            HearingAidPresetControlPointOpcode.READ_PRESET_RESPONSE,
+                            len(presets) == 1,
+                        ]
+                    )
+                    + bytes(presets[0]),
+                )
+                presets.pop(0)
```
zxzxwu, line 367: nit: In Python I will prefer for-loop over while-loop

```suggestion
            for i, preset in enumerate(presets):
                await connection.device.indicate_subscriber(
                    connection,
                    self.hearing_aid_preset_control_point,
                    value=bytes(
                        [
                            HearingAidPresetControlPointOpcode.READ_PRESET_RESPONSE,
                            i == len(presets) - 1,
                        ]
                    )
                    + bytes(preset),
                )
              presets.clear()
```

bumble/profiles/hap.py
```diff
@@ -0,0 +1,599 @@
...
+            self.read_presets_request_in_progress = False
+
+    async def generic_update(self, op: PresetChangedOperation):
+        '''Server API to perform a generic update. It is the responsibility of the caller to modify the preset_records to match the PresetChangedOperation being sent'''
+        await self._notifyPresetOperations(op)
+
+    async def delete_preset(self, index: int):
+        '''Server API to delete a preset. It should not be the current active preset'''
+
+        assert index != self.active_preset_index
```
zxzxwu, line 379: Prefer to raise specific Bumble error

bumble/profiles/hap.py
```diff
@@ -0,0 +1,599 @@
...
+        await connection.device.notify_subscriber(
+            connection,
+            self.active_preset_index_characteristic,
+            value=bytes([self.active_preset_index]),
+        )
+        self.active_preset_index_per_device[
+            connection.device.public_address
+        ] = self.active_preset_index
+
+    async def notify_active_preset(self):
```
zxzxwu, line 482: Always provide a return type
```suggestion
    async def notify_active_preset(self) -> None:
```

zxzxwu (commented):
Could you also provide unit tests? Thanks!

bumble/profiles/hap.py
wescande: Calling pop for each indicate allow to remove the notification that has been sent only.

If we do the for loop and call clear, we are not handling the case in which the remove device doesn't received an indicate.
In such scenario, the service should try to resend the indication when it reconnect later

bumble/profiles/hap.py
wescande: procedure can be interrupt if the remote device doesn't receive the indicate

bumble/profiles/hap.py
wescande: I do not know
HAS is only involving bonded device, so I believe public_address is the one we want to use ?

bumble/profiles/hap.py
wescande: I am not sure how.
I intended this check to be a "friendly warning" to the developer that forgot to negotiate it.
Because it will works without setting the MTU, but some packet will be silently truncated.

If we check the device mtu, we need to have an already negotiated MTU when creating the service, which is not the case.
The service is created, then added to the advertisement of the device then the connection is established and then the MTU is negotiated.

I was hoping that by setting the ATT_DEFAULT_MTU to another value I wouldn't need a manual negotiation but it appear that the MTU negotiation is still needed.

I could simply remove the assert and let other find out as I don't expect this issue to be HAS specific

wescande:
> Could you also provide unit tests? Thanks!

There is no client implemented yet. I am not sure how to test it without and I haven't plan implementing the client part for now

bumble/profiles/hap.py
wescande: I have added documentation to explain the behavior

bumble/profiles/hap.py
wescande: I misunderstood your comment, thinking it was attached to _preset_changed_operation.

I don't think using a for loop is good for the _preset_changed_operation, but it is definitively possible for the read preset one

bumble/profiles/hap.py
zxzxwu: I will suggest to do this on the caller side, and inside HAS we only check and pop warnings?

bumble/profiles/hap.py
zxzxwu: Device might distribute random static address as its identity address, an then the statement here will not be correct.
I think we may have a logic similar to SMP https://github.com/google/bumble/blob/main/bumble/smp.py#L1080 to properly choose a proper identity address. (We may make it a method provided by device.py when Gilles is back)

zxzxwu:
> > Could you also provide unit tests? Thanks!
> 
> There is no client implemented yet. I am not sure how to test it without and I haven't plan implementing the client part for now

We can have a very brief client just for testing like VCP:
* https://github.com/google/bumble/blob/main/bumble/profiles/vcp.py
* https://github.com/google/bumble/blob/main/tests/vcp_test.py

bumble/profiles/hap.py
zxzxwu: BTW, it seems like you are always putting `connection.device.public_address` which is the address of local device (not remote), is that really what you want? The address of remote device is provided by `connection.peer_address`

wescande: I do want the remote device address.
I have updated the code to use `peer_address`

HAS is only exposing service that require encryption, I understand that mean the device have to be bonded and thus the peer_address would be correct.
Let me know if I am wrong.

bumble/profiles/hap.py
```diff
@@ -0,0 +1,657 @@
...
+            self.active_preset_index = next_preset.index
+        else:
+            self.active_preset_index = first_preset.index
+        logging.error(f'index now is {self.active_preset_index}')
+        await self.notify_active_preset()
+
+    async def _on_set_next_preset(self, connection: Optional[Connection], __value__: bytes) -> None:
+        await self.set_next_or_previous_preset(connection, False)
+
+    async def _on_set_previous_preset(self, connection: Optional[Connection], __value__: bytes) -> None:
```
zxzxwu, line 579: What is the purpose of using `__value__`? An I cannot find what does it mean online

zxzxwu (approved):

zxzxwu:
Approved as a milestone. I can merge it when the format/lint errors are fixed.

Still, there are still too many TODOs and loud logs. I hope in following PRs we can stabilize most of them.

PR #534, opened 2024-08-14
[Bug] Edimax BLE Dongle Fails After Teardown and Re-Instantiation

hkpeprah:
### Overview

Replacement for #497 with the logic for the HCI parser removed now that packets are handled without a parser for USB. This bug was observed with some RTK BLE dongles (specifically cheaper CSR dongles as well as the Edimax BT-8500). Sometimes, when a reset is attempted, the host will fail to receive a result, instead receiving an invalid packet or corrupt data. This results in blocking forever in `driver_info_for_host()`. In testing, I found that allowing that to time out, then trying again can work with dongles like the Edimax.

With cheaper CSR dongles, sometimes the first packet in response to a command after the reset will be an invalid packet. This occurs when the Local Version request is sent. The result is similar to the above in which case we wait forever on a packet that isn't coming in. In testing, I found that re-sending the command can sometimes avoid this, but not in all cases, so in the case that we do not get the local version information, we should skip over it and return upstream.

### Summary of Changes
1. In RTK `driver_info_for_host` attempt a HCI reset with a timeout; on time out, retry the HCI reset and assume the host is ready.
2. In RTK `driver_info_for_host` exit if the Local Version Information response is invalid.
3. In `bumble/host.py`, for `send_command()` allow an optional `response_timeout` to allow the caller to time out if a response is not received.
4. In `bumble/host.py`, for `on_packet()`, do not crash if `from_bytes()` fails to parse an HCI packet as may be the case with invalid data.

### DIFF
--- a/bumble/drivers/rtk.py
+++ b/bumble/drivers/rtk.py
@@ -301,6 +301,8 @@ class Driver(common.Driver):
         fw_name: str = ""
         config_name: str = ""
 
+    POST_RESET_DELAY: float = 0.2
+
     DRIVER_INFOS = [
         # 8723A
         DriverInfo(
@@ -495,12 +497,22 @@ class Driver(common.Driver):
 
     @classmethod
     async def driver_info_for_host(cls, host):
-        await host.send_command(HCI_Reset_Command(), check_result=True)
-        host.ready = True  # Needed to let the host know the controller is ready.
+        try:
+            await host.send_command(
+                HCI_Reset_Command(), check_result=True, response_timeout=cls.POST_RESET_DELAY
+            )
+            host.ready = True  # Needed to let the host know the controller is ready.
+        except asyncio.exceptions.TimeoutError:
+            logger.warning("timeout waiting for hci reset, retrying")
+            await host.send_command(HCI_Reset_Command(), check_result=True)
+            host.ready = True
 
-        response = await host.send_command(
-            HCI_Read_Local_Version_Information_Command(), check_result=True
-        )
+        command = HCI_Read_Local_Version_Information_Command()
+        response = await host.send_command(command, check_result=True)
+        if response.command_opcode != command.op_code:
+            logger.error("failed to probe local version information")
+            return None
+
         local_version = response.return_parameters
 
         logger.debug(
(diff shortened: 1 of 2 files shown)


barbibulle (approved):

barbibulle:
This PR is good to go. Can you please just fix the linter/formatter errors before we can merge it?

hkpeprah:
> This PR is good to go. Can you please just fix the linter/formatter errors before we can merge it?

Done!

barbibulle:
> > This PR is good to go. Can you please just fix the linter/formatter errors before we can merge it?
> 
> Done!

There are still some type checker errors (see the output of the GitHub actions that run on the PR, or run`invoke project.pre-commit` to run the checks locally.
Those are not type checking errors that were introduced by your change, but they are only detected now, because your change adds type annotations to the `send_command` functions (which is good, as we're adding type annotations throughout the code base as time goes), which means that now that function will be type-checked, whereas it wasn't before.
The type checker output says:
```
bumble/transport/usb.py:118: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
bumble/host.py:526: error: Incompatible types in assignment (expression has type "Future[Any]", variable has type "None")  [assignment]
bumble/host.py:531: error: Argument 1 to "wait_for" has incompatible type "None"; expected "Union[Future[Never], Generator[Any, None, Never], Awaitable[Never]]"  [arg-type]
bumble/host.py:532: error: "None" has no attribute "result"  [attr-defined]
bumble/host.py:537: error: "HCI_Command_Status_Event" has no attribute "status"  [attr-defined]
```

Those errors can be fixed easily: 
  `self.pending_response` should be explicitly typed, so that the assignment to a non-None value doesn't generate an error (probably something like ` self.pending_response: Future = None` on line 174 would work fine.

hkpeprah:
I'm not sure how to work around this one:
```
bumble/host.py:537: error: "HCI_Command_Status_Event" has no attribute "status"  [attr-defined]
```
It seems like PyLint doesn't like how the constructors for the events are done, as it can't tell the class instance has an attribute called `status`.

barbibulle:
> I'm not sure how to work around this one:
> 
> ```
> bumble/host.py:537: error: "HCI_Command_Status_Event" has no attribute "status"  [attr-defined]
> ```
> 
> It seems like MyPy doesn't like how the constructors for the events are done, as it can't tell the class instance has an attribute called `status`.

The comment `# type: ignore[attr-defined]` should solve the issue I think.

PR #541, opened 2024-08-26
Pandora: Fix l2cap receive rpc

atorodotme:


bumble/pandora/l2cap.py
```diff
@@ -61,6 +61,12 @@ def __init__(self, device: Device, config: Config) -> None:
         self.config = config
         self.sdu_queue: asyncio.Queue = asyncio.Queue()
 
+    def on_channel_sdu(self, sdu):
+        async def handle_sdu():
+            await self.sdu_queue.put(sdu)
+
+        asyncio.create_task(handle_sdu())
```
SilverBzH, line 68: Please use synchronized Queue instead:
https://docs.python.org/3/library/queue.html

```suggestion
    def on_channel_sdu(self, sdu):
        self.sdu_queue.put_nowait(sdu)
```

zxzxwu:
Closed this PR because AOSP no longer use Pandora implementation from Bumble. Someday in the future we will remove all Pandora implementation from Bumble. Avatar can pin to an older implementation to remain a basic availability.

PR #545, opened 2024-08-30
Pandora: refactor l2cap service

SilverBzH:
* Craft the PandoraChannel from the connection_handle and the source_cid
* Fix race on waitDisconnection
* Add ChannelContext to enable mutliple channels on the service

### DIFF
--- a/bumble/pandora/l2cap.py
+++ b/bumble/pandora/l2cap.py
@@ -16,10 +16,12 @@ import asyncio
 import grpc
 import json
 import logging
-import threading
+
+from asyncio import Queue as AsyncQueue, Future
 
 from . import utils
 from .config import Config
+from bumble.core import OutOfResourcesError, InvalidArgumentError
 from bumble.device import Device
 from bumble.l2cap import (
     ClassicChannel,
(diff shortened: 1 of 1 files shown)


bumble/pandora/l2cap.py
```diff
@@ -234,18 +240,18 @@ async def Connect(
         if not l2cap_channel:
             return ConnectResponse(error=COMMAND_NOT_UNDERSTOOD)
 
-        channel = self.channel_to_proto(l2cap_channel)
-        return ConnectResponse(channel=channel)
+        pandora_channel = self.craft_pandora_channel(connection_handle, l2cap_channel)
```
hchataing, line 243: Should you save the channel in `onServiceConnected` here as well ?

SilverBzH: The `connect` l2cap function is awaiting for the connection to happen. However the pandora_channel was not registered internally indeed. Just added the missing piece of code.

8lank (approved):

hchataing (approved):

zxzxwu (approved):

PR #549, opened 2024-09-07
Return ATT_Error_Response on rejected write request

zxzxwu:


zxzxwu:
cc @wescande

zxzxwu:
I think it makes more sense if GATT client also raises ATT_Error instead of ProtocolError?

barbibulle:
> I think it makes more sense if GATT client also raises ATT_Error instead of ProtocolError?

`ATT_Error` is a subclass of `ProtocolError`, with the namespace set to `att`, so it would be perfectly fine to have the client raise `ATT_Error`.

barbibulle (approved):

PR #555, opened 2024-09-16
aics: make it a secondary service

SilverBzH:
Modify tests accordingly

### DIFF
--- a/bumble/profiles/vcp.py
+++ b/bumble/profiles/vcp.py
@@ -123,7 +123,7 @@ class VolumeControlService(gatt.TemplateService):
                 self.volume_control_point,
                 self.volume_flags,
             ],
-            included_services=included_services,
+            included_services=list(included_services),
         )
 
     @property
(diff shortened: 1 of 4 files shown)


bumble/profiles/vcp.py
```diff
@@ -88,6 +88,7 @@ def __init__(
         muted: int = 0,
         change_counter: int = 0,
         volume_flags: int = 0,
+        included_services: List[gatt.Service] = [],
```
zxzxwu, line 91: Should not be list with a default value because it's mutable. (We forgot to change this for `gatt.Service` as well.

```suggestion
        included_services: Sequence[gatt.Service] = (),
```

bumble/profiles/vcp.py
```diff
@@ -117,11 +118,12 @@ def __init__(
...
 
         super().__init__(
-            [
+            characteristics=[
                 self.volume_state,
                 self.volume_control_point,
                 self.volume_flags,
-            ]
+            ],
+            included_services=included_services,
```
zxzxwu, line 126: Since we change on line#91, here we need
```suggestion
            included_services=list(included_services),
```

PR #561, opened 2024-09-27
run_csis_servers: Update `usage` and add docs entry

wpiet:
Amends the usage of the script and the example that show when executed incorrectly. 
Adds basic information to the `examples` section of the documentation.


zxzxwu (approved):

wpiet:
Hi,

I ran the ` invoke project.lint`  and it produced warnings about unused imports in the file I had changed. So I tried removing them, but after I force-pushed, the avatar check started failing. 

Then, I run `black` and it amended the print function I had changed. Hopefully all checks pass now.

zxzxwu:
> Hi,
> 
> I ran the ` invoke project.lint` and it produced warnings about unused imports in the file I had changed. So I tried removing them, but after I force-pushed, the avatar check started failing.
> 
> Then, I run `black` and it amended the print function I had changed. Hopefully all checks pass now.

Failure in Avatar should be an unrelated issue. I am asking Avatar folks to check now.

PR #564, opened 2024-09-27
Fix typo in comment

initializedd:
Fixes a typo that specifies the type of address

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -1766,9 +1766,9 @@ device_host_event_handlers: List[str] = []
 # -----------------------------------------------------------------------------
 class Device(CompositeEventEmitter):
     # Incomplete list of fields.
-    random_address: Address  # Random address that may change with RPA
-    public_address: Address  # Public address (obtained from the controller)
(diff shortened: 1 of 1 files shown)


bumble/device.py
```diff
@@ -1768,7 +1768,7 @@ class Device(CompositeEventEmitter):
     # Incomplete list of fields.
     random_address: Address  # Random address that may change with RPA
     public_address: Address  # Public address (obtained from the controller)
-    static_address: Address  # Random address that can be set but does not change
```
barbibulle, line 1771: The comment wasn't wrong. `static_address` is a Random Static address. So it is *both* Random and Static.
the part of the comment that says "does not change" is intended to convey the Static part.
Admittedly, the comment could be clearer. Maybe a way for fix is to update the comments on both line 1769 and 1771, where the 1769 would say "Random Private address ..." and 1771 would say "Random Static address ..."

initializedd: Apologies, you are correct.

I have amended the commit to clarify the comments further instead.

barbibulle (approved):

PR #567, opened 2024-09-30
avatar: reduce number of shards

SilverBzH:
Having too many shards extends the total testing time. 4 shards is a
good compromise.

Also, run all the shards and does not skip when one shard is failing


SilverBzH:
We should merge this PR only when the flakiness will be tracked down. Otherwise it will just fail at 100%.

SilverBzH:
I am closing this PR.

The test is irrelevant, it take the same times for 4 or 24 shards. I'll let it untouched.

PR #571, opened 2024-10-10
a2dp player

barbibulle:
Simple CLI app to play audio to a speaker or headphones with A2DP.
This includes support for playing SBC, AAC and Ogg/Opus files.


bumble/a2dp.py
```diff
@@ -528,13 +539,113 @@ def __str__(self) -> str:
...
+    class ChannelMode(enum.IntEnum):
+        MONO = 0
+        STEREO = 1
+        DUAL_MONO = 2
+
+    CHANNEL_MODE_BITS = {
+        ChannelMode.MONO: 1 << 0,
+        ChannelMode.STEREO: 1 << 1,
+        ChannelMode.DUAL_MONO: 1 << 2,
+    }
```
zxzxwu, line 558: I think these constants should be flag instead of enum, so that we don't need additional maps and `from_discrete_values` to build and serialize codec informations (also applicable to SBC and AAC).

```py

class OpusCodecInformation:
  """OPUS codec information."""

  class ChannelMode(enum.IntFlag):
    MONO = 0x01
    STEREO = 0x02
    DUAL_MONO = 0x04


  class FrameSize(enum.IntFlag):
    SIZE_10_MILLISECONDS = 0x08
    SIZE_20_MILLISECONDS = 0x10


  class SamplingRate(enum.IntFlag):
    RATE_48000 = 0x80

  sample_rate: SamplingRate
  channel_mode: ChannelMode
  frame_size: FrameSize

  VENDOR_ID: ClassVar[int] = 0xE0
  CODEC_ID: ClassVar[int] = 0x01

  def __bytes__(self) -> bytes:
    return struct.pack(
        '<IHB',
        self.VENDOR_ID,
        self.CODEC_ID,
        self.sample_rate | self.channel_mode | self.frame_size,
    )
```

bumble/a2dp.py
```diff
@@ -663,3 +775,323 @@ async def generate_packets():
...
+    ) -> None:
+        self.read = read
+        self.mtu = mtu
+        self.codec_capabilities = codec_capabilities
+
+    @property
+    def packets(self):
+        async def generate_packets():
+            # pylint: disable=import-outside-toplevel
+            from .avdtp import MediaPacket  # Import here to avoid a circular reference
```
zxzxwu, line 1065: Should we move MediaPacket to a third module so that we can avoid inline imports?

bumble/avdtp.py
```diff
@@ -1397,7 +1429,19 @@ def find_remote_sink_by_codec(
...
+                            ):
+                                if (
+                                    codec_capabilities.media_codec_information.vendor_id
+                                    == vendor_id
+                                    and codec_capabilities.media_codec_information.codec_id
+                                    == codec_id
+                                ):
+                                    has_codec = True
+                            else:
+                                has_codec = True
```
zxzxwu, line 1444: nit: If there isn't additional works here
```suggestion
                            has_codec = (isinstance(
                                codec_capabilities.media_codec_information,
                                VendorSpecificMediaCodecInformation,
                            ) and codec_capabilities.media_codec_information.vendor_id == vendor_id
                            and codec_capabilities.media_codec_information.codec_id == codec_id)
```

bumble/a2dp.py
```diff
@@ -663,3 +775,323 @@ async def generate_packets():
...
+                version = header[4]
+                if version != 0:
+                    raise ValueError(f"version {version} not supported")
+
+                header_type = self.HeaderType(header[5])
+                (granule_position,) = struct.unpack_from("<Q", header, 6)
+                (bitstream_serial_number,) = struct.unpack_from("<I", header, 14)
+                (page_sequence_number,) = struct.unpack_from("<I", header, 18)
+                (crc_checksum,) = struct.unpack_from("<I", header, 22)
+                page_segments = header[26]
```
zxzxwu, line 988:
```suggestion
(granule_position, bitstream_serial_number, page_sequence_number, crc_checksum, page_segments) = struct.unpack_from("<QIIIB", header, 6)
```

bumble/avdtp.py
```diff
@@ -51,6 +51,7 @@
     A2DP_MPEG_2_4_AAC_CODEC_TYPE,
     A2DP_NON_A2DP_CODEC_TYPE,
     A2DP_SBC_CODEC_TYPE,
+    A2DP_VENDOR_MEDIA_CODEC_INFORMATION_CLASSES,
```
zxzxwu, line 54: (Just a comment)

From a spec perspective, importing A2DP from AVDTP seems like a reversed dependencies, but top-down parsing is important for usage. I am thinking whether it's possible to hook parsing table from A2DP, but that will require callers to always import A2DP, or we will need to wrap AVDTP and A2DP stubs inside a public module which is too aggresive for now.

bumble/a2dp.py
```diff
@@ -130,6 +136,11 @@
     MPEG_4_AAC_SCALABLE_OBJECT_TYPE: 'MPEG_4_AAC_SCALABLE_OBJECT_TYPE'
 }
 
+
+OPUS_VENDOR_ID = 0x000000E0
+OPUS_CODEC_ID  = 0x0001
+OPUS_MAX_FRAMES_IN_RTP_PAYLOAD = 15
```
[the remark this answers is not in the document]
barbibulle: Done.

bumble/a2dp.py
barbibulle: Yes, I think that's much better. I've done that for all 3 types now.
Initially, I thought it would be convenient to treat some fields like sampling frequencies as ints instead of enums, but that just complicates everything for no good reason. Using `IntFlag` for everything is the right approach.

bumble/a2dp.py
barbibulle: Done. `MediaPacket` is not in `rtp.py`

bumble/a2dp.py
barbibulle: Done

bumble/avdtp.py
barbibulle: Yes, it would be nice to have some better way to have a sort of dependency injection here. I haven't found a great way to do that that's not intrusive. I'll try to think of some better way.

bumble/avdtp.py
barbibulle: That's not 100% the same: if we encounter a capabilities for a standard (non-vendor) codec, we want `has_codec` to be `True`.

bumble/a2dp.py
zxzxwu: This one is left

zxzxwu (approved):

PR #578, opened 2024-10-24
Add missing parameter 'flags' to ATT_Execute_Write_Request PDU

jmdietrich-gcx:
Bluetooth spec @ Vol 3, Part F - 3.4.6.3 Table 3.36 shows, that the ATT_EXECUTE_WRITE_REQ PDU contains the parameter 'Flags' with size 1 octet, which allows to cancel all prepared writes (0x00) or to immediately write all pending prepared values (0x01).

### DIFF
--- a/bumble/att.py
+++ b/bumble/att.py
@@ -710,7 +710,7 @@ class ATT_Prepare_Write_Response(ATT_PDU):
 
 
 # -----------------------------------------------------------------------------
-@ATT_PDU.subclass([])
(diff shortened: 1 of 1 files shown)


barbibulle (approved):

barbibulle:
Thanks for pointing this out. The timing is great, I had started working on a small PR to support long/queued writes, which will make use of that GATT operation. I'll have a PR ready for review soon.

PR #583, opened 2024-11-04
regression test for GATT unsubscription

barbibulle:
A GATT unsubscription bug was fixed recently. This adds a simple regression test for what was broken.

### DIFF
--- a/tests/gatt_test.py
+++ b/tests/gatt_test.py
@@ -851,7 +851,12 @@ async def test_unsubscribe():
     await async_barrier()
     mock1.assert_called_once_with(ANY, True, False)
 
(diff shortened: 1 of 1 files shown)


tests/gatt_test.py
```diff
@@ -851,7 +851,12 @@ async def test_unsubscribe():
     await async_barrier()
     mock1.assert_called_once_with(ANY, True, False)
 
-    await c2.subscribe()
+    assert len(server.gatt_server.subscribers) == 1
+
+    def callback(_):
```
zxzxwu, line 856: nit: maybe a mock function is better here

zxzxwu (approved):

PR #585, opened 2024-11-08
Add `Volume Offset Control Service`

wpiet:
Adds initial support for VOCS.


bumble/profiles/vocs.py
```diff
@@ -0,0 +1,240 @@
...
+# -----------------------------------------------------------------------------
+
+MIN_VOLUME_OFFSET = -255
+MAX_VOLUME_OFFSET = 255
+CHANGE_COUNTER_MAX_VALUE = 0xFF
+
+SET_VOLUME_OFFSET_OPCODE = 0x01
+
+
+class ErrorCode(enum.IntEnum):
```
barbibulle, line 42: It would be better to use `utils.OpenIntEnum` here so that the code doesn't raise an exception when encountering new/unknown error codes.

bumble/profiles/vocs.py
```diff
@@ -0,0 +1,240 @@
...
+
+    async def _on_write_audio_location(
+        self, connection: Optional[device.Connection], value: bytes
+    ) -> None:
+        self.audio_location_bytes = value
+        await connection.device.notify_subscribers(
+            attribute=self.audio_channel_location,
+            value=self.audio_location_bytes,
+        )
+        self.emit('audio_location', self.audio_location)
```
barbibulle, line 169: Typically, it is preferable to not repeat in the event the changed state if that state already exists as a property of the emitter.
Here, for instance (the same idea applies to the other placed in this PR where events are emitted), a listener can access `audio_location` from the emitter, so there's no need to include that in the event.
a listener would do:
`service.on('audio_location_change', ...)`
and in the event handler, retrieve that actual values as `service.audio_location`.

wpiet: Thank you for the explanation. I removed these unnecessary properties from the events.

bumble/profiles/vocs.py
```diff
@@ -0,0 +1,244 @@
...
+        self.service_proxy = service_proxy
+
+        self.volume_offset_state = gatt.PackedCharacteristicAdapter(
+            service_proxy.get_characteristics_by_uuid(
+                gatt.GATT_VOLUME_OFFSET_STATE_CHARACTERISTIC
+            )[0],
+            '<hB',
+        )
+
+        self.audio_location = gatt.PackedCharacteristicAdapter(
```
barbibulle, line 231: `audio_location` values are enums. We have an enum type for is already, in `bap.AudioLocation`. This could use a `DelegatedCharacteristicAdapter` to encode/decode to/from that type.

bumble/profiles/vocs.py
```diff
@@ -0,0 +1,244 @@
...
+# -----------------------------------------------------------------------------
+class VolumeOffsetControlServiceProxy(gatt_client.ProfileServiceProxy):
+    SERVICE_CLASS = VolumeOffsetControlService
+
+    volume_offset_control_point: gatt_client.CharacteristicProxy
+
+    def __init__(self, service_proxy: gatt_client.ServiceProxy) -> None:
+        self.service_proxy = service_proxy
+
+        self.volume_offset_state = gatt.PackedCharacteristicAdapter(
```
barbibulle, line 224: It would be nice to create a `VolumeOffsetState` `dataclass` with a `volume_offset` and a `change_counter` property, this would be more user-friendly than an `(int, int)` tuple.

bumble/profiles/vocs.py
```diff
@@ -0,0 +1,244 @@
...
+                read=self._on_read_audio_output_description,
+                write=self._on_write_audio_output_description,
+            ),
+        )
+        characteristics.append(self.audio_output_description)
+
+        super().__init__(characteristics=characteristics, primary=False)
+
+    @property
+    def volume_offset_state_bytes(self) -> bytes:
```
barbibulle, line 136: (this comment also applies to the other characteristics in this service)
Instead of having customized bytes serialization methods here, you should have shared adapters, used both for the service and service proxy. 
Look at the `heart_rate_service.py` for an example of how this may be done.

bumble/profiles/vocs.py
wpiet: In the current version of the service I was not sure how to handle events, so I removed them completely. If they are necessary, could you please give me some example?

bumble/profiles/vocs.py
wpiet: Now I used these adapters in conjunction with `from_bytes` method which handles the enum type conversion. Is such usage correct?

bumble/profiles/vocs.py
wpiet: I simplified the service implementation and added standalone classes for each characteristic with shared methods as you suggested.

bumble/profiles/vocs.py
```diff
@@ -0,0 +1,328 @@
...
+
+# -----------------------------------------------------------------------------
+# Constants
+# -----------------------------------------------------------------------------
+
+MIN_VOLUME_OFFSET = -255
+MAX_VOLUME_OFFSET = 255
+CHANGE_COUNTER_MAX_VALUE = 0xFF
+
+SET_VOLUME_OFFSET_OPCODE = 0x01
```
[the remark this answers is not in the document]
wpiet: Okay, I changed that to an enum.

barbibulle (approved):

PR #587, opened 2024-11-12
Replace HCI member imports in device.py

zxzxwu:


zxzxwu:
websockets 14.0 deprecates many APIs used now. @barbibulle Shall we move to 14.0 now, or stay at 13.1 and migrate later?

barbibulle (approved):

barbibulle:
> websockets 14.0 deprecates many APIs used now. @barbibulle Shall we move to 14.0 now, or stay at 13.1 and migrate later?

Since we don't use too many WebSockets features, it's probably Ok to stay on 13.1 for now, until we need something that only exists in >= 14.0, or if we need to update other packages that would result in a conflict with the `websockets` requirements.

PR #591, opened 2024-11-15
Improve Broadcast Scanning

zxzxwu:
* APP: Allow scanning for broadcast without name (check annoucement instead)
* Add BAP dataclass serializer
* Allow nullable codec specific configuration fields
* Remove BAP CodecID (use HCI CodecID/Format instead, though the term is a little bit confusing)

### DIFF
--- a/apps/auracast.py
+++ b/apps/auracast.py
@@ -60,7 +60,7 @@ AURACAST_DEFAULT_ATT_MTU = 256
 class BroadcastScanner(pyee.EventEmitter):
     @dataclasses.dataclass
     class Broadcast(pyee.EventEmitter):
-        name: str
+        name: str | None
         sync: bumble.device.PeriodicAdvertisingSync
         rssi: int = 0
         public_broadcast_announcement: Optional[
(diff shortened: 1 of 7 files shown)


zxzxwu:
I use 3.10 typing annotations - with __annotations__, Python doesn't parse these annotations in runtime, and mypy works well with them in 3.9.

apps/auracast.py
```diff
@@ -274,12 +274,24 @@ async def stop(self) -> None:
...
+                for ad in ads
+                if isinstance(ad, tuple)
+                and ad[0] == bumble.gatt.GATT_BROADCAST_AUDIO_ANNOUNCEMENT_SERVICE
+            )
+        ):
             return
+
+        broadcast_name = (
+            advertisement.data.get(bumble.core.AdvertisingData.BROADCAST_NAME)
+            or 'Unknown'
```
barbibulle, line 293: there's a (small) risk of conflict here. If a real broadcast service uses the name 'Unknown', we don't know the difference. Maybe the best would be to allow passing `None` for the `broadcast_name` parameter of `self.on_new_broadcast`?

barbibulle (approved):

apps/auracast.py
zxzxwu: Done. Skip name when it's None here.

PR #597, opened 2024-11-23
intel hci

barbibulle:
Support for Intel USB controllers.
Tested with BE200 and AX210
NOTE: this driver allows extending the device config parameters via transport name parameters (the previous version had some config parameters hard-coded, which isn't optimal for all users). So users who want to have the same config as before will now have to pass the config extension string when creating the transport.


tools/intel_util.py
```diff
@@ -0,0 +1,154 @@
...
+
+        # Reboot in bootloader mode
+        await driver.load_firmware()
+
+        # Get and print the device info
+        print_device_info(await driver.read_device_info())
+
+
+# -----------------------------------------------------------------------------
+async def do_bootloader(usb_transport: str, force: bool) -> None:
```
SilverBzH, line 93: This command does not work on Linux. I think it's more a transport issue than a driver one

bumble/drivers/intel.py
```diff
@@ -85,18 +389,283 @@ def check(host):
...
+                ]
+            ),
+        )
+
+        # Check if the firmware is already loaded.
+        if (
+            device_info.get(ValueType.CURRENT_MODE_OF_OPERATION)
+            == ModeOfOperation.OPERATIONAL
+        ):
+            logger.debug("firmware already loaded")
```
SilverBzH, line 463: Should we be able to force this ? If we want to override the current firmware

SilverBzH (approved):

tools/intel_util.py
barbibulle: What error do you get? I'll try to find a linux host I can attach my USB device to and check on my end.

bumble/drivers/intel.py
barbibulle: I prefer to leave this method simple(r), and let it up to the caller to decide if they want to reset to bootloader and re-load in case there's already a FW image loaded (which they can get with `read_device_info`)

tools/intel_util.py
SilverBzH: I'll open a bug with the details.

PR #606, opened 2024-11-29
Add `Gaming Audio Profile`

wpiet:
Adds initial support for `Gaming Audio Service`.


bumble/profiles/gmap.py
```diff
@@ -0,0 +1,193 @@
...
+
+        self.gmap_role = Characteristic(
+            uuid=GATT_GMAP_ROLE_CHARACTERISTIC,
+            properties=Characteristic.Properties.READ,
+            permissions=Characteristic.Permissions.READABLE,
+            value=struct.pack('<B', gmap_role),
+        )
+        characteristics.append(self.gmap_role)
+
+        if ugg_features:
```
barbibulle, line 104: This should check for `ugg_features is not None`, because you could pass a 0-valued `ugg_features` parameter, which would evaluate to `False` in the test, but you still want a characteristic to exist.
The same applies to the other characteristics below.

bumble/profiles/gmap.py
```diff
@@ -0,0 +1,193 @@
...
+            self.ugg_features = Characteristic(
+                uuid=GATT_UGG_FEATURES_CHARACTERISTIC,
+                properties=Characteristic.Properties.READ,
+                permissions=Characteristic.Permissions.READABLE,
+                value=struct.pack('<B', ugg_features),
```
barbibulle, line 109: nit: no need for `<` in the struct pattern, since there's no byte order for just one byte.

bumble/profiles/gmap.py
```diff
@@ -0,0 +1,193 @@
...
+        self.gmap_role = Characteristic(
+            uuid=GATT_GMAP_ROLE_CHARACTERISTIC,
+            properties=Characteristic.Properties.READ,
+            permissions=Characteristic.Permissions.READABLE,
+            value=struct.pack('B', gmap_role),
+        )
+        characteristics.append(self.gmap_role)
+
+        if ugg_features is not None:
+            self.ugg_features = Characteristic(
```
zxzxwu, line 105: You still need to set a value when parameters are None, otherwise the attribute will be missing.

wpiet: Would it be more robust if we checked the role before adding conditional characteristics?

So in the line no. 104 we would have:
```
        if gmap_role & GmapRole.UNICAST_GAME_GATEWAY:
            self.ugg_features = Characteristic(
                uuid=GATT_UGG_FEATURES_CHARACTERISTIC,
                properties=Characteristic.Properties.READ,
                permissions=Characteristic.Permissions.READABLE,
                value=struct.pack('B', ugg_features),
            )
            characteristics.append(self.ugg_features)
```

and so on.

zxzxwu: Correct. It's not really related to Bluetooth, but from a readability perspective, you should always set a value for any attribute, especially you when annotate them as `Optional`, people will expect there is such an attribute even they are `None`.
Consider if you are using Java, such a pattern will probably be warned that you may access a member without initialization.

rymanluk: Hi @zxzxwu , that would mean that we will have characteristics exposed which are not really needed.

Also from the testing point of view, I see value in having an option to create invalid test device which is missing some characteristics to verify proper error handling on the DUT. 

Does it makes sense?

We also had an idea  that some of the characteristic could be created as per spec (depends on the role) but in fact as it is now gives good test flexibility.

Other options would be to extend this constructor with a bool's which would tell if given characteristic should be created but this seems to be overkill.

zxzxwu: You don't need to create a characteristic. What is needed here is
```suggestion
            self.ugg_features = Characteristic(
            ...
          else:
            self.ugg_features = None
```

Or set it like
```py
class GamingAudioService(TemplateService):
    UUID = GATT_GAMING_AUDIO_SERVICE

    gmap_role: Characteristic
    ugg_features: Optional[Characteristic] = None
    ugt_features: Optional[Characteristic] = None
    bgs_features: Optional[Characteristic] = None
    bgr_features: Optional[Characteristic] = None
```

wpiet: Thank you, now it is clear. I added the latter suggestion.

barbibulle (approved):

PR #608, opened 2024-12-05
add startupDelay and connectionPriority params to BtBench snippets

barbibulle:
These were supported by the UI, but not yet by the mobly snippet API. See examples.

### DIFF
--- a/extras/android/BtBench/app/src/main/java/com/github/google/bumble/btbench/Advertiser.kt
+++ b/extras/android/BtBench/app/src/main/java/com/github/google/bumble/btbench/Advertiser.kt
@@ -1,0 +1,39 @@
+package com.github.google.bumble.btbench
+
+import android.annotation.SuppressLint
+import android.bluetooth.BluetoothAdapter
+import android.bluetooth.le.AdvertiseCallback
+import android.bluetooth.le.AdvertiseData
+import android.bluetooth.le.AdvertiseSettings
+import android.bluetooth.le.AdvertiseSettings.ADVERTISE_MODE_LOW_LATENCY
+import android.os.Build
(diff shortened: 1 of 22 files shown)


extras/android/BtBench/app/src/main/java/com/github/google/bumble/btbench/Advertiser.kt
```diff
@@ -0,0 +1,39 @@
...
+import android.os.Build
+import java.util.logging.Logger
+
+private val Log = Logger.getLogger("btbench.advertiser")
+
+class Advertiser(private val bluetoothAdapter: BluetoothAdapter) : AdvertiseCallback() {
+    @SuppressLint("MissingPermission")
+    fun start() {
+        val advertiseSettingsBuilder = AdvertiseSettings.Builder()
+            .setAdvertiseMode(ADVERTISE_MODE_LOW_LATENCY)
```
atorodotme, line 18: If I remember correctly, we discussed being able to control the advertising mode?

atorodotme (approved):

extras/android/BtBench/app/src/main/java/com/github/google/bumble/btbench/Advertiser.kt
barbibulle: The advertising mode doesn't affect the performance during the benchmark, because we always stop advertising before starting a scenario. Do you think it is needed for some other reason?

atorodotme: Oh I see. Thank you for the clarification. No need to control it if it doesn't impact the performance

PR #616, opened 2024-12-29
Avoid cancelling futures

zxzxwu:
Cancelling a future doesn't provide enough information to its awaiter, and may leak `asyncio.CancelledError` to out of APIs.

### DIFF
--- a/bumble/core.py
+++ b/bumble/core.py
@@ -16,7 +16,7 @@
 # Imports
 # -----------------------------------------------------------------------------
 from __future__ import annotations
-import dataclasses
+import asyncio
 import enum
 import struct
 from typing import List, Optional, Tuple, Union, cast, Dict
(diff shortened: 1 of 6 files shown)


zxzxwu:
I realized that `abort_on` also raises `asyncio.CancelledError` and it's more popular...
They cannot be easily captured and overwriten inside the function, so probably we need to try and catch all code await abort_on.

zxzxwu:
A truth is that `asyncio.enture_future` is probably always same as `asyncio.create_task` when given argument is a coroutine - it never works when it's not running on a event loop. So, I think maybe we can create a new future instead of the original one, and set the result or exception there like

```py
class AbortableEventEmitter(EventEmitter):
    def abort_on(self, event: str, awaitable: Awaitable[_T]) -> asyncio.Future[_T]:
        """
        Set a coroutine or future to abort when an event occur.
        """
        real_future = asyncio.ensure_future(awaitable)
        if real_future.done():
            return real_future

        wrapped_future: asyncio.Future[_T] = asyncio.get_running_loop().create_future()

        def on_event(*_):
            if wrapped_future.done():
                return
            if isinstance(wrapped_future, asyncio.Task):
                wrapped_future.cancel()

            from bumble.core import CancelledError

            wrapped_future.set_exception(
                CancelledError(f'abort: {event} event occurred.')
            )

        def on_done(_):
            self.remove_listener(event, on_event)
            wrapped_future.set_result(real_future.result())

        self.once(event, on_event)
        real_future.add_done_callback(on_done)
        return wrapped_future
```

PR #618, opened 2025-01-06
support multiple event factories

barbibulle:
While there's no way to ensure the uniqueness of HCI vendor event codes across vendors, it is still possible to have vendor event factories that don't overlap (by design), such as the generic set of Android vendor HCI events which multiple vendors do implement (and in that case the vendors ensure that they don't re-use the same event codes for other events).
This change allows having multiple vendor factories, which will be invoked in registration order when an HCI vendor event is encountered, instead of a single factory. It is up to the caller to ensure that they do not register factories with overlapping IDs.

### DIFF
--- a/bumble/hci.py
+++ b/bumble/hci.py
@@ -5070,7 +5070,7 @@ class HCI_Event(HCI_Packet):
     hci_packet_type = HCI_EVENT_PACKET
     event_names: Dict[int, str] = {}
     event_classes: Dict[int, Type[HCI_Event]] = {}
-    vendor_factory: Optional[Callable[[bytes], Optional[HCI_Event]]] = None
+    vendor_factories: list[Callable[[bytes], Optional[HCI_Event]]] = []
 
     @staticmethod
     def event(fields=()):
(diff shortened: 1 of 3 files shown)


zxzxwu (approved):

bumble/hci.py
```diff
@@ -5148,8 +5161,8 @@ def from_bytes(cls, packet: bytes) -> HCI_Event:
         elif event_code == HCI_VENDOR_EVENT:
             # Invoke all the registered factories to see if any of them can handle
             # the event
-            if cls.vendor_factory:
-                if event := cls.vendor_factory(parameters):
+            for vendor_factory in cls.vendor_factories:
+                if event := vendor_factory(parameters):
                     return event
 
             # No factory, or the factory could not create an instance,
```
zxzxwu, line 5168: Should here return HCI_Vendor_Event instead of HCI_Event?

barbibulle: Yes. Done. I've also added a test for it.

PR #630, opened 2025-01-22
add support for ACL and ISO HCI packet queues

barbibulle:
ISO packets for BIS and CIS are now sent to the controller via an HCI packet queue that respects the buffering requirements of the controller.
In addition, the packet queues are made visible to the API users (advanced use case), and can emit 'drain' events when packets are marked as completed by the controller (this can be used for sending data at the right rate, rather than rely on a timer).
NOTE: the change to use the `drain` event rather than a timer in the `auracast` app is not included in this PR (coming in a later PR), but the use of the `drain` even is add, just to print the buffer levels as the packets are being sent (and we can see that with a timer approach, packets end up being buffered, which may become a problem eventually for long running streams, but this will get fixed in the upcoming PR).

### DIFF
--- a/bumble/host.py
+++ b/bumble/host.py
@@ -21,7 +21,6 @@ import collections
 import dataclasses
 import logging
 import struct
-import itertools
 
 from typing import (
     Any,
@@ -34,6 +33,8 @@ from typing import (
     cast,
     TYPE_CHECKING,
 )
+
+import pyee
 
 from bumble.colors import color
 from bumble.l2cap import L2CAP_PDU
@@ -60,7 +61,19 @@ logger = logging.getLogger(__name__)
 
 
 # -----------------------------------------------------------------------------
-class AclPacketQueue:
+class DataPacketQueue(pyee.EventEmitter):
+    """
+    Flow-control queue for host->controller data packets (ACL, ISO).
+
+    The queue holds packets associated with a connection handle. The packets
+    are sent to the controller, up to a maximum total number of packets in flight.
+    A packet is considered to be "in flight" when it has been sent to the controller
+    but not completed yet. Packets are no longer "in flight" when the controller
+    declares them as completed.
+
+    The queue emits a 'drain' event whenever one or more packets are completed.
+    """
+
     max_packet_size: int
 
     def __init__(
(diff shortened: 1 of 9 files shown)


bumble/host.py
```diff
@@ -729,17 +914,31 @@ def on_hci_command_complete_event(self, event):
...
             event.connection_handles, event.num_completed_packets
         ):
             if connection := self.connections.get(connection_handle):
-                connection.acl_packet_queue.on_packets_completed(num_completed_packets)
-            elif connection_handle not in itertools.chain(
-                self.cis_links.keys(),
-                self.sco_links.keys(),
-                itertools.chain.from_iterable(self.bigs.values()),
-            ):
+                connection.acl_packet_queue.on_packets_completed(
```
zxzxwu, line 924: May use the new method?
```suggestion
                if queue := self.get_data_packet_queue(connection_handle):
                  queue.on_packets_completed(num_completed_packets, connection_handle)
```

bumble/host.py
```diff
@@ -597,11 +715,78 @@ def send_l2cap_pdu(self, connection_handle: int, cid: int, pdu: bytes) -> None:
...
+            return connection.acl_packet_queue
+
+        if iso_link := self.cis_links.get(connection_handle) or self.bis_links.get(
+            connection_handle
+        ):
+            return iso_link.packet_queue
+
+        return None
+
+    def send_iso_sdu(self, connection_handle: int, sdu: bytes) -> None:
```
zxzxwu, line 734: Besides using drain events, I think maybe we can also allow tracking the completeness of each packet, or even make sending methods async, such as:

```py
@overload
async def send_iso_sdu(self, connection_handle: int, sdu: bytes, wait_for_completed: Literal[False]) -> None:
    ...

@overload
def send_iso_sdu(self, connection_handle: int, sdu: bytes, wait_for_completed: Literal[True]) -> None:
    ...

def send_iso_sdu(self, connection_handle: int, sdu: bytes,  wait_for_completed: Literal[True, False]) -> None | Awaitable[None]:
    ...
    packets = []
    while bytes_remaining:
        packet = ...
        packets.append(packet)
        iso_link.packet_queue.enqueue(packet)
    if wait_for_completed:
        return asyncio.gather([packet.completed for packet in packets])

```

For ISO there is probably just a small difference, but ACL links are usually used by multiple clients so they may never drain.

zxzxwu (approved):

bumble/host.py
zxzxwu: Actually for ISO usage, users can also be benefit from per-packet tracking. Because if we only have drain events, users must wait until all packets are completed, so if they want to have an audio buffer of multiple packets, there will still be a gap without any in flight packet like:
```py
def on_drain(queue):
  # At this moment no packet is in flight
  for _ in range(5):
    queue.host.send_iso_pdu(...)
```

If they can track each packet
```py
packets = Deque()
while True:
  if len(packets) < audio_buffer_size or await packets[0].completed:
    packet = ...
    packets.pop()
    packets.append(packet)
```

bumble/host.py
barbibulle: done

bumble/host.py
barbibulle: I ended up renaming the even from `drain` to `flow` (even if technically `drain` doesn't imply `empty`, it is sometimes used that way, so could be confusing, so `flow` makes it more clear that it's about packets going from one side of queue to the other)

PR #633, opened 2025-01-23
fix advertising parameter usage for legacy advertising

markusjellitsch:
when using `start_adverting()` the usage of `advertising_interval_min` and  `advertising_interval_max` for legacy advertising is not consistent with extended advertising (The advertising_interval given as input param should be calculated in 0.625 units)

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -380,8 +380,8 @@ class LegacyAdvertiser:
         # Set the advertising parameters
         await self.device.send_command(
             hci.HCI_LE_Set_Advertising_Parameters_Command(
-                advertising_interval_min=self.device.advertising_interval_min,
-                advertising_interval_max=self.device.advertising_interval_max,
+                advertising_interval_min=int(self.device.advertising_interval_min / 0.625),
+                advertising_interval_max=int(self.device.advertising_interval_max / 0.625),
                 advertising_type=int(self.advertising_type),
                 own_address_type=self.own_address_type,
                 peer_address_type=self.peer_address.address_type,
@@ -2583,8 +2583,8 @@ class Device(CompositeEventEmitter):
         auto_restart: bool = False,
         advertising_data: Optional[bytes] = None,
         scan_response_data: Optional[bytes] = None,
-        advertising_interval_min: Optional[int] = None,
-        advertising_interval_max: Optional[int] = None,
+        advertising_interval_min: Optional[float] = None,
+        advertising_interval_max: Optional[float] = None,
     ) -> None:
         """Start legacy advertising.
 


zxzxwu:
This API has been widely used, so we shouldn't roughly alter its spec. At least we should check its type to determine whether it should be considered 1 or 0.625 ms.

markusjellitsch:
Yes understand. At the moment it's a bit confusing and inconsistent, when you call ` start_advertising() `and legacy adv is used it is 0.625 ms, for extended adv it is 1.

barbibulle:
I think the change here makes sense. It was never the intention of the API to accept values in units of 0.625 (as a general rule, the Device API tries to hide internal spec details). All those intervals are supposed to be in milliseconds. In fact, for extended advertising, that's what the implementation does. 
So this should be considered a bug fix, not an API change.

markusjellitsch:
@barbibulle I reverted advertising_interval to `float`.

markusjellitsch:
I checked with `invoke project.pre-commit` and this seems fine (see log below). However Github Actions v3 seems deprecated from now on. Is there anything I can do from my side. 


`device_test.py::test_cis
device_test.py::test_cis_setup_failure
  /home/markus/projects/bumble/bumble/utils.py:467: FutureWarning: Only for testing
    warnings.warn(msg, FutureWarning)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================== 347 passed, 1 skipped, 7 warnings in 8.17s ==================
All good!`

barbibulle:
I'll need to update the GitHub actions, looks like some actions have become obsolete. I'll get to it now.

barbibulle:
Ah, looks like you've already done that :-)

bumble/device.py
```diff
@@ -2587,8 +2587,8 @@ async def start_advertising(
         auto_restart: bool = False,
         advertising_data: Optional[bytes] = None,
         scan_response_data: Optional[bytes] = None,
-        advertising_interval_min: Optional[float] = None,
-        advertising_interval_max: Optional[float] = None,
+        advertising_interval_min: Optional[int] = None,
```
barbibulle: Why not keep `float` here? While not strictly necessary, I think it's Ok to let the caller pass in non-integer millisecond values, since the spec supports it.

markusjellitsch: I reverted it because of pylint expecting int. It would require more code changes (adapt all internal advertising interval member variables to non-int). In the long run I'd definetely allow the caller to set the complete range of allowed values.

I am not sure if this change should be in the scope of this PR :-)

barbibulle: Let's leave like as int for now. I thought that `float` was supposed to imply that you could pass an `int` too (quoting https://typing.readthedocs.io/en/latest/spec/special-types.html#special-cases-for-float-and-complex ): 
```
Special cases for float and complex
Python’s numeric types complex, float and int are not subtypes of each other, but to support common use cases, the type system contains a straightforward shortcut: when an argument is annotated as having type float, an argument of type int is acceptable; similar, for an argument annotated as having type complex, arguments of type float or int are acceptable.
```

But maybe pylint isn't following that rule. We can leave that for a separate PR.

barbibulle (approved):

PR #643, opened 2025-02-06
auracast audio io

barbibulle:
This allows the auracast app to use the host's sound devices for input and output, in addition to other input and output types (file, stdio, external process).
The dependency on lc3 has changed, because PyPI does not allow `git` external dependencies.
Also included are small changes to some of the LE Audio GATT services, to use object types instead of raw bytes or tuples where appropriate (there are still maybe a few characteristics left to 'objectify', which can be done in subsequent PRs).
A later PR will include improvements to the typing support for GATT adapters.

### DIFF
--- a/bumble/audio/io.py
+++ b/bumble/audio/io.py
@@ -23,15 +23,19 @@ from concurrent.futures import ThreadPoolExecutor
 import dataclasses
 import enum
 import logging
+import pathlib
 from typing import (
     AsyncGenerator,
     BinaryIO,
     TYPE_CHECKING,
 )
+import sys
 import wave
 
+from bumble.colors import color
+
 if TYPE_CHECKING:
-    import sounddevice
+    import sounddevice  # type: ignore[import-untyped]
 
 
 # -----------------------------------------------------------------------------
@@ -78,9 +82,10 @@ def check_audio_output(output: str) -> bool:
     if output == 'device' or output.startswith('device:'):
         try:
             import sounddevice
-        except ImportError as exc:
+        except (ImportError, OSError) as exc:
             raise ValueError(
-                'audio output not available (sounddevice python module not installed)'
+                'audio output not available '
+                '(sounddevice python module not installed or failed to load)'
             ) from exc
 
         if output == 'device':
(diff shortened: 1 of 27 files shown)


bumble/audio/io.py
```diff
@@ -0,0 +1,550 @@
...
+# -----------------------------------------------------------------------------
+@dataclasses.dataclass
+class PcmFormat:
+    class Endianness(enum.Enum):
+        LITTLE = 0
+        BIG = 1
+
+    class SampleType(enum.Enum):
+        FLOAT32 = 0
+        INT16 = 1
```
zxzxwu, line 58: Why only s16 and f32? What about s16/24 and u16/24/32?

bumble/audio/io.py
```diff
@@ -0,0 +1,550 @@
...
+
+    endianness: Endianness
+    sample_type: SampleType
```
zxzxwu, line 61: There isn't any usage of this attribute inside this PR. When does it get used? Or it's just a reserved field?

bumble/profiles/le_audio.py
```diff
@@ -57,6 +69,44 @@ class Entry:
         tag: Metadata.Tag
         data: bytes
 
+        def decode(self) -> Any:
```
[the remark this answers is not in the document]
barbibulle: I took a simple approach here, using mostly basic builtin types and existing enums rather than dedicated classes (mostly strings, ints and enums). But if a different approach makes senses, we can evaluate that in a later PR.

bumble/audio/io.py
barbibulle: Those are the only two types currently used. If/when we need other types (for example support some form of conversion to/from other types), they can be added.

bumble/audio/io.py
barbibulle: It is used implicitly when instances of the classes are printed out (like in a debug log).
But is should also be used after calling `create_audio_input` to check that the sample type is supported (the code currently only checks the number of channels). I'll fix that.

PR #644, opened 2025-02-09
Advertising Set Info Transfer

zxzxwu:


bumble/device.py
```diff
@@ -794,6 +794,17 @@ async def remove(self) -> None:
         )
         del self.device.extended_advertising_sets[self.advertising_handle]
 
+    async def periodic_info_transfer(
```
barbibulle, line 797: I suggest naming this method `transfer_periodic_info` to be consistent with the usual convention of `verb_object` order.

barbibulle (approved):

bumble/device.py
zxzxwu: Done

PR #646, opened 2025-02-11
Improve AdvertisingData type annotations

zxzxwu:
* Add overloads to provide better return type hints
* Make advertising data type enum so they can be considered constants

### DIFF
--- a/bumble/core.py
+++ b/bumble/core.py
@@ -1,4 +1,4 @@
-# Copyright 2021-2022 Google LLC
+# Copyright 2021-2025 Google LLC
 #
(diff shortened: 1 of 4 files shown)


bumble/core.py
```diff
@@ -90,7 +90,7 @@ class BaseError(BaseBumbleError):
 
     def __init__(
         self,
-        error_code: Optional[int],
+        error_code: int | None,
```
barbibulle, line 93: Using | for unions is only supported for Python >= 3.10, so we can't use it yet.

barbibulle (approved):

PR #648, opened 2025-02-16
Add setup_data_path for iso queues

pstrueb:
The auracast app only starts work again if we call setup_data_path after instantiating the IsoPacketStream object.

Before this change, the Number of completed packages event from controller to host always returns a zero handle like so:
```
< ISO Data TX: Handle 39 flags 0x02 dlen 104                                                                                                                                      #97 284.682500
> HCI Event: Number of Completed Packets (0x13) plen 5                                                                                                                            #98 284.682700
        Num handles: 1
        Handle: 0
        Count: 1
```
After this change, The broadcast is sent without any errors and can also be received with a pair of earbuds.

The setup that was used two test is with transport serial.
Tested with:
1. nrf54l15-dk with hci_uart firmware 
2. nrf52dongle with hci_uart over usb cdc

### DIFF
--- a/apps/auracast.py
+++ b/apps/auracast.py
@@ -965,6 +965,11 @@ async def run_transmit(
                 bumble.device.IsoPacketStream(big.bis_links[1], 64),
             ]
 
+            for queue in iso_queues:
+                await queue.iso_link.setup_data_path(
+                    direction=queue.iso_link.Direction.HOST_TO_CONTROLLER
+            )
+
             def on_flow():
                 data_packet_queue = iso_queues[0].data_packet_queue
                 print(
--- a/apps/auracast.py
+++ b/apps/auracast.py
@@ -692,7 +692,7 @@ async def run_receive(
         def on_new_broadcast(broadcast: BroadcastScanner.Broadcast) -> None:
             if scan_result.done():
                 return
-            if broadcast_id is None or broadcast.broadcast_:id == broadcast_id:
+            if broadcast_id is None or broadcast.broadcast_id == broadcast_id:
                 scan_result.set_result(broadcast)
 
         scanner.on('new_broadcast', on_new_broadcast)
--- a/apps/auracast.py
+++ b/apps/auracast.py
@@ -962,7 +962,7 @@ async def run_transmit(
             for bis_link in big.bis_links:
                 await bis_link.setup_data_path(
                     direction=bis_link.Direction.HOST_TO_CONTROLLER
-            )
+                )
 
             iso_queues = [
                 bumble.device.IsoPacketStream(big.bis_links[0], 64),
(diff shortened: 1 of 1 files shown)


apps/auracast.py
```diff
@@ -965,6 +965,11 @@ async def run_transmit(
                 bumble.device.IsoPacketStream(big.bis_links[1], 64),
             ]
 
+            for queue in iso_queues:
```
barbibulle, line 968: Thanks for catching that. Somehow this got accidentally removed in a recent PR and went unnoticed (it still worked with the BT controller on my Pixel phone...).
However, it might be more straightforward to express this as:
```
for bis_link in big.bis_links:
    await bis_link.setup_data_path(
        direction=bis_link.Direction.HOST_TO_CONTROLLER
)
```
above line 963

barbibulle:
Would you be able to share the `.conf` files you used to build the HCI UART sample apps on Zephyr with ISO packets enabled? (I have a few nrf54l15-dk and nrf52dongle boards, I'd love to be able to use them for LE audio, but the config "out of the box" doesn't seem to have all the controller features enabled, and my very simple attempt at trying to add things to the config didn't quite work out)

pstrueb:
> Would you be able to share the `.conf` files you used to build the HCI UART sample apps on Zephyr with ISO packets enabled? (I have a few nrf54l15-dk and nrf52dongle boards, I'd love to be able to use them for LE audio, but the config "out of the box" doesn't seem to have all the controller features enabled, and my very simple attempt at trying to add things to the config didn't quite work out)

It dont see why not :). I put it here: [hci_uart](https://github.com/pstrueb/hci_uart). Im not sure if this covers all the usecases since I tested mostly broadcasting but it sure is a starting point. See sample.yaml for some hints how to build. The flash_dongle.sh script can be used to flash the nrf52dongle after putting it into bootloader mode.

apps/auracast.py
```diff
@@ -692,7 +692,7 @@ async def run_receive(
         def on_new_broadcast(broadcast: BroadcastScanner.Broadcast) -> None:
             if scan_result.done():
                 return
-            if broadcast_id is None or broadcast.broadcast_id == broadcast_id:
+            if broadcast_id is None or broadcast.broadcast_:id == broadcast_id:
```
barbibulle, line 695: Accidental keystroke?

barbibulle:
This just needs a formatting pass before we can commit (check the error message in the output of the GH Actions)

pstrueb:
> This just needs a formatting pass before we can commit (check the error message in the output of the GH Actions)

Sorry buts its not really clear for me from the logs what the formatter wanted to change or what the linter complained about. I was able to run black locally. So we will see if this is ok now.

barbibulle (approved):

PR #669, opened 2025-04-12
Cleanup relative imports

zxzxwu:
PEP-8 and Google Pystyle both suggest to use absolute import instead of relative import.

### DIFF
--- a/bumble/a2dp.py
+++ b/bumble/a2dp.py
@@ -26,9 +26,9 @@ from typing import Awaitable, Callable
 from typing_extensions import ClassVar, Self
 
 
-from .codecs import AacAudioRtpPacket
-from .company_ids import COMPANY_IDENTIFIERS
-from .sdp import (
+from bumble.codecs import AacAudioRtpPacket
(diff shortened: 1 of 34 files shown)


zxzxwu:
I think later we should introduce isort or ruff to make sure we have a more consistent importing style?

zxzxwu:
Ruff implements relative import replacement: https://docs.astral.sh/ruff/rules/relative-imports/ and it's extremely fast to do Black and Pylint's work...

However, Ruff style is not fully compatible with Black, so if we want to introduce Ruff, we might need to abandon Black.

barbibulle (approved):

barbibulle:
ruff is a great tool, so it might make sense to switch to that at some point in the future, but I think it's still best to stick with Black + other tools for now, as that's pretty stable. Let's revisit that in a year or so.

zxzxwu:
Sounds good. Ruff doesn't fully cover pylint yet, so we can wait until it has a better coverage.

PR #670, opened 2025-04-13
Make all event emitters abortable and async

zxzxwu:
Since Bumble is almost always used in async context, and async handlers are commonly used, we can make all emitters async, and also abortable to simplify management.

### DIFF
--- a/bumble/utils.py
+++ b/bumble/utils.py
@@ -38,7 +38,8 @@ from typing import (
 )
 from typing_extensions import Self
 
-from pyee import EventEmitter
+import pyee
+import pyee.asyncio
 
 from bumble.colors import color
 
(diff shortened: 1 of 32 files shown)


barbibulle:
Using `AsyncIOEventEmitter` is definitely the right thing to do.
But for `AbortableEventEmitter`, I think this would be a good opportunity to refactor the use of this class.
It really shouldn't be a class (and also, its name is misleading, because the event emitter isn't itself "abortable", it is, instead, an event emitter that will cancel a future when emitting some event.
A good refactor would be to make `abort_on` a utility function instead of a class method, named `cancel_on_event(event_emitter, event, awaitable)`, which would do the same as `abort_on`, but without needing a class. (also, the term `cancel` is more appropriate than `abort` here, since that's the term used by `asyncio`).

zxzxwu:
I agree. But I think we can still have an EventEmitter class(`BaseBumbleEventEmitter`?) derived from AsyncIOEventEmitter for future usage.

bumble/utils.py
```diff
@@ -184,38 +229,48 @@ def close(self) -> None:
...
-        self.on(event, on_event)
-        future.add_done_callback(on_done)
-        return future
+    emitter.on(event, on_event)
+    future.add_done_callback(on_done)
+    return future
+
+
+# -----------------------------------------------------------------------------
+class BaseBumbleEventEmitter(pyee.asyncio.AsyncIOEventEmitter):
```
barbibulle, line 263: Maybe we could keep this name short, just calling it `EventEmitter`? Since it is in the `utils` module, that already provides a form of namespace.

barbibulle (approved):

bumble/utils.py
zxzxwu: Done

PR #678, opened 2025-05-03
fix a few timescale adjustments

barbibulle:
There were still a few places where the timescales were not properly converted between the internal (HCI) level and the external (Device API) level.
The basic rule is that at the "user" level, the Device API should always try to hide some of the low-level details, including some the HCI-level timescales (where some time values are expressed, for example, in units of 1.25ms or 0.625ms).
This means that data structure and parameters at the `host.py` level (including parameters for events emitted by the `Host` object) use the low level HCI timescales, whereas public data structures and parameters at the `device.py` level use higher level "natural" units (seconds, milliseconds, ...).

This change also includes support for `HCI_LE_Enhanced_Connection_Complete_V2_Event` which was missing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
     "prompt_toolkit >= 3.0.16; platform_system!='Emscripten'",
     "prettytable >= 3.6.0; platform_system!='Emscripten'",
     "protobuf >= 3.12.4; platform_system!='Emscripten'",
-    "pyee >= 8.2.2",
+    "pyee >= 13.0.0",
     "pyserial-asyncio >= 0.5; platform_system!='Emscripten'",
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
     "prompt_toolkit >= 3.0.16; platform_system!='Emscripten'",
     "prettytable >= 3.6.0; platform_system!='Emscripten'",
     "protobuf >= 3.12.4; platform_system!='Emscripten'",
-    "pyee >= 8.2.2",
+    "pyee >= 13.0.0",
```
zxzxwu, line 29: What's the purpose here?

zxzxwu (approved):

PR #679, opened 2025-05-03
gbg/pairing ios

barbibulle:
Enhance the pairing app so that it can work better with iOS devices.


bumble/core.py
```diff
@@ -1351,6 +1351,12 @@ class Type(utils.OpenIntEnum):
         THREE_D_INFORMATION_DATA                            = 0x3D
         MANUFACTURER_SPECIFIC_DATA                          = 0xFF
 
+    class Flags(enum.IntFlag):
+        LE_LIMITED_DISCOVERABLE_MODE = 1 << 0
+        LE_GENERAL_DISCOVERABLE_MODE = 1 << 1
+        BR_EDR_NOT_SUPPORTED = 1 << 2
+        SIMULTANEOUS_LE_BR_EDR_CAPABLE = 1 << 3
```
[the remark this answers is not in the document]
barbibulle: done

bumble/device.py
```diff
@@ -5057,6 +5072,15 @@ def on_flush(self):
...
     @host_event_handler
     def on_link_key(self, bd_addr, link_key, key_type):
+        authenticated = key_type in (
+            hci.HCI_AUTHENTICATED_COMBINATION_KEY_GENERATED_FROM_P_192_TYPE,
+            hci.HCI_AUTHENTICATED_COMBINATION_KEY_GENERATED_FROM_P_256_TYPE,
+        )
+        pairing_keys = PairingKeys()
+        pairing_keys.link_key = PairingKeys.Key(
+            value=link_key, authenticated=authenticated
+        )
```
zxzxwu, line 5082: Seems like a repeat of `if self.keystore:...`

bumble/device.py
```diff
@@ -5069,13 +5093,15 @@ def on_link_key(self, bd_addr, link_key, key_type):
...
 
             utils.cancel_on_event(
-                self, 'flush', self.update_keys(str(bd_addr), pairing_keys)
+                self, Device.EVENT_FLUSH, self.update_keys(str(bd_addr), pairing_keys)
             )
 
         if connection := self.find_connection_by_bd_addr(
             bd_addr, transport=PhysicalTransport.BR_EDR
         ):
+            connection.link_key = link_key
```
zxzxwu, line 5102: Saving link_key is a little bit weird here. In spec, link key is almost always not associated with connection but an address.Saving link_key is a little bit weird here. In spec, link key is almost always not associated with a connection but an address.

If we worry about optional keystore might not be set, we may make it required and put an `WriteOnlyMemoryKeystore` default implementation which only records keys without providing them.

zxzxwu (commented):
Sorry, somehow I missed to send the comments.

The design of link_key storage looks sus than flags, so I only put a comment without approval.

bumble/device.py
barbibulle: Fixed

bumble/device.py
barbibulle: You are 100% right. I had added `link_key` here to go along with the existing `link_key_type`, but now that you mention it, neither of those really belong here. They only belong in the key store. 
I removed `link_key`, but I'm leaving `link_key_type` as-is for now, because it seems that Avatar relies on it, but we should fix the Avatar test to look up the link key in the key store rather than in a property of the connection. Once that's fixed, we can remove the property.

zxzxwu (approved):

PR #683, opened 2025-05-05
Implement builtin cryptography primitives

zxzxwu:
1000 times of AES-CMAC:
* Cryptography: 0.05s
* Builtin: 1.62s

1000 times of ECDH:
* Cryptography: 0.15s
* Builtin: 7.46s

Obviously, pure Python implementation is pretty slow, but it's acceptable since each invocation is still less than 0.01s.


zxzxwu:
The implementation is modified from other repos
* AES - pyaes (MIT)
* CMAC - pycrypto (public domain)
* ECC - Porting from AOSP/Rootcanal Rust implementation (Apache, Google-self authored)

bumble/crypto.py
```diff
@@ -23,21 +23,27 @@
...
+from bumble import _crypto
+
+try:
+    from cryptography.hazmat.primitives import ciphers
+    from cryptography.hazmat.primitives.ciphers import algorithms
+    from cryptography.hazmat.primitives.ciphers import modes
+    from cryptography.hazmat.primitives.asymmetric import ec
+    from cryptography.hazmat.primitives import cmac
+
+    _use_builtin_backend = False
```
barbibulle, line 41: I'd like to propose a slightly different way of splitting the implementation. Instead of relying on `if __use_builtin_backend / else`, we could have 3 files: `crypto.py` with the common parts, `crypto_builtin.py` with the new builtin implementation, and `crypto_cryptography.py` with the one going through the `cryptography` module.
Then, at the top of `crypto.py`, you can `try from crypto_cryptography import EccKey, e, aes_mac except ImportError: from crypto_builtin import EccKey, e, aes_mac`. 
This way, the two variants are well isolated.
(for EccKey, you could have an abstract `BaseEccKey` in `crypto.py` that the two variants would subclass)

zxzxwu: Sure, this will be much cleaner.

zxzxwu: Done. Moving these modules into a subpackage _crypto.

bumble/crypto.py
```diff
@@ -25,19 +25,15 @@
...
-    ECDH,
-    EllipticCurvePrivateKey,
-    EllipticCurvePublicNumbers,
-    EllipticCurvePrivateNumbers,
-    SECP256R1,
-)
-from cryptography.hazmat.primitives import cmac
-from typing import Tuple
+
+try:
```
barbibulle, line 30: This could be merged in `bumble/crypto/__init__.py`, and the other files in `bumble/crypto` instead of `bumble/_crypto`.

zxzxwu: Do you mean we don't keep `bumble/crypto.py` anymore, and move original `bumble/crypto.py` to `bumble/crypto/__init__.py` and import the backends there?

barbibulle: Yes. I think we can also remove the _ prefix for the files in the crypto directory, they could be simply named builtin.py and cryptography.py

barbibulle (approved):

bumble/crypto.py
zxzxwu: Done. My initial idea is to hide some implementation details, but it should still be fine to expose them.

PR #684, opened 2025-05-07
cryptography: bump version to 44.0.3 to fix python parsing

SilverBzH:
Bug: 404336381


barbibulle (approved):

zxzxwu:
For Rust usage, maybe we can turn to use builtin crypto backend?

https://github.com/google/bumble/pull/683

SilverBzH:
> For Rust usage, maybe we can turn to use builtin crypto backend?
> 
> #683

Either is fine to me. I think this is independent from the bump versioning.

PR #685, opened 2025-05-08
A2DP example: Codec selection based on file type

SilverBzH:
Currently support SBC and AAC

### DIFF
--- a/examples/run_a2dp_source.py
+++ b/examples/run_a2dp_source.py
@@ -16,28 +16,43 @@
 # Imports
 # -----------------------------------------------------------------------------
 import asyncio
-import sys
-import os
 import logging
+import os
+import sys
+from dataclasses import dataclass
 
-from bumble.colors import color
-from bumble.device import Device
-from bumble.transport import open_transport_or_link
-from bumble.core import PhysicalTransport
+import ffmpeg
+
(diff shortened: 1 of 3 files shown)


examples/run_a2dp_source.py
```diff
@@ -16,28 +16,43 @@
...
 import logging
+import os
+import sys
+from dataclasses import dataclass
 
-from bumble.colors import color
-from bumble.device import Device
-from bumble.transport import open_transport_or_link
-from bumble.core import PhysicalTransport
+import ffmpeg
```
barbibulle, line 24: We might not really need ffmpeg for this. Can you check if the parsers that are already in the code base are sufficient for what you need? (see `apps/player/player.py` for an example of how `AacParser`, `SbcParser` and `OpusParse` are used)

SilverBzH: I've used ffmpeg to retrieve the information from the file metadata, including the type of encoding ; which might not be needed if we just look at the extension like in the player.py indeed.

However, by looking at the player.py, I don't think this CL is needed, I could just use the player.py instead.

SilverBzH: Just tested it and the player is fully enough. This PR is not needed.

PR #686, opened 2025-05-09
HCI: Avoid patching __init__

zxzxwu:
op_code is usually set to class (not instance) inside decorators, and passing it to __init__ is not a common scenario, so here we change the order of __init__ to force it a keyword argument.

Also replace some legacy `registered` commands with listed fields.

### DIFF
--- a/bumble/hci.py
+++ b/bumble/hci.py
@@ -23,7 +23,8 @@ import functools
 import logging
 import secrets
 import struct
-from typing import Any, Callable, Iterable, Optional, Union, ClassVar
+from collections.abc import Sequence
+from typing import Any, Callable, Iterable, Optional, Union, TypeVar, ClassVar
 from typing_extensions import Self
 
 from bumble import crypto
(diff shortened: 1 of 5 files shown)


zxzxwu:
There are some code not well tested, so I left this PR draft.

bumble/hci.py
```diff
@@ -3738,39 +3757,15 @@ class HCI_LE_Read_Advertising_Physical_Channel_Tx_Power_Command(HCI_Command):
...
-            {
-                'parser': HCI_Object.parse_length_prefixed_bytes,
-                'serializer': functools.partial(
-                    HCI_Object.serialize_length_prefixed_bytes, padded_size=32
-                ),
-            },
-        )
-    ]
-)
+@HCI_Command.command([('advertising_data', 'v')])
```
barbibulle, line 3760: This won't pad to 32 bytes if `advertising_data` is less than 32 bytes.

bumble/hci.py
```diff
@@ -3738,39 +3757,15 @@ class HCI_LE_Read_Advertising_Physical_Channel_Tx_Power_Command(HCI_Command):
...
-            {
-                'parser': HCI_Object.parse_length_prefixed_bytes,
-                'serializer': functools.partial(
-                    HCI_Object.serialize_length_prefixed_bytes, padded_size=32
-                ),
-            },
-        )
-    ]
-)
+@HCI_Command.command([('scan_response_data', 'v')])
```
barbibulle, line 3768: Same comment as above.

barbibulle (commented):
Looks good in general. I'll take a second close look when it's ready for a final review.

zxzxwu:
Thanks! I would like to release a version first for crypto changes and merge this PR later.

bumble/hci.py
zxzxwu: Done. Reverted.

bumble/hci.py
zxzxwu: Reverted.

zxzxwu:
Generally done, but later some changes in this PR will be replaced by dataclasses.

barbibulle (approved):

PR #688, opened 2025-05-10
Dataclass-based ASCS Packets

zxzxwu:
This has been used in [Navi](https://github.com/google/bt-navi-tests/blob/main/navi/bumble_ext/hci.py) for a while. Recently I resolved several critical issues and I think it's a time to bring it into Bumble upstream.

An obvious benefit to use dataclasses is IDE and type checker integration. 

image

Unlike https://github.com/google/bumble/pull/430, this design properly utilize the existing effort, especially field specs. For nested fields, we may need to set another field in metadata like `Annotated[list, 1, '[' or ']' ]` as stack operator, though usually we only support up to 2 layers.

return_parameters_fields cannot be benifit from this, because current type checker cannot associate two classes. But BTW, return_parameters_fields is actually not necessary to be inside decorator, so later we may have a PR to move them from decorators to classes.

### DIFF
--- a/bumble/hci.py
+++ b/bumble/hci.py
@@ -108,6 +108,36 @@ def phy_list_to_bits(phys: Optional[Iterable[Phy]]) -> int:
 
 
 # -----------------------------------------------------------------------------
+# Field Metadata
+# -----------------------------------------------------------------------------
+# Field specification can be:
+# - a dict with "serializer", "parser", "size", "mapper" keys
+# - a callable that takes (packet, offset) and returns (new_offset, value) (deserialize only)
+# - a string of
+#   - ">2" and ">4" for 2-byte and 4-byte big-endian integers
+#   - "*" for all remaining bytes in the packet
+#   - "v" for variable length bytes with a leading length byte
+# - an integer [1, 4] for 1-byte, 2-byte or 4-byte unsigned little-endian integers
+# - an integer [-2, -1] for 1-byte, 2-byte signed little-endian integers
+FieldSpec = Union[dict[str, Any], Callable[[bytes, int], tuple[int, Any]], str, int]
+
+
+@dataclasses.dataclass
+class FieldMetadata:
+    spec: FieldSpec
+    list_begin: bool = False
+    list_end: bool = False
+
+
+def metadata(
+    spec: FieldSpec, list_begin: bool = False, list_end: bool = False
+) -> dict[str, Any]:
+    return {
+        "bumble.hci": FieldMetadata(spec=spec, list_begin=list_begin, list_end=list_end)
+    }
+
+
+# -----------------------------------------------------------------------------
 # Constants
 # -----------------------------------------------------------------------------
 # fmt: off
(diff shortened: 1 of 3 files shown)


zxzxwu:
I believe there are still a lot of places to polish, so this is just an initial draft for reviews.

Shahnuxs (approved):

barbibulle:
I like the idea. Having support for proper typechecking via dataclasses is a great way to go (easier that designing type-checking plugins). We'll need to make sure that we keep things still compatible with Python >= 3.9.

zxzxwu:
Thanks! `typing.Annotated` is introduced in Python 3.9, so compatability shouldn't be a problem (and even under 3.9 we can import it from `typing_extensions`). But of course we will need some test cases to handle them.

I will work on https://github.com/google/bumble/pull/686 first, and them continue here.

zxzxwu:
A problem here is that with `from __future__ import annotations`, Annotations including metadata will become string in Runtime. Evaluating them from HCI module is not a good idea because it's not always possible to get the symbols.

There are 2 solutions to do this:
1. Avoid using `from __future__ import annotations` when dataclass-based packets are used. It seems PEP 563 will never become mandatory (replaced by PEP 649), so it should be safe, but it requires some rework to remove this import.
2. Using dataclasses.field(metadata={...}) to store field info, but this will be a little bit long-winded.

zxzxwu:
This PR is ready for review. Let's get started from ASCS packets since the test is more comprehensive here.

bumble/hci.py
```diff
@@ -107,6 +107,16 @@ def phy_list_to_bits(phys: Optional[Iterable[Phy]]) -> int:
     return phy_bits
 
 
+def metadata(
+    spec: Any, list_begin: bool = False, list_end: bool = False
+) -> dict[str, Any]:
+    return {
```
barbibulle, line 135: I think we should put all our metadata in a single namespace. The expectation for dataclasses metadata is that each "type extension" uses its own namespace, and you can have multiple such extensions. In our case, we just need one, to represent the "bumble HCI types". We can name it `bumble.hci`. The data type for what's in the metadata dict for that namespace is up to us. One model is to use a dictionary, with a "spec" entry that has for value the exact same "spec" data as what's currently in the decorator's `fields` entry, plus some new options to represent the list start and end (in my example, I used "options" as a generic dict-based way to extend the data type, with a "list_begin" and "list_end" entry. 
This would give us a metadata dict like this:
```
"bumble.hci": {
    "spec": ...
    "options": {
        "list_start": True
    }
}
```
or even up-leveling the options with something like:
```
"bumble.hci": {
    "spec": ...,
    "list_start": True
}
```
But we could also define a structured type here instead of a generic dict, possibly something like:
```
@dataclass
def HciTypeMetadata:
    spec: dict[str, Any] | int | str | ...
    list_start: bool = False
    list_end: bool = False
```
so we'd have this form of metadata:
```
"bumble.hci": HciTypeMetadata(spec=..., list_start=True)
```

I think either way is fine. Maybe the `HciTypeMetadata` structured type is a bit cleaner. But since it's all internal (the metadata will be created by a helper function), we can easily change that without impacting all the declarations, only the object factories need to worry about it.

bumble/hci.py
```diff
@@ -107,6 +107,36 @@ def phy_list_to_bits(phys: Optional[Iterable[Phy]]) -> int:
     return phy_bits
 
 
+# -----------------------------------------------------------------------------
+# Field Metadata
+# -----------------------------------------------------------------------------
+# Field specification can be:
+# - a dict with "serializer", "parser", "size", "mapper" keys
+# - a callable that takes a packet and returns a value (serialize only)
```
barbibulle: nit: `de-serialize only`

PR #692, opened 2025-05-13
AVDTP: Prevent error on parsing empty capabilities

zxzxwu:


zxzxwu:
Not sure whether Bumble should handle this. More like a bug of remote side.

PR #700, opened 2025-05-20
Fix wrong remove iso data path parameter

zxzxwu:
Unlike `HCI_LE_Setup_ISO_Data_Path_Command`, `HCI_LE_Remove_ISO_Data_Path_Command` takes a flag parameter instead of an enum parameter.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -1464,7 +1464,7 @@ class _IsoLink:
             check_result=True,
         )
 
-    async def remove_data_path(self, direction: _IsoLink.Direction) -> int:
+    async def remove_data_path(self, directions: Iterable[_IsoLink.Direction]) -> int:
         """Remove a data path with controller on given direction.
 
         Args:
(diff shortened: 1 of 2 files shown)


zxzxwu:
Maybe we should allow passing a bitmap instead of convert them inside the function.

bumble/device.py
```diff
@@ -1476,7 +1476,7 @@ async def remove_data_path(self, direction: _IsoLink.Direction) -> int:
         response = await self.device.send_command(
             hci.HCI_LE_Remove_ISO_Data_Path_Command(
                 connection_handle=self.handle,
-                data_path_direction=direction,
+                data_path_direction=1 << direction,
```
barbibulle, line 1479: `remove_data_path` could maybe take an `directions: Iterable[_IsoLink.Direction]`?

barbibulle (approved):

bumble/device.py
zxzxwu: Done.

PR #701, opened 2025-05-27
speaker app: enable opus, enable more options

barbibulle:
Adding support for Opus playback. Should work in Chrome (not tested on other browsers).
This shifts from using MSE, which isn't very good for low latency, to using WebAudio and WebCodecs, which gives us a much better latency, and supports both AAC and Opus streams.
This PR also includes a few command line options to allow controlling which bitrates and sampling frequencies are advertised. 
Also added to the UI: display the current bitrate (averaged over the last ~30 buffers).

### DIFF
--- a/apps/speaker/speaker.py
+++ b/apps/speaker/speaker.py
@@ -50,8 +50,10 @@ from bumble.a2dp import (
     make_audio_sink_service_sdp_records,
     A2DP_SBC_CODEC_TYPE,
     A2DP_MPEG_2_4_AAC_CODEC_TYPE,
+    A2DP_NON_A2DP_CODEC_TYPE,
     SbcMediaCodecInformation,
     AacMediaCodecInformation,
+    OpusMediaCodecInformation,
 )
 from bumble.utils import AsyncRunner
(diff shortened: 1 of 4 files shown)


apps/speaker/speaker.py
```diff
@@ -675,7 +743,26 @@ def speaker_cli(ctx, device_config):
...
+    help='Enable a sampling frequency (may be specified more than once)',
+)
+@click.option(
+    '--bitrate',
+    metavar='BITRATE',
+    type=int,
+    help='Supported bitrate (AAC only)',
+)
+@click.option(
+    '--vbr/--no-vbr', is_flag=True, default=True, help='Enable VBR (AAC only)'
```
SilverBzH, line 765: Should default be `false` if it is only for aac ?

SilverBzH (approved):

apps/speaker/speaker.py
barbibulle: It's ignored for non-AAC codecs, but I think it's convenient for AAC to have VBR on by default.

PR #702, opened 2025-05-27
add advertise and scan options to the bench app

barbibulle:
This allows experimenting with the behavior of the controller when sending/receiving data while a scan and/or advertising is still going on.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -99,9 +99,9 @@ logger = logging.getLogger(__name__)
 # fmt: off
 # pylint: disable=line-too-long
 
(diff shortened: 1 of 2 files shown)


bumble/device.py
```diff
@@ -99,9 +99,9 @@
 # fmt: off
 # pylint: disable=line-too-long
 
-DEVICE_MIN_SCAN_INTERVAL                      = 25
+DEVICE_MIN_SCAN_INTERVAL                      = 2.5
```
richardissuperman, line 102: does this mean changing min scan interval to 2.5 milli seconds?

PR #708, opened 2025-06-09
add passkey delegate support for pairing

barbibulle:
As discussed in #705 
Passkey generation has a default implementation in `PairingDelegate` which can be overridden by subclasses.


bumble/smp.py
```diff
@@ -975,9 +967,20 @@ def display_or_input_passkey(
...
+
+            async def display_passkey():
+                await self.display_passkey()
+                if next_steps is not None:
+                    next_steps()
+
+            try:
+                utils.cancel_on_event(
+                    self.connection,
+                    'disconnection',
```
zxzxwu:
```suggestion
                    self.connection.EVENT_DISCONNECTION,
```

bumble/smp.py
```diff
@@ -1850,19 +1853,23 @@ def on_smp_pairing_public_key_command(
             elif self.pairing_method == PairingMethod.PASSKEY:
                 self.send_pairing_confirm_command()
         else:
-            if self.pairing_method == PairingMethod.PASSKEY:
-                self.display_or_input_passkey()
 
-            # Send our public key back to the initiator
-            self.send_public_key_command()
+            def next_steps():
```
zxzxwu: To enforce type checking
```suggestion
            def next_steps() -> None:
```

bumble/smp.py
```diff
@@ -946,14 +945,7 @@ def display_passkey(self) -> None:
...
 
-        try:
-            utils.cancel_on_event(
-                self.connection,
-                'disconnection',
-                self.pairing_config.delegate.display_number(self.passkey, digits=6),
-            )
-        except Exception as error:
-            logger.warning(f'exception while displaying number: {error}')
+        await self.pairing_config.delegate.display_number(self.passkey, digits=6)
```
zxzxwu, line 947: nit: This will become blocking, but it looks intended. Maybe we should have a better documentation about which methods are and should be blocking.

bumble/pairing.py
```diff
@@ -222,6 +223,14 @@ async def key_distribution_response(
             ),
         )
 
+    async def generate_passkey(self) -> int:
+        """
+        Return a passkey value between 0 and 999999 (inclusive).
+        """
+
+        # By default, generate a random passkey.
+        return secrets.randbelow(1000000)
```
zxzxwu, line 232: Unrelated: Currently, both PairingConfig and PairingDelegate don't have information about the pairing session. I think we may need to add such information like a connection attribute in the future.

zxzxwu (approved):

bumble/smp.py
barbibulle: all the delegate methods are made async so that they can block if needed. I'm planning on doing a complete doc of the pairing delegate, where this will be explained in details.

bumble/smp.py
barbibulle: done

bumble/pairing.py
barbibulle: Actually, the pairing configs can use the connection context if they need it. The `PairingConfig` and `PairingDelegate` base classes don't store it, because they don't need it, but `pairing_config_factory` takes a `connection` arg, so if your config or delegate needs the context, it can keep it.
Ex:
`device.pairing_config_factory = lambda connection: MyPairingConfig(connection)`

bumble/smp.py
barbibulle: I ended up creating a help function for the common call: `utils.cancel_on_event(self.connection, 'disconnection', foo()` --> `self.connection.cancel_on_disconnection(foo())`

PR #710, opened 2025-06-13
Dataclass-based HCI packets

zxzxwu:


zxzxwu:
Unrelated: Later we can define a `SpectableEnum` class, so that some dict specs could be simplified as
```py

class SpecableEnum(utils.OpenIntEnum):

    @classmethod
    def type_spec(cls, size: int = 1):
        return {'size': size, 'mapper': lambda x: cls(x).name}

@HCI_Event.dataclass_event
@dataclasses.dataclass
class HCI_Connection_Complete_Event(HCI_Event):
    '''
    See Bluetooth spec @ 7.7.3 Connection Complete Event
    '''

    class LinkType(SpecableEnum):
        SCO = 0x00
        ACL = 0x01
        ESCO = 0x02

    status: int = field(metadata=metadata(STATUS_SPEC))
    connection_handle: int = field(metadata=metadata(2))
    bd_addr: int = field(metadata=metadata(Address.parse_address))
    link_type: int = field(metadata=metadata(LinkType.type_spec(1)))
    encryption_enabled: int = field(metadata=metadata(1))
```

barbibulle:
> Unrelated: Later we can define a `SpectableEnum` class, so that some dict specs could be simplified as
> 
> ```python
> class SpecableEnum(utils.OpenIntEnum):
> 
>     @classmethod
>     def type_spec(cls, size: int = 1):
>         return {'size': size, 'mapper': lambda x: cls(x).name}
> 
> @HCI_Event.dataclass_event
> @dataclasses.dataclass
> class HCI_Connection_Complete_Event(HCI_Event):
>     '''
>     See Bluetooth spec @ 7.7.3 Connection Complete Event
>     '''
> 
>     class LinkType(SpecableEnum):
>         SCO = 0x00
>         ACL = 0x01
>         ESCO = 0x02
> 
>     status: int = field(metadata=metadata(STATUS_SPEC))
>     connection_handle: int = field(metadata=metadata(2))
>     bd_addr: int = field(metadata=metadata(Address.parse_address))
>     link_type: int = field(metadata=metadata(LinkType.type_spec(1)))
>     encryption_enabled: int = field(metadata=metadata(1))
> ```

I like this idea. We can also consider a simple variant on this idea: return directly the `metadata` value, which would make the field definition even shorter: `link_type: int = field(metadata=LinkType.type_metadata(1))`

bumble/hci.py
```diff
@@ -1995,6 +2004,19 @@ def __str__(self):
...
 
 
+# -----------------------------------------------------------------------------
+@dataclasses.dataclass
+class HCI_Dataclass_Object(HCI_Object):
+    def __post_init__(self) -> None:
+        self.fields = HCI_Object.fields_from_dataclass(self)
+
+    @classmethod
+    def parse_from(cls, data: bytes, offset: int) -> tuple[int, Self]:
```
barbibulle: I suggest calling this `parse_from_bytes` to be consistent with similar methods elsewhere.

bumble/hci.py
```diff
@@ -2302,6 +2325,14 @@ def inner(cls: type[Command]) -> type[Command]:
 
         return inner
 
+    @classmethod
+    def dataclass_command(cls, subclass):
+        # TODO: Move this to __post_init__ when all packets become dataclasses.
+        subclass.parameters = property(
```
barbibulle: The reason why `parameters` is a static field, and not computed on the fly is that when an HCI object is created from bytes, the `parameters` byte array is already known, so it can the stored as-is, and therefore doesn't have to be re-serialized when the same object is converted back to bytes. When an HCI object is created from fields, `parameters` is obtained by serialization. Since it is almost always the case that an HCI object will be either created from bytes, or serialized to bytes, or both (for example in the HCI bridge case), this sort of works well. It could be improved by only computing the `parameters` bytes lazily when the object is converted to bytes, in case someone has a use for an HCI object that is created from fields and never serialized

bumble/hci.py
```diff
@@ -6876,12 +6795,15 @@ class HCI_LE_CS_Test_End_Complete_Event(HCI_LE_Meta_Event):
...
 # -----------------------------------------------------------------------------
-@HCI_Event.event([('status', STATUS_SPEC)])
+@HCI_Event.dataclass_event
+@dataclasses.dataclass
 class HCI_Inquiry_Complete_Event(HCI_Event):
     '''
     See Bluetooth spec @ 7.7.1 Inquiry Complete Event
     '''
 
+    status: int = field(metadata=metadata(STATUS_SPEC))
```
barbibulle, line 6808: we could define `STATUS_METADATA=metadata(STATUS_SPEC)`

barbibulle (approved):

bumble/hci.py
zxzxwu: Indeed, parameters should be evaluated statically, but currently if we want to keep both legacy and class-based packet constructors, there isn't a simple way to keep both working as is, since __post_init__ in a non-dataclass are ignored in derived dataclasses.

When finally all classes have become dataclasses, things will be simpler. Here we can set a functools.cache property to prevent repeated serialization.

bumble/hci.py
zxzxwu: I think this could be merged with `SpecableEnum` later? Since it's not too long here.

Finally we may replace all `*_SPEC` with `SpecableEnum`.

bumble/hci.py

PR #711, opened 2025-06-16
Migrate all HCI_Extended_Event to dataclasses

zxzxwu:
* Make all enums in hci spectable
* Rename HCI_Extended_Event.dataclass_event to HCI_Extended_Event.event and remove the old one
* Minor changes in vendor/android/hci.py

### DIFF
--- a/bumble/hci.py
+++ b/bumble/hci.py
@@ -108,6 +108,17 @@ def phy_list_to_bits(phys: Optional[Iterable[Phy]]) -> int:
     return phy_bits
 
 
+class SpecableEnum(utils.OpenIntEnum):
+
+    @classmethod
+    def type_spec(cls, size: int):
+        return {'size': size, 'mapper': lambda x: cls(x).name}
+
+    @classmethod
+    def type_metadata(cls, size: int, list_begin: bool = False, list_end: bool = False):
+        return metadata(cls.type_spec(size))
+
+
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


bumble/hci.py
```diff
@@ -108,6 +108,17 @@ def phy_list_to_bits(phys: Optional[Iterable[Phy]]) -> int:
     return phy_bits
 
 
+class SpectableEnum(utils.OpenIntEnum):
```
barbibulle: I recommend `Specable` rather than `Spectable` (no `t`)

bumble/hci.py
```diff
@@ -108,6 +108,17 @@ def phy_list_to_bits(phys: Optional[Iterable[Phy]]) -> int:
...
+
+    @classmethod
+    def type_spec(cls, size: int):
+        return {'size': size, 'mapper': lambda x: cls(x).name}
+
+    @classmethod
+    def metadata(cls, size: int, list_begin: bool = False, list_end: bool = False):
```
barbibulle: Maybe `type_metadata` is a safer name, because `metadata` could refer to some other class/object metatada for a class already.

bumble/hci.py
zxzxwu: Oops, that's a typo. Done.

bumble/hci.py
zxzxwu: Done.

PR #718, opened 2025-07-02
Migrate L2CAP packets to dataclasses

zxzxwu:


bumble/l2cap.py
```diff
@@ -232,43 +203,44 @@ def __str__(self) -> str:
...
     '''
 
-    classes: dict[int, type[L2CAP_Control_Frame]] = {}
-    code = 0
-    name: str
+    classes: ClassVar[dict[int, type[L2CAP_Control_Frame]]] = {}
+    fields: ClassVar[hci.Fields] = ()
+    code: int = dataclasses.field(default=0, init=False)
+    name: str = dataclasses.field(default='', init=False)
+    _parameters: Optional[bytes] = dataclasses.field(default=None, init=False)
```
barbibulle: This should be called `data` rather than `parameters`, as per the spec (`parameters` is really a command/event concept).

barbibulle (approved):

bumble/l2cap.py
zxzxwu: Done, renamed `parameters` -> `data`

PR #720, opened 2025-07-02
Update pyproject.toml to python 3.10

adjscent:
Currently there is an issue if bumble sticks to 3.8 

```
 × No solution found when resolving dependencies for split (python_full_version == '3.8.*'    
  │ and sys_platform != 'emscripten'):
  ╰─▶ Because the requested Python version (>=3.8) does not satisfy Python>=3.9.0 and
      pylint==3.3.1 depends on Python>=3.9.0, we can conclude that pylint==3.3.1 cannot be     
      used.
      And because bumble[development] depends on pylint==3.3.1 and your project
      requires bumble[development], we can conclude that your project's requirements are       
      unsatisfiable.

      hint: The `requires-python` value (>=3.8) includes Python versions that are not
      supported by your dependencies (e.g., pylint==3.3.1 only supports >=3.9.0). Consider     
      using a more restrictive `requires-python` value (like >=3.9.0).
```

Updating to 3.10 solves the issue.


pyproject.toml
```diff
@@ -8,7 +8,7 @@ dynamic = ["version"]
 description = "Bluetooth Stack for Apps, Emulation, Test and Experimentation"
 readme = "README.md"
 authors = [{ name = "Google", email = "bumble-dev@google.com" }]
-requires-python = ">=3.8"
+requires-python = ">=3.10"
```
zxzxwu: Our policy is to support all maintained Python versions: https://devguide.python.org/versions/

So currently we still need to support Python 3.9+.

adjscent: Changed to 3.9.  The build will also work with 3.9. I preemptively retired 3.9 myself since it's going to EOL this year.

PR #733, opened 2025-07-29
Fix L2CAP_Control_Frame errors

zxzxwu:
- Correct length calculation
- Rename data -> payload to avoid name collision with fields
- Remove __init__
- Add some tests

### DIFF
--- a/tests/l2cap_test.py
+++ b/tests/l2cap_test.py
@@ -22,12 +22,8 @@ import random
 import pytest
 
 from bumble.core import ProtocolError
-from bumble.l2cap import (
-    L2CAP_Connection_Request,
-    ClassicChannelSpec,
-    LeCreditBasedChannelSpec,
-)
-from .test_utils import TwoDevices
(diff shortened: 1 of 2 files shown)


tests/l2cap_test.py
```diff
@@ -137,6 +145,40 @@ def on_close(which, event):
...
+    channel_manager.send_control_frame(
+        connection,
+        l2cap.L2CAP_LE_SIGNALING_CID,
+        l2cap.L2CAP_Information_Request(
+            identifier=channel_manager.next_identifier(connection),
+            info_type=info_type,
+        ),
+    )
+
+    response = await asyncio.wait_for(info_rsp.get(), timeout=_TIMEOUT)
```
barbibulle: You could use `test_utils.async_barrier` instead of a wait. That barrier will ensure that all async tasks that could run without blocking will have completed.

zxzxwu: Done

PR #734, opened 2025-07-31
Support LE Subrating

khsiao-google:
Handle the following HCI commands / events and make device API for them

1. HCI_LE_Subrate_Change
2. HCI_LE_Set_Default_Subrate
3. HCI_LE_Subrate_Request

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -6208,11 +6208,23 @@ class Device(utils.CompositeEventEmitter):
             f'{connection.peer_address} as {connection.role_name}, '
             f'{connection_parameters}'
         )
-        connection.parameters = Connection.Parameters(
-            connection_parameters.connection_interval * 1.25,
-            connection_parameters.peripheral_latency,
-            connection_parameters.supervision_timeout * 10.0,
-        )
+        if (
+            connection.parameters.connection_interval
+            != connection_parameters.connection_interval * 1.25
+        ):
+            connection.parameters = Connection.Parameters(
+                connection_parameters.connection_interval * 1.25,
+                connection_parameters.peripheral_latency,
+                connection_parameters.supervision_timeout * 10.0,
+            )
+        else:
+            connection.parameters = Connection.Parameters(
+                connection_parameters.connection_interval * 1.25,
+                connection_parameters.peripheral_latency,
+                connection_parameters.supervision_timeout * 10.0,
+                connection.parameters.subrate_factor,
+                connection.parameters.continuation_number,
+            )
         connection.emit(connection.EVENT_CONNECTION_PARAMETERS_UPDATE)
 
     @host_event_handler
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -2066,7 +2066,7 @@ class DeviceConfiguration:
     le_simultaneous_enabled: bool = False
     le_privacy_enabled: bool = False
     le_rpa_timeout: int = DEVICE_DEFAULT_LE_RPA_TIMEOUT
-    le_subrate_enabled: bool = True
+    le_subrate_enabled: bool = False
     classic_enabled: bool = False
     classic_sc_enabled: bool = True
     classic_ssp_enabled: bool = True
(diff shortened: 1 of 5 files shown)


bumble/device.py
```diff
@@ -1787,6 +1789,8 @@ class Parameters:
         connection_interval: float  # Connection interval, in milliseconds. [LE only]
         peripheral_latency: int  # Peripheral latency, in number of intervals. [LE only]
         supervision_timeout: float  # Supervision timeout, in milliseconds.
+        subrate_factor: int = 1
+        continuation_number: int = 0
```
zxzxwu: Maybe provide a reference to spec about why the default is 1 / 0.

bumble/device.py
```diff
@@ -6226,6 +6232,22 @@ def on_connection_phy_update_failure(self, connection, error):
...
+        subrate_factor: int,
+        peripheral_latency: int,
+        continuation_number: int,
+        supervision_timeout: int,
+    ):
+        connection.parameters.subrate_factor = subrate_factor
+        connection.parameters.peripheral_latency = peripheral_latency
+        connection.parameters.continuation_number = continuation_number
+        connection.parameters.supervision_timeout = supervision_timeout * 10
+        connection.emit(connection.EVENT_LE_SUBRATE_CHANGE)
```
zxzxwu, line 6277: In on_connection_parameters_update, connection.parameters is always a new object where subrate info is dropped, so we need to change the pattern there. (BTW, should subrate parameters remain when connection update?)

bumble/hci.py
```diff
@@ -6460,6 +6491,22 @@ class HCI_LE_BIGInfo_Advertising_Report_Event(HCI_LE_Meta_Event):
...
 
+# -----------------------------------------------------------------------------
+@HCI_LE_Meta_Event.event
+@dataclasses.dataclass
+class HCI_LE_Subrate_Change_Event(HCI_LE_Meta_Event):
+    '''
+    See Bluetooth spec @ 7.7.65.35 LE Subrate Change event
+    '''
+
+    status: int = field(metadata=metadata(1))
```
zxzxwu:
```suggestion
    status: int = field(metadata=metadata(STATUS_SPEC))
```

bumble/device.py
```diff
@@ -2410,6 +2415,7 @@ def __init__(
         self.le_privacy_enabled = config.le_privacy_enabled
         self.le_rpa_timeout = config.le_rpa_timeout
         self.le_rpa_periodic_update_task: Optional[asyncio.Task] = None
+        self.le_subrate_enabled = config.le_subrate_enabled
```
zxzxwu, line 2422: You need to send HCI_LE_Set_Host_Feature_Command to enable this feature like https://github.com/google/bumble/blob/main/bumble/device.py#L2785

bumble/device.py
khsiao-google: We can discuss about the pattern on Monday.

As to connection update, in 7.8.18 LE Connection Update command, it states that

If this command completes successfully and the connection interval has changed,
then the subrating factor shall be set to 1 and the continuation number to 0. In this
case, Max_Latency must be interpreted in underlying connection events. Otherwise the
subrating factor and continuation number shall be unchanged and Max_Latency must
be interpreted in subrated events.

Therfore, we can keep it as default.

bumble/device.py
khsiao-google: Added

bumble/hci.py
khsiao-google: Changed.

barbibulle:
The way I understand the spec, I think that when there's a LE Connection Update Complete event, if the subrate factor and continuation number are only reset if there's a change in the connection interval. So if there's no change, the Connection.Parameters object that's created in the `on_connection_parameters_update` method should reuse the existing values for `subrate_factor` and `continuation_number`.

khsiao-google:
> The way I understand the spec, I think that when there's a LE Connection Update Complete event, if the subrate factor and continuation number are only reset if there's a change in the connection interval. So if there's no change, the Connection.Parameters object that's created in the `on_connection_parameters_update` method should reuse the existing values for `subrate_factor` and `continuation_number`.

I think that's a great point. Already modify the on_connection_parameters_update so that it only resets when connection interval changes.

barbibulle (approved):

PR #737, opened 2025-08-02
Update license field to use proper SPDX identifier

leifdreizler:
This changes the license field to be a valid [SPDX identifier](https://spdx.org/licenses) aligning with [PEP 639](https://peps.python.org/pep-0639/#project-source-metadata). This populates the `license_expression` field in the PyPI API and is used by downstream tools including deps.dev

These changes were generated by Claude (sorry it wasn't Gemini) after reviewing the license and manifest files in your repository, but opened and reviewed by me. Please let me know if the analysis is incorrect and thanks for being an OSS maintainer.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,6 +7,7 @@ name = "bumble"
 dynamic = ["version"]
 description = "Bluetooth Stack for Apps, Emulation, Test and Experimentation"
 readme = "README.md"
+license = "Apache-2.0"
 authors = [{ name = "Google", email = "bumble-dev@google.com" }]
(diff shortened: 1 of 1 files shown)


barbibulle:
Thanks for the PR. Shouldn’t `pyproject.toml` also include a `license-files` entry as well, pointing to `LICENSE?

leifdreizler:
Yes, happy to make that change! ✅

I have only been making that when folks had the older `license = {file = "LICENSE"}` file sub key to minimize my changes, but perhaps I should just add it 🤔

PR #758, opened 2025-08-17
Sorting VID/PIDs and adding observed values on ZEXMTE

XenoKovah:
Sorting VID/PIDs and adding observed values on ZEXMTE (https://zexmtebluetooth.com/#Products) devices


XenoKovah:
Nevermind, CLA BS not working

PR #760, opened 2025-08-18
Enhance transports

zxzxwu:
* Support IPv6 schema
* Add transport integration tests
* Add UNIX socket server

### DIFF
--- a/bumble/transport/android_emulator.py
+++ b/bumble/transport/android_emulator.py
@@ -77,21 +77,17 @@ async def open_android_emulator_transport(spec: Optional[str]) -> Transport:
 
     # Parse the parameters
     mode = 'host'
-    server_host = 'localhost'
-    server_port = '8554'
+    server_address = 'localhost:8554'
     if spec:
         params = spec.split(',')
         for param in params:
             if param.startswith('mode='):
                 mode = param.split('=')[1]
-            elif ':' in param:
-                server_host, server_port = param.split(':')
             else:
-                raise TransportSpecError('invalid parameter')
+                server_address = param
 
     # Connect to the gRPC server
-    server_address = f'{server_host}:{server_port}'
-    logger.debug(f'connecting to gRPC server at {server_address}')
+    logger.debug('connecting to gRPC server at %s', server_address)
     channel = grpc.aio.insecure_channel(server_address)
 
     service: Union[EmulatedBluetoothServiceStub, VhciForwardingServiceStub]
--- a/bumble/transport/__init__.py
+++ b/bumble/transport/__init__.py
@@ -185,11 +185,17 @@ async def _open_transport(scheme: str, spec: Optional[str]) -> Transport:
 
         return await open_android_netsim_transport(spec)
 
-    if scheme == 'unix':
+    if scheme in ('unix', 'unix-client'):
         from bumble.transport.unix import open_unix_client_transport
 
         assert spec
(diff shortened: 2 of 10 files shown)


bumble/transport/android_emulator.py
```diff
@@ -85,7 +85,7 @@ async def write(self, packet):
             if param.startswith('mode='):
                 mode = param.split('=')[1]
             elif ':' in param:
-                server_host, server_port = param.split(':')
+                server_host, server_port = param.rsplit(':', maxsplit=1)
```
barbibulle: To properly support IPv6 addresses, we need a more robust parser to separate the host and port. (`':' in param` isn't sufficient, since you could have an IPv6 address that has a `:`, but not followed by `:<port>`). I think we should factor out a shared function (maybe in `common.py`?) that knows how to parse IPv6 addresses with ports of the form `[address]:port` like `[2001:db8:85a3:8d3:1319:8a2e:370:7348]:8000` (i.e enclosing the address in `[]` is mandatory when followed by a port number).

zxzxwu: Though it's not very complicated to do this, I am wondering if that's necessary - because we don't allow any schema without a complete `address:port` pair, so even `if params and ':' in params[0]` returns a false positive, it will still fail later such as

```
# python apps/controller_info.py android-netsim::::

  File "/usr/local/google/home/joshwu/projects/bumble/bumble/transport/android_netsim.py", line 447, in open_android_netsim_transport
    port = int(port_str)
ValueError: invalid literal for int() with base 10: ''

# or
grpc.aio._call.AioRpcError: <AioRpcError of RPC that terminated with:
        status = StatusCode.UNAVAILABLE
        details = "DNS resolution failed for :::1: C-ares status is not ARES_SUCCESS qtype=AAAA name=:::1 is_balancer=0: Domain name not found"
        debug_error_string = "UNKNOWN:Error received from peer  {grpc_status:14, grpc_message:"DNS resolution failed for :::1: C-ares status is not ARES_SUCCESS qtype=AAAA name=:::1 is_balancer=0: Domain name not found"}"
>
```

Also this might violate metadata schema defined in open_transport
```py
    scheme, *tail = name.split(':', 1)
    spec = tail[0] if tail else None
    metadata = None
    if spec:
        # Metadata may precede the spec
        if spec.startswith('['):
            metadata_str, *tail = spec[1:].split(']')
            spec = tail[0] if tail else None
            metadata = dict([entry.split('=') for entry in metadata_str.split(',')])

```

zxzxwu: Oh wait, I found this section is totally not neceassary, since the grpc module directly parses the port, so we can remove it here.

For other transports, I think we should keep the current schema to maintain the compatibility with transport metadata?

barbibulle: You're right, we always have the port explicitly specified, so it's fine to split out the portion after the last `:`, and not require `[]` around IPv6 addresses. This will keep the metadata parser simple (if we needed, we could differentiate `[foo=bar]` metadata from `[xx:yy:zz] IPv6`, but we don't need to here)

barbibulle (approved):

bumble/transport/android_emulator.py
zxzxwu: Still, I found gRPC cannot parse such address, so as you said, we can heuristically distinguish between metadata and IPv6 address (in another PR).

PR #768, opened 2025-08-29
add support for data type classes

barbibulle:
This adds a class-based approach to advertising data (and other "data types"). The old ad-hoc model is still supported, but this is a cleaner approach with better type safety.
The main class `DataType` is defined in `core.py` but the subclasses are defined in `data_types.py`, to avoid circular dependencies.

### DIFF
--- a/bumble/data_types.py
+++ b/bumble/data_types.py
@@ -25,7 +25,7 @@ from __future__ import annotations
 import dataclasses
 import math
 import struct
-from typing import Any, ClassVar, Literal, Optional, TypeVar, Union, overload
+from typing import Any, ClassVar, Sequence
 
 from typing_extensions import Self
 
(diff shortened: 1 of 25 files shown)


bumble/data_types.py
```diff
@@ -0,0 +1,1018 @@
...
+            and self.ad_data == other.ad_data
+        )
+
+
+@dataclasses.dataclass
+class ListOfServiceUUIDs(core.DataType):
+    """Base class for complete or incomplete lists of UUIDs."""
+
+    _uuid_size: ClassVar[int] = 0
+    uuids: list[core.UUID]
```
zxzxwu: Sequence could be better

zxzxwu (approved):
This is awesome!

PR #769, opened 2025-09-01
Add typing for host.py

khsiao-google:


bumble/hci.py
```diff
@@ -6468,7 +6471,10 @@ class HCI_LE_BIG_Sync_Established_Event(HCI_LE_Meta_Event):
     irc: int = field(metadata=metadata(1))
     max_pdu: int = field(metadata=metadata(2))
     iso_interval: int = field(metadata=metadata(2))
-    connection_handle: int = field(metadata=metadata(2, list_begin=True, list_end=True))
+    num_bis: int = field(metadata=metadata(1))
```
zxzxwu: This is not corrrect. When list metadata is applied, the first byte will be implicitly considered the length of list.

If you need num_bis, just take `len(connection_handle)`.

bumble/host.py
```diff
@@ -1006,15 +1012,43 @@ def on_hci_le_connection_complete_event(self, event):
...
+            event.connection_handle,
+            event.role,
+            event.peer_address_type,
+            event.peer_address,
+            event.connection_interval,
+            event.peripheral_latency,
+            event.supervision_timeout,
+            event.central_clock_accuracy,
+        )
+        self.on_hci_le_connection_complete_event(le_connection_complete_event)
```
zxzxwu: Not good. We can just simply let `on_hci_le_connection_complete_event` take both types of events.

bumble/host.py
```diff
@@ -1006,15 +1012,43 @@ def on_hci_le_connection_complete_event(self, event):
...
```
zxzxwu: ditto

bumble/host.py
```diff
@@ -1182,10 +1232,12 @@ def on_hci_le_cis_request_event(self, event):
...
         )
 
-    def on_hci_le_create_big_complete_event(self, event):
+    def on_hci_le_create_big_complete_event(
+        self, event: hci.HCI_LE_Create_BIG_Complete_Event
+    ):
         self.bigs[event.big_handle] = set(event.connection_handle)
         if self.iso_packet_queue is None:
-            logger.warning("BIS established but ISO packets not supported")
+            raise ValueError("BIG established but ISO packets not supported")
```
zxzxwu: Avoid using primitive errors. Use errors defined in `bumble.core` instead.

bumble/host.py
```diff
@@ -1208,8 +1260,13 @@ def on_hci_le_create_big_complete_event(self, event):
             event.iso_interval,
         )
 
-    def on_hci_le_big_sync_established_event(self, event):
+    def on_hci_le_big_sync_established_event(
+        self, event: hci.HCI_LE_BIG_Sync_Established_Event
+    ):
         self.bigs[event.big_handle] = set(event.connection_handle)
+        if self.iso_packet_queue is None:
+            raise ValueError("BIS established but ISO packets not supported")
```
zxzxwu: ditto

bumble/host.py
```diff
@@ -1263,11 +1326,11 @@ def on_hci_le_periodic_advertising_sync_transfer_received_v2_event(self, event):
             event.advertiser_clock_accuracy,
         )
 
-    def on_hci_le_cis_established_event(self, event):
+    def on_hci_le_cis_established_event(self, event: hci.HCI_LE_CIS_Established_Event):
         # The remaining parameters are unused for now.
         if event.status == hci.HCI_SUCCESS:
             if self.iso_packet_queue is None:
-                logger.warning("CIS established but ISO packets not supported")
+                raise ValueError("CIS established but ISO packets not supported")
```
zxzxwu: ditto

bumble/hci.py
khsiao-google: I don't quite understand though.
1. Shouldn't we save the exact parameters as the spec mentioned? Wouldn't it be more clear?
2. Should I not use the sequence?

bumble/host.py
khsiao-google: Is InvalidPacketError suitable for this scenario?

bumble/hci.py
zxzxwu: > Shouldn't we save the exact parameters as the spec mentioned? Wouldn't it be more clear?

Actually that becomes more unclear in an object-oriented language.

The pattern of a single byte N leading following N items is very common in the Bluetooth spec, and the reason why there is such a pattern is that the spec is appied for a serialized bytes sequence. However, in an OO language, most linear containers provides a convenient way to tell the length of containers, and then it becomes ambiguous to have an independent length parameter: Should the independent parameter be respected, or the length of containers?
Since we can always infer it from `len(seq)`, we choose to hide this information in the Python abstraction, and this shouldn't impact the clarity, because for most of time people don't want to pass invalid packets to controllers.

If you observe Android PDL, you will find that they take the similar mechanism: https://cs.android.com/android/platform/superproject/main/+/main:packages/modules/Bluetooth/system/pdl/hci/hci_packets.pdl;drc=373030b9ea870e79cf1fe231f722df0451c0c10e;l=6685

> Should I not use the sequence?

Yes, you can still do this.

bumble/host.py
zxzxwu: InvalidStateError should be proper.

PR #771, opened 2025-09-03
Replace core.ConnectionParameters by Connection.Parameters in device.py

khsiao-google:


bumble/device.py
```diff
@@ -1833,34 +1833,28 @@ def __init__(
...
             PhysicalTransport.BR_EDR,
             device.public_address,
             None,
-            peer_address,
-            None,
-            role,
+            pending_connection.peer_address,
             None,
+            pending_connection.role,
+            parameters,
```
zxzxwu: Can we also pass keywords here? There are too many args.

bumble/device.py
```diff
@@ -2341,6 +2335,15 @@ def from_config_file_with_hci(
...
         return cls.from_config_with_hci(config, hci_source, hci_sink)
 
+    @dataclass
+    class _PendingConnection:
+        peer_address: hci.Address
+        role: hci.Role
+
+        def __init__(self, peer_address: hci.Address, role: hci.Role):
+            self.peer_address = peer_address
+            self.role = role
```
zxzxwu: Shouldn't create `__init__` for dataclass.

bumble/device.py
```diff
@@ -1833,34 +1833,28 @@ def __init__(
 
     # [Classic only]
     @classmethod
-    def incomplete(cls, device, peer_address, role):
+    def create_from_pending_connection(
```
zxzxwu: maybe `from_pending_connection` is enough.

bumble/device.py
```diff
@@ -6183,7 +6191,7 @@ def on_connection_encryption_key_refresh(self, connection: Connection):
     @host_event_handler
     @with_connection_from_handle
     def on_connection_parameters_update(
-        self, connection: Connection, connection_parameters: core.ConnectionParameters
+        self, connection: Connection, connection_parameters: Connection.Parameters
```
zxzxwu: Incorrect

bumble/device.py
```diff
@@ -1833,34 +1833,28 @@ def __init__(
...
     # [Classic only]
     @classmethod
-    def incomplete(cls, device, peer_address, role):
+    def from_pending_connection(
+        cls,
+        device: Device,
+        pending_connection: Device._PendingConnection,
+        handle: int,
+        parameters: Parameters,
+    ):
```
barbibulle: The return type is missing

bumble/device.py
```diff
@@ -1833,34 +1833,28 @@ def __init__(
...
         """
-        Instantiate an incomplete connection (ie. one waiting for a HCI Connection
-        Complete event).
-        Once received it shall be completed using the `.complete` method.
+        Instantiate an connection.
```
barbibulle: typo: an -> a
Maybe also be more descriptive. Like "Instantiate a Connection from a _PendingConnection and its now-resolved parameters"

bumble/device.py
```diff
@@ -5457,9 +5456,16 @@ def on_connection(
         self_resolvable_address: Optional[hci.Address],
         peer_resolvable_address: Optional[hci.Address],
         role: hci.Role,
-        connection_parameters: Optional[core.ConnectionParameters],
+        connection_interval: int,
+        peripheral_latency: int,
+        supervision_timeout: int,
     ) -> None:
         # Convert all-zeros addresses into None.
```
barbibulle, line 5467: this comment should be below, just before line 5469

bumble/device.py
```diff
@@ -5457,9 +5456,16 @@ def on_connection(
...
+        connection_parameters = Connection.Parameters(
```
barbibulle: this code could be moved below, where `Connection.from_pending_connection` is called (is isn't used anywhere else). You could even skip the `connection_parameters=` local assignment, and just call the constructor inline, since it isn't referenced anywhere else.

bumble/device.py
```diff
@@ -5457,9 +5456,16 @@ def on_connection(
         self_resolvable_address: Optional[hci.Address],
         peer_resolvable_address: Optional[hci.Address],
         role: hci.Role,
```
barbibulle, line 5462: It looks like `role` should be `Optional[hci.Role]`, since for incomplete connections, the role isn't yet known and the event will be emitted with `None` as the `role` parameter.

bumble/host.py
```diff
@@ -1068,6 +1060,8 @@ def on_hci_connection_complete_event(
                 None,
                 None,
                 None,
```
barbibulle: The last 3 parameters should be `0, 0, 0` instead of `None, None, None` since they are not declared as `Optional`.

bumble/device.py
zxzxwu: I think we can separate host connection event into LE and Classic, since they are getting more diverged and have different handling logic.

bumble/device.py
```diff
@@ -5449,15 +5411,58 @@ def _complete_le_extended_advertising_connection(
...
-    def on_connection(
+    def on_connection_complete(
+        self,
+        connection_handle: int,
+        peer_address: hci.Address,
+        connection_interval: int,
+        peripheral_latency: int,
+        supervision_timeout: int,
+    ) -> None:
+        connection_role = self.connection_roles.pop(peer_address)
```
zxzxwu: Still safer to have a fallback
```suggestion
        connection_role = self.connection_roles.pop(peer_address, hci.Role.PERIPHERAL)
```

bumble/device.py
```diff
@@ -5449,15 +5411,58 @@ def _complete_le_extended_advertising_connection(
...
+                connection_interval * 1.25,
+                peripheral_latency,
+                supervision_timeout * 10.0,
+            ),
+        )
+        self.connections[connection_handle] = connection
+
+        self.emit(self.EVENT_CONNECTION, connection)
+
+        return
```
zxzxwu: Meaningless return

bumble/device.py
```diff
@@ -6183,27 +6173,27 @@ def on_connection_encryption_key_refresh(self, connection: Connection):
     @host_event_handler
     @with_connection_from_handle
     def on_connection_parameters_update(
-        self, connection: Connection, connection_parameters: core.ConnectionParameters
+        self,
+        connection: Connection,
+        connection_interval,
+        peripheral_latency,
+        supervision_timeout,
```
zxzxwu:
```suggestion
        connection_interval: int,
        peripheral_latency: int,
        supervision_timeout: int,
```

bumble/device.py
```diff
@@ -2367,7 +2333,7 @@ def __init__(
         self.le_connecting = False
         self.disconnecting = False
         self.connections = {}  # Connections, by connection handle
-        self.pending_connections = {}  # Connections, by BD address (BR/EDR only)
+        self.connection_roles = {}  # Connections, by BD address (BR/EDR only)
```
zxzxwu:
```suggestion
        self.connection_roles = {}  # Local connection roles, by BD address (BR/EDR only)
```

bumble/device.py
khsiao-google: There are another place with self.connection_roles.pop(peer_address, None). Should I also make it self.connection_roles.pop(peer_address, hci.Role.PERIPHERAL)?

bumble/host.py
```diff
@@ -1060,14 +1051,12 @@ def on_hci_connection_complete_event(
 
             # Notify the client
             self.emit(
-                'connection',
+                'connection_complete',
```
zxzxwu, line 1054: I think we prefer using names of 'classic_connection' and 'le_connection'? At least we should be precisely classic/le_connection(_complete).

@barbibulle

zxzxwu (approved):

bumble/host.py
barbibulle: Yes. The naming rule for events is that even names are nouns, not verbs or adjectives. So `connection` is the right choice. If we need two events, one for LE and one for Classic, that would be `le_connection` and `classic_connection`.

zxzxwu: @khsiao-google Please address this comment in another PR.

PR #772, opened 2025-09-04
HAP: Slightly Pythonic refactor

zxzxwu:
* Add missing type annotations
* Avoid `__value__` and `_` arg names (this will be a problem for override).
* Replace while-pop with for loop


zxzxwu:
cc @wescande

bumble/profiles/hap.py
```diff
@@ -447,16 +448,16 @@ def get_op_index(op: PresetChangedOperation) -> int:
...
+        for i, op in enumerate(op_list):
             try:
                 await connection.device.indicate_subscriber(
                     connection,
                     self.hearing_aid_preset_control_point,
-                    value=op_list[0].to_bytes(len(op_list) == 1),
+                    value=op.to_bytes(i == len(op_list) - 1),
                 )
-                # Remove item once sent, and keep the non sent item in the list
-                op_list.pop(0)
```
wescande: Why don't we need to remove it from the list anymore ?

My understanding is that this list is sent anytime the device reconnect, and if we keep it in the list, connecting twice will send the list of event twice too, unless we remove the event when the indication got acknowledge (hence why we don't clear all at once, but only one at a time)

bumble/profiles/hap.py
```diff
@@ -532,48 +534,51 @@ async def set_active_preset(self, value: bytes) -> None:
...
-            break
-
-        if not first_preset:  # If no other preset are available
+        records = sorted(
+            [
+                record
+                for record in self.preset_records.values()
+                if record.is_available()
+            ],
+            key=lambda record: record.index,
```
wescande, line 555: And now I wonder if the initial code was working as expected. Because we don't give a key to sort, it only iterate on inserting order in the dict, which make it rely on whatever initialize the server.

"Worked in my code", but that is not as nice as what you are doing here

bumble/profiles/hap.py
zxzxwu: I didn't notice that, so indeed we still need a queue for that.

bumble/profiles/hap.py
zxzxwu: Not really. Presets are still sorted with the index. It also passes test_set_next_preset_will_loop_to_first.

wescande (approved):

khsiao-google (approved):

PR #773, opened 2025-09-05
HAP: wait for MTU to process reconnection event

wescande:
When HAP reconnect, it sends indication of all events that happen during the disconnection.
But it should wait for the profile to be ready and for the MTU to have been negotiated or else the remote may not be ready yet.

As a side effect of this, the current GattServer doesn't re-populate the handle of subscriber during a reconnection, we have to bypass this check to send the notification

### DIFF
--- a/bumble/profiles/hap.py
+++ b/bumble/profiles/hap.py
@@ -277,8 +277,10 @@ class HearingAccessService(gatt.TemplateService):
             def on_pairing(*_: Any) -> None:
                 self.on_incoming_paired_connection(connection)
 
-            if connection.peer_resolvable_address:
-                self.on_incoming_paired_connection(connection)
+            @connection.on(connection.EVENT_CONNECTION_ATT_MTU_UPDATE)
+            def on_mtu_update(*_: Any) -> None:
+                if connection.peer_resolvable_address:
+                    self.on_incoming_paired_connection(connection)
 
         self.hearing_aid_features_characteristic = gatt.Characteristic(
             uuid=gatt.GATT_HEARING_AID_FEATURES_CHARACTERISTIC,
@@ -457,6 +459,7 @@ class HearingAccessService(gatt.TemplateService):
                     connection,
                     self.hearing_aid_preset_control_point,
                     value=op_list[0].to_bytes(len(op_list) == 1),
+                    force=True, # Gatt server forget who was subscribed to notification
                 )
                 # Remove item once sent, and keep the non sent item in the list
                 op_list.pop(0)


bumble/profiles/hap.py
```diff
@@ -457,6 +459,7 @@ def get_op_index(op: PresetChangedOperation) -> int:
                     connection,
                     self.hearing_aid_preset_control_point,
                     value=op_list[0].to_bytes(len(op_list) == 1),
+                    force=True, # Gatt server forget who was subscribed to notification
```
barbibulle: this will work, but isn't completely compliant. A GATT server is supposed to cache the CCCD descriptor values across disconnections of bonded clients. I see now that the GATT server implementation doesn't do that (all client configurations are lost on disconnection). I will file a bug to fix that in a future release.

bumble/profiles/hap.py
```diff
@@ -277,8 +277,10 @@ def on_disconnection(_reason) -> None:
             def on_pairing(*_: Any) -> None:
                 self.on_incoming_paired_connection(connection)
 
-            if connection.peer_resolvable_address:
-                self.on_incoming_paired_connection(connection)
+            @connection.on(connection.EVENT_CONNECTION_ATT_MTU_UPDATE)
+            def on_mtu_update(*_: Any) -> None:
+                if connection.peer_resolvable_address:
+                    self.on_incoming_paired_connection(connection)
```
barbibulle: shouldn't this also wait for the connection to be encrypted?

bumble/profiles/hap.py
wescande: Do you prefer I wait on the GATT to be fixed ? Or would this temporary bypass is fine for you ?

barbibulle: No need to wait, this will work fine as it is for now.

zxzxwu: Submitting a workaround is fine, but there had better to be a TODO comment, and later when we fix GATT server we will revert the workaround.

bumble/profiles/hap.py
wescande: You are right, Spec mention that indication should be sent only when the encryption is done.

I am not sure what is the better way of doing this with bumble. I updated the patch to us on_encryption_change event.

Let me know if you see a better way

bumble/profiles/hap.py
wescande: Comment updated to have a TODO that mention GATT notification subscription not being persistent

bumble/profiles/hap.py
barbibulle: Thats looks fine. 
But the `on_pairing` handler above should probably be removed, because when you're pairing, you'll also get an `on_encryption_change` callback. If you're connecting and were already previously paired, you'll also get just `on_encryption_change`, but no `on_pairing`. It's probably best to not end up calling `on_incoming_paired_connection` twice in the case where you are pairing.

wescande: I made a couple change to try to have a working scenario while implementing your comment.

I need `on_pairing`, because this event can be sent after the `on_encryption_change`, and we need to compare address between reconnection. When there is no resolvable address, it consider it's a new device and the history of preset changed is dropped.

We also should check the MTU as we don't want to truncate the content of the notification.

So I made multiples listener on all the event, and I made sure to make the method safe to be called multiples times.

Hopefully that should handle all the cases

PR #775, opened 2025-09-10
Remove the word 'complete' from function name

khsiao-google:
Also remove le only parameters when calling classic connection.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -1782,15 +1782,22 @@ class Connection(utils.CompositeEventEmitter):
 
     @dataclass
     class Parameters:
-        connection_interval: float  # Connection interval, in milliseconds. [LE only]
-        peripheral_latency: int  # Peripheral latency, in number of intervals. [LE only]
-        supervision_timeout: float  # Supervision timeout, in milliseconds.
-        subrate_factor: int = (
-            1  # See Bluetooth spec Vol 6, Part B - 4.5.1 Connection events
-        )
-        continuation_number: int = (
-            0  # See Bluetooth spec Vol 6, Part B - 4.5.1 Connection events
-        )
+        """
+        LE connection parameters.
+
+        Attributes:
+          connection_interval: Connection interval, in milliseconds.
+          peripheral_latency: Peripheral latency, in number of intervals.
+          supervision_timeout: Supervision timeout, in milliseconds.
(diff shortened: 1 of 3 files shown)


bumble/host.py
```diff
@@ -1051,7 +1051,7 @@ def on_hci_connection_complete_event(
 
             # Notify the client
             self.emit(
-                'connection_complete',
+                'connection',
                 event.connection_handle,
                 event.bd_addr,
                 0,
```
zxzxwu: We should remove unused parameters here.

bumble/host.py
```diff
@@ -1051,7 +1051,7 @@ def on_hci_connection_complete_event(
 
             # Notify the client
             self.emit(
-                'connection_complete',
+                'connection',
```
zxzxwu: classic_connection

bumble/device.py
```diff
@@ -1781,9 +1781,9 @@ def on_connection_encryption_key_refresh(self):
             pass
 
     @dataclass
-    class Parameters:
-        connection_interval: float  # Connection interval, in milliseconds. [LE only]
-        peripheral_latency: int  # Peripheral latency, in number of intervals. [LE only]
+    class Parameters:  # LE only
```
zxzxwu: I think we should use docstring.
```suggestion
    class Parameters:
      '''LE-only connection parameters.
      
      Attributes:
        connection_interval: ...
        peripheral_latency: ...
      '''
```

zxzxwu (approved):

bumble/device.py
```diff
@@ -1782,15 +1782,22 @@ def on_connection_encryption_key_refresh(self):
...
-        peripheral_latency: int  # Peripheral latency, in number of intervals. [LE only]
-        supervision_timeout: float  # Supervision timeout, in milliseconds.
-        subrate_factor: int = (
-            1  # See Bluetooth spec Vol 6, Part B - 4.5.1 Connection events
-        )
-        continuation_number: int = (
-            0  # See Bluetooth spec Vol 6, Part B - 4.5.1 Connection events
-        )
+        """
+        LE only connection parameters.
```
barbibulle: I suggest just saying "LE connection parameters".

.coverage
zxzxwu, line 1: Do not add this

bumble/device.py
zxzxwu: @khsiao-google I think Gilles means to change the description?

PR #792, opened 2025-09-26
Fix - deprecated warning for datetime.utcnow() with Python >= 3.12

markusjellitsch:
Replaced `datetime.utcnow()` with `datetime.now(tz=datetime.timezone.utc)` to fix following deprecated warning:

` Use timezone-aware objects to represent datetimes in UTC; e.g. by calling .now(datetime.timezone.utc)`

This warning can be very annoying when running pytest tests with enabled HCI snooping.

### DIFF
--- a/bumble/snoop.py
+++ b/bumble/snoop.py
@@ -64,7 +64,7 @@ class BtSnooper(Snooper):
     Snooper that saves HCI packets using the BTSnoop format, based on RFC 1761.
     """
 
-    IDENTIFICATION_PATTERN = b'btsnoop\0'
+    IDENTIFICATION_PATTERN = b"btsnoop\0"
     TIMESTAMP_ANCHOR = datetime.datetime(2000, 1, 1)
     TIMESTAMP_DELTA = 0x00E03AB44A676000
     ONE_MS = datetime.timedelta(microseconds=1)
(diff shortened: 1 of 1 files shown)


bumble/snoop.py
```diff
@@ -129,7 +135,7 @@ def create_snooper(spec: str) -> Generator[Snooper, None, None]:
           records will be written to that file if it can be opened/created.
           The keyword args that may be referenced by the string pattern are:
             now: the value of `datetime.now()`
-            utcnow: the value of `datetime.utcnow()`
+            utcnow: the value of 'datetime.now(tz=datetime.timezone.utc)'
```
barbibulle: It would be best to keep the backquotes here instead of regular quotes.

barbibulle (changes requested):

bumble/snoop.py
```diff
@@ -85,7 +85,13 @@ def snoop(self, hci_packet: bytes, direction: Snooper.Direction) -> None:
 
         # Compute the current timestamp
         timestamp = (
-            int((datetime.datetime.utcnow() - self.TIMESTAMP_ANCHOR) / self.ONE_MS)
+            int(
+                (
+                    datetime.datetime.now(tz=datetime.timezone.utc)
+                    - self.TIMESTAMP_ANCHOR
```
barbibulle, line 91: I think that `TIMESTAMP_ANCHOR` would need to be updated as well, to be a timestamp *with* an explicit UTC timezone.

markusjellitsch: update

> I think that `TIMESTAMP_ANCHOR` would need to be updated as well, to be a timestamp _with_ an explicit UTC timezone.

updated

bumble/snoop.py
markusjellitsch: updated

PR #798, opened 2025-10-20
HFP: Change configuration attribute types to Sequence

khsiao-google:


bumble/hfp.py
```diff
@@ -716,7 +716,7 @@ class HfLoopTermination(HfpProtocolError):
         """Termination signal for run() loop."""
 
     supported_hf_features: int
-    supported_audio_codecs: list[AudioCodec]
+    supported_audio_codecs: collections.abc.Sequence[AudioCodec]
```
zxzxwu: If we don't want to use the raw reference directly, annotating the actual implementation (list, as copied) should be better.

bumble/hfp.py
```diff
@@ -716,7 +716,7 @@ class HfLoopTermination(HfpProtocolError):
         """Termination signal for run() loop."""
 
     supported_hf_features: int
-    supported_audio_codecs: list[AudioCodec]
+    supported_audio_codecs: Sequence[AudioCodec]
```
zxzxwu: ditto

PR #801, opened 2025-10-21
feat(intel): clarify firmware/DDC flow and preserve driver metadata

zyanwu-google:
- Add explanatory comments across intel driver to clarify metadata parsing.
- Ensure driver selection preserves runtime options (e.g. "intel/ddc_override:AABB")
  so driver-specific metadata is passed through to the host and available to
  drivers via host.hci_metadata.
- Ensure transport parsing regex and metadata extraction so transport/source
  metadata is populated and visible to drivers.
- Example usage: passing [driver=intel/ddc_override:AABB] will be preserved and
  can be consumed by the Intel driver to apply a DDC override blob.

### DIFF
--- a/bumble/drivers/__init__.py
+++ b/bumble/drivers/__init__.py
@@ -49,6 +49,10 @@ async def get_driver_for_host(host: Host) -> Optional[Driver]:
     driver_classes: dict[str, type[Driver]] = {"rtk": rtk.Driver, "intel": intel.Driver}
     probe_list: Iterable[str]
     if driver_name := host.hci_metadata.get("driver"):
+        # The "driver" metadata may include runtime options after a '/' (for example
+        # "intel/ddc=..."). Keep only the base driver name (the portion before the
+        # first slash) so it matches a key in driver_classes (e.g. "intel").
+        driver_name = driver_name.split("/")[0]
         # Only probe a single driver
         probe_list = [driver_name]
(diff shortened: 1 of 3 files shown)


bumble/drivers/__init__.py
```diff
@@ -49,6 +49,10 @@ async def get_driver_for_host(host: Host) -> Optional[Driver]:
     driver_classes: dict[str, type[Driver]] = {"rtk": rtk.Driver, "intel": intel.Driver}
     probe_list: Iterable[str]
     if driver_name := host.hci_metadata.get("driver"):
+        # The "driver" metadata may include runtime options after a '/' (for example
+        # "intel/ddc=..."). Keep only the base driver name (the portion before the
+        # first slash) so it matches a key in driver_classes (e.g. "intel").
+        driver_name = driver_name.split("/")[0]
```
zxzxwu, line 55: Then what's the purpose to pass the remaining part like `/ddc`? Cannot we just not to pass them?

zyanwu-google: This is what intel driver format already expects in its init function. Please see https://github.com/google/bumble/blob/main/bumble/drivers/intel.py#L359

zyanwu-google: Please re-open if you hav further question.

PR #804, opened 2025-10-23
Fixed bug where it's possible for rfcomm tx_credit to go negative resulting in l2cap disconnect from peripheral

graynode:
While conducting large data transfers over rfcomm I found our device would disconnect after 3-400kb. Looking deeper I found it was possible for the tx credit to go negative which is what was causing the l2cap connection to be disconnected by the other device as this violates the standard. 

see attached HCI log
[bumble-rfcomm-hci-negative-tx-credit.txt](https://github.com/user-attachments/files/23108172/bumble-rfcomm-hci-negative-tx-credit.txt)

### DIFF
--- a/bumble/rfcomm.py
+++ b/bumble/rfcomm.py
@@ -671,15 +671,17 @@ class DLC(utils.EventEmitter):
     def process_tx(self) -> None:
         # Send anything we can (or an empty frame if we need to send rx credits)
         rx_credits_needed = self.rx_credits_needed()
-        while (self.tx_buffer and self.tx_credits > 0) or (
-            rx_credits_needed > 0 and self.tx_credit > 0
-        ):
+        while (self.tx_buffer and self.tx_credits > 0) or rx_credits_needed > 0:
             # Get the next chunk, up to MTU size
             if rx_credits_needed > 0:
-                chunk = bytes([rx_credits_needed]) + self.tx_buffer[: self.mtu - 1]
-                self.tx_buffer = self.tx_buffer[len(chunk) - 1 :]
+                chunk = bytes([rx_credits_needed])
                 self.rx_credits += rx_credits_needed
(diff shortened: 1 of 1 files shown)


bumble/rfcomm.py
```diff
@@ -671,7 +671,7 @@ def rx_credits_needed(self) -> int:
     def process_tx(self) -> None:
         # Send anything we can (or an empty frame if we need to send rx credits)
         rx_credits_needed = self.rx_credits_needed()
-        while (self.tx_buffer and self.tx_credits > 0) or rx_credits_needed > 0:
+        while (self.tx_buffer and self.tx_credits > 0) or (rx_credits_needed > 0 and self.tx_credit > 0):
```
barbibulle: This fix does indeed prevent `tx_credit` from going negative (in the case where `rx_credits_needed` is not 0 but `tx_credit is 0`), but it goes a bit too far: it prevents sending rx credits in that case, even though it may be desirable to do so.
I think the right fix here is to keep the `while` condition as it was (i.e "either there's some data that can be sent, or there's a need to send credits), but then in the `if rx_credits_needed > 0` case, only include data and set `tx_credit_spent` if we have tx credits. Something like this maybe:
```
if rx_credits_needed > 0:
    # Send rx credits
    chunk = bytes([rx_credits_needed])
    self.rx_credits += rx_credits_needed
    if self.tx_buffer and self.tx_credits > 0:
        # Send some data along with the credits
        chunk += self.tx_buffer[: self.mtu - 1]
        self.tx_buffer = self.tx_buffer[len(chunk) - 1 :]
        tx_credits_spent = True
    else:
        tx_credits_spent = False
```

graynode: Good catch. I just tested it locally and it works. I've updated the PR to reflect these changes.

PR #809, opened 2025-10-31
[Typing] Add controller.py typing

khsiao-google:


bumble/controller.py
```diff
@@ -1881,7 +2030,7 @@ def on_hci_le_setup_iso_data_path_command(
 
     def on_hci_le_remove_iso_data_path_command(
         self, command: hci.HCI_LE_Remove_ISO_Data_Path_Command
-    ) -> bytes:
+    ):
```
zxzxwu: Maybe we should mark all of them `-> Optional[bytes]` using this chance?

khsiao-google: Sure

bumble/controller.py
```diff
@@ -159,13 +159,13 @@ def __init__(
         self.total_num_le_acl_data_packets = 64
         self.iso_data_packet_length = 960
         self.total_num_iso_data_packets = 64
-        self.event_mask = 0
-        self.event_mask_page_2 = 0
+        self.event_mask = b'\x00'
+        self.event_mask_page_2 = b'\x00'
```
zxzxwu: Event masks are bit masks, so it's better to store them as integers (though we don't use them for now). We can convert command parameters as integers to resolve this.

bumble/controller.py
```diff
@@ -193,8 +197,8 @@ def __init__(
         self.local_name = 'Bumble'
 
         self.advertising_interval = 2000  # Fixed for now
-        self.advertising_data = None
-        self.advertising_timer_handle = None
+        self.advertising_data: Optional[bytes] = None
+        self.advertising_timer_handle: Optional[asyncio.Handle] = None
```
zxzxwu: nit: I will prefer to annotate this in class level instead of init

bumble/controller.py
```diff
@@ -481,8 +489,8 @@ def on_link_peripheral_connection_complete(
                 logger.debug(
                     f'New CENTRAL connection handle: 0x{connection_handle:04X}'
                 )
-        else:
-            connection = None
+            else:
+                connection = None
```
zxzxwu: connection should have been assigned in line#476?

khsiao-google: Weird, don't know why it is shifted. Maybe when I use AI to generate the return?

bumble/controller.py
khsiao-google: Is there a reason to this?

bumble/controller.py
khsiao-google: Changing these (event_mask, event_mask_page_2, le_event_mask) from bit to integer will cause the test to fail.

bumble/controller.py
zxzxwu: There isn't a authorative guideline for this like PEP-8 (probably because annotation is introduced "recently"), but I believe the entire community tends to the way I mentioned.

The essense of type annotations are actually build-runtime (despite Python cannot be really built) comments, and it's a special case to infer type inside `__init__` functions. Type analyzers must look deeply into the functions to see what they actually are, while annotations in class bodies are more straight-forward.

Consider a case that in `__init__`, an attribute is not always assigned with the same type of value, but determined by something else like below:

```py
class A:
  def __init__(self, a: bool) -> None:
    if a:
      self.b = 1
    else:
      self.b = '2'
```

The annotation will become


```py
class A:
  def __init__(self, a: bool) -> None:
    self.b: int | str
    if a:
      self.b = 1
    else:
      self.b = '2'
```

Which is less clear for me than

```py
class A:
  b: int | str
  def __init__(self, a: bool) -> None:
    if a:
      self.b = 1
    else:
      self.b = '2'
```

bumble/controller.py
zxzxwu: How does it fail?

khsiao-google: All the test will never be proceed. It will stuck at a2dp_test.py and then timeout.

PR #823, opened 2025-11-08
Refactor classic emulation with LMP protocol

zxzxwu:
This allows us a more consistent semantic of inter-controller operations, removing many functions from link module, and also a better integration some post-complete actions.

### DIFF
--- a/bumble/controller.py
+++ b/bumble/controller.py
@@ -23,9 +23,9 @@ import itertools
 import logging
 import random
 import struct
-from typing import TYPE_CHECKING, Any, Optional, Union
+from typing import TYPE_CHECKING, Any, Optional, Union, cast
 
-from bumble import hci
+from bumble import hci, lmp
 from bumble.colors import color
 from bumble.core import PhysicalTransport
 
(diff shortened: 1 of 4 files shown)


bumble/controller.py
```diff
@@ -303,25 +309,20 @@ async def wait_for_termination(self) -> None:
...
-        for cis_handle in itertools.chain(
-            self.central_cis_links.keys(), self.peripheral_cis_links.keys()
-        ):
-            max_handle = max(max_handle, cis_handle)
-            if cis_handle == handle:
-                # Already used, continue searching after the current max
-                handle = max_handle + 1
-        return handle
+        current_handles = set(
+            cast(Connection | CisLink, link).handle
```
khsiao-google: Can we unite the naming convention in the bumble? Do the connection here stand for classic connection? I think we need a doc to specify the naming convention.

zxzxwu: Usually Connection refers to ACL connection, either LE or BR/EDR, same as the convention used in host and device. The only difference is in controller it also includes SCO. We may saperate SCO from Connection to make a consistency.

PR #824, opened 2025-11-10
Add remote name request

khsiao-google:


bumble/controller.py
```diff
@@ -158,6 +158,7 @@ class Controller:
     le_rpa_timeout: int = 0
     sync_flow_control: bool = False
     local_name: str = 'Bumble'
+    remote_name: dict[hci.Address, bytes]
```
zxzxwu: remote_names

bumble/controller.py
```diff
@@ -895,6 +906,35 @@ def on_classic_sco_connection_complete(
...
+        self.send_lmp_packet(
+            peer_address,
+            lmp.LmpNameRes(
+                name_offset=name_offset,
+                name_length=len(self.local_name.encode('utf-8')),
+                name_fregment=self.local_name.encode('utf-8')[name_offset],
+            ),
+        )
+
+    def on_remote_name_respond(
```
zxzxwu: Handlers should be called `on_[noun]`
```suggestion
    def on_remote_name_response(
```

bumble/controller.py
```diff
@@ -895,6 +906,35 @@ def on_classic_sco_connection_complete(
             )
         )
 
+    def on_remote_name_request(self, peer_address: hci.Address, name_offset: int):
```
zxzxwu: optional - We may unify all handlers here later?
```suggestion
    def on_classic_remote_name_request(self, peer_address: hci.Address, name_offset: int):
```

bumble/controller.py
```diff
@@ -895,6 +906,35 @@ def on_classic_sco_connection_complete(
...
+                name_fregment=self.local_name.encode('utf-8')[name_offset],
+            ),
+        )
+
+    def on_remote_name_respond(
+        self,
+        peer_address: hci.Address,
+        name_offset: int,
+        name_length: int,
+        name_fregment: int,
```
zxzxwu: bytes?

bumble/lmp.py
```diff
@@ -304,3 +304,21 @@ class LmpSwitchReq(Packet):
...
+
+
+@Packet.subclass
+@dataclass
+class LmpNameRes(Packet):
+    opcode = Opcode.LMP_NAME_RES
+
+    name_offset: int = field(metadata=hci.metadata(2))
+    name_length: int = field(metadata=hci.metadata(3))
+    name_fregment: int = field(metadata=hci.metadata('v'))
```
zxzxwu: Though not used, should be `'*'` for plain bytes without header
`'v'`: the first byte is length N, followed by N bytes, parsed object is a bytes of body without leanding length

bumble/controller.py
```diff
@@ -895,6 +906,35 @@ def on_classic_sco_connection_complete(
...
+    ):
+        self.remote_name[peer_address] += name_fregment.to_bytes(length=1)
+        if name_offset + 1 < name_length:
+            self.send_lmp_packet(peer_address, lmp.LmpNameReq(name_offset + 1))
```
zxzxwu: Optional: since we are not actually limited by packet length , we can just send the entire body of name.

tests/device_test.py
```diff
@@ -789,6 +789,17 @@ async def test_accept_classic_connection(roles: tuple[hci.Role, hci.Role]):
...
 
+# -----------------------------------------------------------------------------
+@pytest.mark.asyncio
+async def test_remote_name_request():
+    devices = TwoDevices()
+    devices[0].classic_enabled = True
+    devices[1].classic_enabled = True
+    await devices.setup_connection()
+    remote_name = await devices[0].request_remote_name(devices[1].public_address)
+    assert remote_name == "Bumble"
```
zxzxwu: It's better to set the before requests, or it will rely on the default value

```suggestion
@pytest.mark.asyncio
async def test_remote_name_request():
    devices = TwoDevices()
    devices[0].classic_enabled = True
    devices[1].classic_enabled = True
    expected_name = devices[1].name = "An Awesome Name"
    await devices[0].power_on()
    await devices[1].power_on()
    actual_name = await devices[0].request_remote_name(devices[1].public_address)
    assert actual_name == expected_name
```

zxzxwu (approved):

PR #896, opened 2026-03-03
add basic support for SCO packets over USB

barbibulle:
Still WIP


zxzxwu:
@barbibulle This PR doesn't work with Intel controllers - but I asked Gemini to fix it https://github.com/google/bumble/pull/930, and now I can receive SCO but only with CVSD, but I guess transparent packets may work.

It seems the major difference is that alt setting must be set dynamically? And also input packets must be properly segmented.

barbibulle:
@zxzxwu Would it be possible to rebase #930 on this commit, so that I could easily see what the gemini fix does? (maybe there's a way to do that easily from the GitHub interface, but I don't see it)

zxzxwu:
Sure, I rebased it and pushed to https://github.com/google/bumble/pull/930.

barbibulle:
> Sure, I rebased it and pushed to #930.

Thanks.
I took a quick look at the gemini fix. It's interesting. There's one good element in there, but also some funny mistakes :-) In fact, it works by accident, because of a bug that it created but that ends up calling something with a value that works. Here's what's going on (based on a quick experiment with an Intel controller, I need to do some more experiementation with it). With my Intel controller, Alt Setting 6 does not work. When specifying `+sco=0`, the usb transport will auto-select the setting with the larget packet size, which is this one. The reason it works with the gemini change is that it created a `set_sco_config` to change the alt setting. It created that function with the alternate setting as the second parameter. But then it gets confused, having another function `async def set_sco_config(self, active: bool, air_mode: int)` which uses `air_mode` as its second parameter. In the end, it ends up using neither, because in the body of `set_sco_config`, it will set the alternate setting obtained from earlier from `self.sco_alt_setting =sco_interface.getAlternateSetting()`...
So, skipping all that spaghetti, all you need for the Intel controller to return SCO packets is select an alternate setting that's not 6. I tried it, using `usb:0+sco=1` instead of `usb:0+sco=0`, and I get SCO packets. 
Gemini did the right thing in adding some code to split the buffers into smaller packets, and using multi-packet transactions. That part I will incorporate in my PR.
The audio sending look in the sample app is good too, I had not had a chance to do that yet. So that's cool.

zxzxwu:
Checking on BE200, CVSD is also support. mSBC codec is not supported (despite claimed in supported codec command), but transparent is supported, so we can encode and decode them in host.

bumble/transport/usb.py
```diff
@@ -60,332 +84,609 @@ def load_libusb():
...
+            device.setInterfaceAltSetting(
+                sco_interface.getNumber(), sco_interface.getAlternateSetting()
+            )
+
+        # The source and sink can now start
+        source.start()
+        sink.start()
+
+        # Create a thread to process events
+        self.event_thread = threading.Thread(target=self.run)
```
zxzxwu: Can this be a daemon?

zxzxwu (commented):
@barbibulle Thanks! What's your plan for this PR? We would like to use BE200 for some purposes especially on SCO and ISO.

barbibulle:
I was thinking of just adding the outgoing packet splitting and then try to merge it, then add the multi-packet input parsing as a second PR (this will address an issue reported earlier). For Intel controllers, we can either document that some alternate settings may not work, and that users should select a setting other than 6 manually, or we can make the auto-selection skip config 6, but that may not be ideal, since it might not be broken on all controllers, and since it has a larger packet size, may be more efficient. What do you think?

zxzxwu:
Maybe we can allow passing a spec to select alt settings with a filter? For example:
* `sco=0` : For auto selection
* `sco=N` : For fixed selection
* `sco=N,M` : For auto selection from either N or M
* `sco=N-M` : For auto selection from N to M

barbibulle:
> Maybe we can allow passing a spec to select alt settings with a filter? For example:
> 
> * `sco=0` : For auto selection
> * `sco=N` : For fixed selection
> * `sco=N,M` : For auto selection from either N or M
> * `sco=N-M` : For auto selection from N to M

I suggest we keep is simple for now, with 0 for auto-selection and N for manual selection.
Users are likely going to have to experiment with their controllers to see what works and what doesn't. Using `usb-probe.py --verbose` you see a list of alternate modes and their packets sizes, so if `sco=0` doesn't work, you can try others that are listed.

barbibulle:
With this latest commit, I was able to send and receive audio with a few different controllers. I only tested with CVSD.
I'm still getting audio that's somewhat glitchy (but recognizable), and I have not had time to dig deeper to understand if that's related to the timing of sending audio over USB/HCI, or some other issue.
I suggest we try to merge this PR as it is now, so that others may experiment with it and we can come up with improvements to fix the audio glitches in subsequent PRs. (I'd also like to push a change to support concatenated events in HCI USB transfers, as reported in a recent issue, which I'd like to have done on top of this PR rather than a parallel change).

barbibulle:
NOTE: with my Intel controller, selecting alternate mode 6 doesn't work. Alternate 5 sort-of works, but after a few packets of audio I get a hardware error event... Alternate settings 1, 2, 3 and 4 work better, but I still have some audio glitches.
Laird BT851 (Broadcom controller) works (albeit also with audio glitches), same with some Realtek controllers.

apps/controller_loopback.py
```diff
@@ -36,55 +40,112 @@
...
-            )
 
         self.last_timestamp = now
 
-        if self.expected_cid == self.packet_count:
+        if self.expected_counter == self.packet_count:
             print(color('@@@ Received last packet', 'green'))
             self.done.set()
 
+    def on_sco_packet(self, connection_handle: int, packet) -> None:
```
zxzxwu, line 127: ### Bug in `apps/controller_loopback.py` with SCO connection type

When running `controller_loopback.py` with `--connection-type sco` (or `-t sco`), the test will hang indefinitely after sending all packets.

This is because `self.done.wait()` is called in `run()`, but `self.done.set()` is never called for SCO packets.

`self.done.set()` is only called in `on_l2cap_pdu()`, which is only registered for `'l2cap_pdu'` events (ACL data).
For SCO data, `'sco_packet'` event is registered to `on_sco_packet()`, but `on_sco_packet()` currently only prints the packet and does not perform any accounting or trigger `self.done.set()`.

```python
    def on_sco_packet(self, connection_handle: int, packet) -> None:
        print("---", connection_handle, packet)
```

#### Suggested Fix

You can implement `on_sco_packet` similarly to `on_l2cap_pdu` to support throughput and RTT measurement for SCO loopback:

```python
    def on_sco_packet(self, connection_handle: int, packet) -> None:
        now = time.time()
        # The first 2 bytes of the payload are the counter, similar to ACL loopback
        (counter,) = struct.unpack_from("H", packet.data, 0)
        rtt = now - self.send_timestamps[counter]
        self.rtts.append(rtt)
        print(f'<<< Received SCO packet {counter}: {len(packet.data)} bytes, RTT={rtt:.4f}')
        assert connection_handle == self.connection_handle
        assert counter == self.expected_counter
        self.expected_counter += 1
        if counter == 0:
            self.start_timestamp = now
        else:
            elapsed_since_start = now - self.start_timestamp
            elapsed_since_last = now - self.last_timestamp
            self.bytes_received += len(packet.data)
            instant_rx_speed = len(packet.data) / elapsed_since_last
            average_rx_speed = self.bytes_received / elapsed_since_start
            if self.mode == 'throughput':
                print(
                    color(
                        f'@@@ RX speed: instant={instant_rx_speed:.4f},'
                        f' average={average_rx_speed:.4f},',
                        'cyan',
                    )
                )

        self.last_timestamp = now

        if self.expected_counter == self.packet_count:
            print(color('@@@ Received last SCO packet', 'green'))
            self.done.set()
```

zxzxwu (approved):

apps/controller_loopback.py
barbibulle: fixed

PR #897, opened 2026-03-06
Refine dependencies

zxzxwu:
* Move dev-only deps to `[dependency-groups]`
* Move dependencies for Android transport to `android` group
* Add a `all` option including all other options

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -12,7 +12,7 @@
 
 	// Use 'postCreateCommand' to run commands after the container is created.
     "postCreateCommand": 
-        "python -m pip install '.[build,test,development,documentation]'",
+        "python -m pip install '.[all]' --group dev",
 
     // Configure tool-specific properties.
     "customizations": {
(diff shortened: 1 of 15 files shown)


barbibulle:
I think this PR goes in the right direction. It is unfortunate that there isn't a way to specify a "default" for when no optional dependencies are specified, so this will create a breaking change for users, which we'll have to deal with by documenting the change well, and maybe also trying to print out useful errors when we can. 
I suggest that for optional transports, which users won't get by default anymore, we use a try/except when importing the optional dependencies and print out the proper message (like "please install bumble[xyz]") to be helpful.

barbibulle (approved):

zxzxwu:
For now I will just separate `android` out as it causes the most issues

zyv:
Oh, you folks are so awesome: now I don't have to pull in this heavy grpcio/protobuf stuff for nothing. And thank you so much for Bumble! Owing to it, I was able to implement RFCOMM support in my Python library in no time... and it's much more stable and pleasant to use than some other Bluetooth libraries I tried in the past.

P.S. The only "issue" I encountered so far is that provided virtual HCI doesn't answer inquiries and support pairing/encryption, which is limiting for testing - but this comment doesn't belong to this ticket :) and you probably know about it and have your reasons anyways.

PR #912, opened 2026-04-16
fix: add input validation to prevent remote crash from empty/malforme…

ibondarenko1:
…d PDUs

Add length checks in from_bytes() for ATT and SMP protocol parsers to prevent IndexError crashes from empty PDUs sent by remote Bluetooth devices. Also add buffer size limit and UTF-8 error handling in HFP protocol to prevent memory exhaustion and decode crashes.

- bumble/att.py: validate PDU is non-empty before accessing pdu[0]
- bumble/smp.py: validate PDU is non-empty before accessing pdu[0]
- bumble/hfp.py: limit buffer to 64KB, handle invalid UTF-8 gracefully

These issues can be triggered by a remote Bluetooth device sending malformed packets, causing denial of service on the host.

### DIFF
--- a/bumble/att.py
+++ b/bumble/att.py
@@ -249,6 +249,8 @@ class ATT_PDU:
 
     @classmethod
     def from_bytes(cls, pdu: bytes) -> ATT_PDU:
+        if not pdu:
+            raise ValueError("Empty ATT PDU")
         op_code = pdu[0]
 
         subclass = ATT_PDU.pdu_classes.get(op_code)
--- a/bumble/att.py
+++ b/bumble/att.py
@@ -42,7 +42,7 @@ from typing_extensions import TypeIs
 
 from bumble import hci, l2cap, utils
 from bumble.colors import color
-from bumble.core import UUID, InvalidOperationError, ProtocolError
+from bumble.core import UUID, InvalidOperationError, InvalidPacketError, ProtocolError
 from bumble.hci import HCI_Object
 
 # -----------------------------------------------------------------------------
@@ -250,7 +250,7 @@ class ATT_PDU:
     @classmethod
     def from_bytes(cls, pdu: bytes) -> ATT_PDU:
         if not pdu:
-            raise ValueError("Empty ATT PDU")
+            raise InvalidPacketError("Empty ATT PDU")
         op_code = pdu[0]
 
         subclass = ATT_PDU.pdu_classes.get(op_code)
--- a/bumble/hfp.py
+++ b/bumble/hfp.py
@@ -84,12 +84,15 @@ class HfpProtocol:
     def feed(self, data: bytes | str) -> None:
         # Convert the data to a string if needed
         if isinstance(data, bytes):
-            data = data.decode('utf-8')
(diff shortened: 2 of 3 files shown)


bumble/att.py
```diff
@@ -249,6 +249,8 @@ class ATT_PDU:
 
     @classmethod
     def from_bytes(cls, pdu: bytes) -> ATT_PDU:
+        if not pdu:
+            raise ValueError("Empty ATT PDU")
```
zxzxwu: Please use [core.InvalidPacketError](https://github.com/google/bumble/blob/f2824ee6b86872c31d8f913a27cf095381ad4ab5/bumble/core.py#L153)

bumble/smp.py
```diff
@@ -215,6 +215,8 @@ class SMP_Command:
 
     @classmethod
     def from_bytes(cls, pdu: bytes) -> SMP_Command:
+        if not pdu:
+            raise ValueError("Empty SMP PDU")
```
zxzxwu: ditto

bumble/hfp.py
```diff
@@ -84,12 +84,15 @@ def __init__(self, dlc: rfcomm.DLC) -> None:
...
-            data = data.decode('utf-8')
+            data = data.decode('utf-8', errors='replace')
 
         logger.debug(f'<<< Data received: {data}')
 
         # Add to the buffer and look for lines
         self.buffer += data
+        if len(self.buffer) > 65536:
+            logger.warning("HFP buffer overflow, truncating")
+            self.buffer = self.buffer[-65536:]
```
zxzxwu: Maybe we can just abort new packets after overflow?

```suggestion
MAX_LINE_LENGTH = 65536
...
        if len(self.buffer) > MAX_LINE_LENGTH:
            logger.error("HFP buffer overflow, abort")
            return
```

zxzxwu (commented):
Thanks for your PR, but I am wondering when do we need such handling?

For single packet parsing, we won't crash because there is a try-except umbrella at the packet sink root.

For HFP, despite the issue is possible, "normal" products qualified by SIG and authenticated by users should not send such invalid packets, and users can still define their own handling.

ibondarenko1:
Thanks for the review @zxzxwu — pushed follow-up commit addressing both points:

1. **att.py / smp.py**: switched from `ValueError` to `core.InvalidPacketError` so the exception fits the existing taxonomy and is caught by the packet-sink umbrella try/except as you noted.

2. **hfp.py**: replaced truncation with abort-on-overflow using a `MAX_BUFFER_SIZE` class constant (64KB). When incoming data would overflow, we log a warning and drop only the *new* data via early return, keeping existing partial-packet state intact so a valid in-flight AT command can still parse.

Happy to adjust naming or buffer size if preferred.

ibondarenko1:
@googlebot I signed it!

ibondarenko1:
Gentle ping @zxzxwu — review fixes are in commit 444f43f, CLA is green, and checks pass. Let me know if any further changes are needed before merge. Thanks!

zxzxwu (approved):

PR #915, opened 2026-04-25
improve type hints for notify/indicate subscriber(s) methods

dlech:
Pyright expects generic type parameters to be specified for the Attribute class, otherwise it treats the type as Unknown which can trigger reportUnknownMemberType errors.

This can be solved by using a generic type parameter for these methods which also has the benefit of making sure that the value parameter has the correct type for the attribute.


dlech:
Updated to fix failing CI (passing `invoke project.pre-commit` locally now).

barbibulle (approved):

PR #918, opened 2026-04-27
avdtp: bound message assembler to drop truncated PDUs (DoS prevention)

ibondarenko1:
## Summary

Truncated AVDTP frames from a remote peer crash the L2CAP channel by raising `IndexError` in `MessageAssembler.on_pdu`.

`MessageAssembler.on_pdu` unconditionally indexes `pdu[0]`, `pdu[1]`, and (for START packets) `pdu[2]`. A peer sending a 0-, 1-, or 2-byte AVDTP frame triggers `IndexError`, which propagates up through the L2CAP read loop and tears down the channel.

This is the same DoS class as:
- #912 (empty ATT PDU) — merged
- #914 (unbounded SDP recursion) — merged

## Fix

Validate length before each access. Empty PDUs and packets shorter than the type-specific minimum are logged at warning level and dropped; the assembler stays alive so the L2CAP channel is not torn down.

Both helpers in this fix mirror the conservative warn-and-drop approach already used in `MessageAssembler` for unexpected continuations.

## Test plan

- [x] New parameterized regression `test_message_assembler_truncated_pdu` covering 4 truncated inputs (empty, 1-byte SINGLE, 1-byte START, 2-byte START) — all four raise `IndexError` pre-fix and pass cleanly post-fix.
- [x] Full `tests/avdtp_test.py` suite passes (43 tests).
- [x] `black bumble/ tests/` clean.

### DIFF
--- a/bumble/avdtp.py
+++ b/bumble/avdtp.py
@@ -311,6 +311,13 @@ class MessageAssembler:
     def on_pdu(self, pdu: bytes) -> None:
         self.packet_count += 1
 
+        # Drop empty PDUs sent by remote — accessing pdu[0] below would
+        # raise IndexError, propagating up to the L2CAP read loop and
+        # tearing down the channel. Same class as #912 (ATT empty PDU).
+        if len(pdu) < 1:
+            logger.warning('AVDTP message assembler: empty PDU dropped')
+            return
+
         transaction_label = pdu[0] >> 4
         packet_type = Protocol.PacketType((pdu[0] >> 2) & 3)
         message_type = Message.MessageType(pdu[0] & 3)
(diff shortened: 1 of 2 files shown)


tests/avdtp_test.py
```diff
@@ -120,6 +120,31 @@ def test_messages(message: avdtp.Message):
...
+    log + drop and stay alive so the L2CAP channel survives."""
+    completed = []
+
+    def callback(transaction_label, message):
+        completed.append((transaction_label, message))
+
+    assembler = avdtp.MessageAssembler(callback)
+    # Must not raise; nothing should be delivered to callback either.
+    assembler.on_pdu(pdu)
+    assert completed == []
```
zxzxwu:
```suggestion
    assert not completed
```

Or use mock.

bumble/avdtp.py
```diff
@@ -311,6 +311,13 @@ def reset(self) -> None:
     def on_pdu(self, pdu: bytes) -> None:
         self.packet_count += 1
 
+        # Drop empty PDUs sent by remote — accessing pdu[0] below would
+        # raise IndexError, propagating up to the L2CAP read loop and
+        # tearing down the channel. Same class as #912 (ATT empty PDU).
+        if len(pdu) < 1:
```
zxzxwu:
```suggestion
        if not pdu:
```

zxzxwu (approved):

ibondarenko1:
Both nits addressed in 07b5e33 — `if not pdu:` and `assert not completed`. Thanks for the quick review!

ibondarenko1:
Cross-reference for the Google VRP triage team: this PR is tracked at https://issuetracker.google.com/issues/506803574 (reporter: sactransport2000@gmail.com).

PR #923, opened 2026-05-11
Fix AVDTP endpoint resource leak by clearing the in_use flag on strea…

laozhuxinlu:
Fix an AVDTP endpoint resource leak where the `in_use flag` was not cleared if a stream was closed or aborted before the RTP channel was established. Previously, this caused the endpoint to remain permanently locked, improperly rejecting subsequent SetConfiguration attempts with a SEP_IN_USE error. This patch ensures local_endpoint.in_use is properly reset to 0 when transitioning directly to the IDLE state. Additionally, it wires up the missing local endpoint event trigger for Abort commands to properly notify application listeners.

### DIFF
--- a/bumble/avdtp.py
+++ b/bumble/avdtp.py
@@ -2039,6 +2039,7 @@ class Stream:
 
         if self.rtp_channel is None:
             # No channel to release, we're done
+            self.local_endpoint.in_use = 0
             self.change_state(State.IDLE)
         else:
             # TODO: set a timer as we wait for the RTP channel to be closed
@@ -2050,6 +2051,7 @@ class Stream:
         await self.local_endpoint.on_abort_command()
         if self.rtp_channel is None:
             # No need to wait
+            self.local_endpoint.in_use = 0
             self.change_state(State.IDLE)
         else:
             # Wait for the RTP channel to be closed


bumble/avdtp.py
```diff
@@ -2039,6 +2039,7 @@ async def on_close_command(self) -> Message | None:
 
         if self.rtp_channel is None:
             # No channel to release, we're done
+            self.local_endpoint.in_use = 0
```
zxzxwu, line 2042: It seems in_use is always used to determine whether configure is allowed for the stream endpoint, which is logically equal to `stream.state == State.IDLE`, since in the AVDTP spec, IDLE is the only state which allows setting configuration.

However, another fact is that we release streams very implicitly - only on another configuration procedure, and this requires `in_use == 0`. 

I think we should have a more aligned and clear lifecycle for streams and local endpoints.

Maybe we can declare in_use as a property = `self.stream is not None and self.stream.state == State.IDLE`?

We may also clean up `protocol.streams` table when stream state transitted to IDLE.

zxzxwu: Another solution is just to remove `in_use` attribute and always consider streams in use, so we only need to properly remove them on release.

laozhuxinlu: Hi Josh, your suggestion sounds like a refactoring requirement for the current flawed design, but here it's a bug fix for an abnormal value of the `in_use` flag. We can discuss your thoughts further in the future. How do you think about that?
Thanks

zxzxwu (approved):

PR #928, opened 2026-05-20
AVDTP: Avoid explicit in_use management

zxzxwu:
Delegate in_use to a property checking if stream state is IDLE.

### DIFF
--- a/bumble/avdtp.py
+++ b/bumble/avdtp.py
@@ -2156,8 +2156,15 @@ class DiscoveredStreamEndPoint(StreamEndPoint, StreamEndPointProxy):
         in_use: int,
         capabilities: Iterable[ServiceCapabilities],
     ) -> None:
-        StreamEndPoint.__init__(self, seid, media_type, tsep, in_use, capabilities)
-        StreamEndPointProxy.__init__(self, protocol, seid)
+        StreamEndPoint.__init__(
+            self,
+            seid=seid,
+            media_type=media_type,
+            tsep=tsep,
+            in_use=in_use,
+            capabilities=capabilities,
+        )
+        StreamEndPointProxy.__init__(self, protocol=protocol, seid=seid)
 
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


zxzxwu:
cc @laozhuxinlu

bumble/avdtp.py
```diff
@@ -1947,8 +1947,6 @@ async def close(self) -> None:
             self.rtp_channel = None
 
         # Release the endpoint
```
uier: Should remove comment as well

bumble/avdtp.py
```diff
@@ -2156,14 +2166,27 @@ def __init__(
...
 
 # -----------------------------------------------------------------------------
 class LocalStreamEndPoint(StreamEndPoint, utils.EventEmitter):
     stream: Stream | None
 
+    @property
+    def in_use(self) -> int:
+        if self.stream is None:
+            return 0
+        return 1 if self.stream.state != State.IDLE else 0
```
uier: why not 
```python
if self.stream is None or self.stream.state == State.IDLE:
    return 0
return 1
```

bumble/avdtp.py
```diff
@@ -2186,7 +2209,14 @@ def __init__(
         capabilities: Iterable[ServiceCapabilities],
         configuration: Iterable[ServiceCapabilities] | None = None,
     ):
-        StreamEndPoint.__init__(self, seid, media_type, tsep, 0, capabilities)
+        StreamEndPoint.__init__(
+            self,
+            seid=seid,
+            media_type=media_type,
+            tsep=tsep,
+            in_use=0,
```
uier: It seems `LocalStreamEndPoint` wont use the `in_use` and will compute the `in_use` based on steam state. But it's a required field in `SteamEndPoint`... I think it's okay, just feel a bit redundant here.

bumble/avdtp.py
zxzxwu: Done

bumble/avdtp.py
zxzxwu: Done

```py

    def in_use(self) -> int:
        if self.stream and self.stream.state != State.IDLE:
            return 1
        return 0
```

bumble/avdtp.py
zxzxwu: I turned to make `SteamEndPoint` an ABC without `__init__`, so we don't need to init `_in_use` at all.

cc @barbibulle

uier (approved):

PR #932, opened 2026-05-25
fix(usb): support LE ISO data over Bulk endpoints

zxzxwu:
Some USB Bluetooth controllers (like Intel BE200) transport LE ISO data over Bulk endpoints instead of Isochronous endpoints when Interface 1 alternate settings are not fully configured or supported for LE Audio.

This change:
1. Adds a workaround in the host layer to intercept ACL packets using CIS/BIS handles on Bulk In, reconstructing them into ISO packets.
2. Supports sending HCI ISO Data packets over the default Bulk Out endpoint in the USB transport layer.

This enables a full Bulk-only transport for LE Audio ISO on these controllers without requiring Interface 1 altsetting activation (+sco is not needed).

TAG=agy
CONV=8b9a01f7-32cb-4a83-9300-23c4b688d861

### DIFF
--- a/bumble/host.py
+++ b/bumble/host.py
@@ -247,6 +247,7 @@ class Host(utils.EventEmitter):
     bis_links: dict[int, IsoLink]
     sco_links: dict[int, ScoLink]
     bigs: dict[int, set[int]]
+    link_ts_flags: dict[int, int]
     acl_packet_queue: DataPacketQueue | None = None
     le_acl_packet_queue: DataPacketQueue | None = None
     iso_packet_queue: DataPacketQueue | None = None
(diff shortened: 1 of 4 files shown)


zxzxwu:
ISO can be transmitted now on Intel BE200 and ASUS BT-540, but there are still some problems (melformed?)

Fixed

zxzxwu:
A problem is that vendors have different bulk interfaces definition, though very similar, and a major problem is whether to include timestamp. Here it proposed a self-adaptive solution to determine if timestamp is present.

barbibulle (approved):
Great!

barbibulle (commented):
This PR mentions support for sending ISO packets as well ("2. USB Transport Layer (Bulk Out):
   - Adds support for sending HCI ISO Data packets over the default
     Bulk Out endpoint when Isochronous endpoints are not enabled.") but I don't see the corresponding code here. Are you submitting that in a separate PR?

zxzxwu:
Somehow I removed it probably during rebase...

PR #935, opened 2026-06-02
fix: host race condition on hci acl data packets

klow68:
After some analysis, we found that we had some race condition on acl data packets which arrived before the connection complete

This PR add a fix for the race condition where ACL data can arrive before the corresponding Connection Complete event.
Use a buffer to store packets by handle and replaying them once the connection is established.
And buffered packets are removed after a delay of 100ms

Analysis:
When a BLE central connects to the nRF52840 dongle and immediately sends an SMP Pairing Request, the ACL data packet is delivered to Bumble before the LE Connection Complete event due to USB endpoint ordering, causing Bumble to drop the pairing request for an unknown connection handle and leaving the pairing stuck until the connection times out.

### DIFF
--- a/bumble/host.py
+++ b/bumble/host.py
@@ -17,6 +17,7 @@
 # -----------------------------------------------------------------------------
 from __future__ import annotations
 
+import time
 import asyncio
 import collections
 import dataclasses
@@ -256,6 +257,8 @@ class Host(utils.EventEmitter):
     long_term_key_provider: Callable[[int, bytes, int], Awaitable[bytes | None]] | None
     link_key_provider: Callable[[hci.Address], Awaitable[bytes | None]] | None
 
+    _ACL_STALE_TIMEOUT_S: float = 0.100  # 100 ms
+
     def __init__(
         self,
         controller_source: TransportSource | None = None,
@@ -293,6 +296,9 @@ class Host(utils.EventEmitter):
         self.link_key_provider = None
         self.pairing_io_capability_provider = None  # Classic only
         self.snooper: Snooper | None = None
+        self._pending_acl: dict[int, list[tuple[float, hci.HCI_AclDataPacket]]] = {}
+        self._cleanup_task: asyncio.Task[None] | None = None
+        self.on("le_connection", self._drain_buffered_acl)
 
         # Connect to the source and sink if specified
         if controller_source:
@@ -1033,6 +1039,16 @@ class Host(utils.EventEmitter):
         if connection := self.connections.get(packet.connection_handle):
             connection.on_hci_acl_data_packet(packet)
             return
+        logger.warning(
+            f"ACL data arrived before Connection Complete for handle {packet.connection_handle} — buffering"
+        )
+        self._pending_acl.setdefault(packet.connection_handle, []).append(
+            (time.monotonic(), packet)
+        )
+        if self._cleanup_task is None or self._cleanup_task.done():
+            self._cleanup_task = asyncio.get_event_loop().create_task(
+                self._clean_pending_acl_to_old()
+            )
 
         # WORKAROUND: Some controllers (e.g. Intel BE200) send ISO data wrapped in ACL packets
         # using the CIS handle.
@@ -1119,6 +1135,37 @@ class Host(utils.EventEmitter):
 
     def on_l2cap_pdu(self, connection: Connection, cid: int, pdu: bytes) -> None:
         self.emit('l2cap_pdu', connection.handle, cid, pdu)
+
+    async def _clean_pending_acl_to_old(self) -> None:
+        """Periodically discard buffered ACL packets that have exceeded the timeout."""
+        while self._pending_acl:
+            await asyncio.sleep(self._ACL_STALE_TIMEOUT_S)
+            now = time.monotonic()
+            for handle in list(self._pending_acl):
+                acl_to_keep, acl_timed_out = [], []
+                for ts, packet in self._pending_acl[handle]:
+                    (acl_to_keep if (now - ts) <= self._ACL_STALE_TIMEOUT_S else acl_timed_out).append((ts, packet))
+                for ts, _ in acl_timed_out:
+                    logger.critical(
+                        f"Dropping stale buffered ACL packet for handle {handle}"
+                        f" (age {(now - ts) * self._ACL_STALE_TIMEOUT_S * 10000} ms > "
+                        f"{self._ACL_STALE_TIMEOUT_S * self._ACL_STALE_TIMEOUT_S * 10000} ms timeout)"
+                    )
+                if acl_to_keep:
+                    self._pending_acl[handle] = acl_to_keep
+                else:
+                    del self._pending_acl[handle]
+
+    def _drain_buffered_acl(self, handle: int, *_args: Any, **_kwargs: Any) -> None:
+        """Once le connection happen emit the buffered packets"""
+        queued = self._pending_acl.pop(handle, [])
+        if not queued:
+            return
+        for ts, packet in queued:
+            logger.info(
+                f"Replaying buffered ACL packet for handle handle (age {(time.monotonic() - ts) * 1000} ms)"
+            )
+            self.on_hci_acl_data_packet(packet)
 
     def on_command_processed(
         self, event: hci.HCI_Command_Complete_Event | hci.HCI_Command_Status_Event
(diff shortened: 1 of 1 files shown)


bumble/host.py
```diff
@@ -293,6 +296,9 @@ def __init__(
         self.link_key_provider = None
         self.pairing_io_capability_provider = None  # Classic only
         self.snooper: Snooper | None = None
+        self._pending_acl: dict[int, list[tuple[float, hci.HCI_AclDataPacket]]] = {}
+        self._cleanup_task: asyncio.Task[None] | None = None
```
barbibulle: It's probably not necessary to have an async cleanup task here. It should be sufficient to queue the packets that arrive before the connection event, and remove "stale" packets in the queue at that time. The queue will get emptied when the connection event is handled (after delivering the packets).

bumble/host.py
```diff
@@ -1119,6 +1136,49 @@ def on_hci_iso_data_packet(self, packet: hci.HCI_IsoDataPacket) -> None:
...
+                acl_to_keep: list[tuple[float, hci.HCI_AclDataPacket]] = []
+                acl_timed_out: list[tuple[float, hci.HCI_AclDataPacket]] = []
+                for ts, packet in self._pending_acl[handle]:
+                    (
+                        acl_to_keep
+                        if (now - ts) <= self._ACL_STALE_TIMEOUT_S
+                        else acl_timed_out
+                    ).append((ts, packet))
+                for ts, _ in acl_timed_out:
+                    logger.info(
```
barbibulle: Logging for each stale packet seems overly verbose (definitely too verbose at `info` level)

bumble/host.py
```diff
@@ -1119,6 +1136,49 @@ def on_hci_iso_data_packet(self, packet: hci.HCI_IsoDataPacket) -> None:
...
 
+    async def _clean_pending_acl_to_old(self) -> None:
+        """Periodically discard buffered ACL packets that have exceeded the timeout."""
+        while self._pending_acl:
+            await asyncio.sleep(self._ACL_STALE_TIMEOUT_S)
+            now = time.monotonic()
+            for handle in list(self._pending_acl):
+                acl_to_keep: list[tuple[float, hci.HCI_AclDataPacket]] = []
+                acl_timed_out: list[tuple[float, hci.HCI_AclDataPacket]] = []
+                for ts, packet in self._pending_acl[handle]:
```
barbibulle: Keeping a list of stale packets seems unnecessary. A simple list comprehension with a condition, or a filter, should be enough.

bumble/host.py
```diff
@@ -1119,6 +1136,49 @@ def on_hci_iso_data_packet(self, packet: hci.HCI_IsoDataPacket) -> None:
...
+
+    def _drain_buffered_acl_sync(self, handle: int, *args: Any, **kwargs: Any) -> None:
+        """Create an async task to replay buffered ACL packet"""
+        self._cleanup_task = asyncio.get_event_loop().create_task(
+            self._drain_buffered_acl(handle)
+        )
+
+    async def _drain_buffered_acl(self, handle: int) -> None:
+        """Replay all buffered ACL packet"""
+        await asyncio.sleep(0.1)
```
barbibulle: What is the purpose of this?

bumble/host.py
```diff
@@ -1119,6 +1136,49 @@ def on_hci_iso_data_packet(self, packet: hci.HCI_IsoDataPacket) -> None:
...
+        queued = self._pending_acl.pop(handle, [])
+        if not queued:
+            return
+        for ts, packet in queued:
```
barbibulle, line 1149: It might be a good idea to insert a sleep(0) to allow other tasks to do some work in between two packets (in a normal situation, there should be some task switching between ACL packet arrivals)

bumble/host.py
klow68: I modified the _drain_buffered_acl_sync to do the cleanup at the same time it goes through the queue.
Removed the cleanup_task but found out a mistake that the `_drain_acl_buffer` was not stored.
Should we store it or is it fine since it's a very fast task to not store it at all ?

bumble/host.py
klow68: Going through the all list just for logging seems indeed way to much, I just removed it

bumble/host.py
klow68: acl_timed_out removed, along with the cleanup task

bumble/host.py
klow68: This is some leftovers from some tests, sorry
removed

bumble/host.py
klow68: Nice catch !
Thanks a lot for you're review :)

klow68:
Do I need to do more changes on this PR ? :)

barbibulle:
> Do I need to do more changes on this PR ? :)

I think you'll need to re-base your branch on the latest from main in order for the tests to pass.

klow68:
PR rebased

bumble/host.py
```diff
@@ -1120,6 +1133,25 @@ def on_hci_iso_data_packet(self, packet: hci.HCI_IsoDataPacket) -> None:
     def on_l2cap_pdu(self, connection: Connection, cid: int, pdu: bytes) -> None:
         self.emit('l2cap_pdu', connection.handle, cid, pdu)
 
+    def _drain_buffered_acl_sync(self, handle: int, *args: Any, **kwargs: Any) -> None:
```
barbibulle, line 1136: This function doesn't seem to be used anymore

bumble/host.py
```diff
@@ -1120,6 +1133,25 @@ def on_hci_iso_data_packet(self, packet: hci.HCI_IsoDataPacket) -> None:
...
+    async def _drain_buffered_acl(self, handle: int) -> None:
+        """Replay all buffered ACL packet"""
+        queued = self._pending_acl.pop(handle, [])
+        if not queued:
+            return
+        now = time.monotonic()
+        logger.info(f"Replaying buffered ACL packets for handle: {handle}")
+        for ts, packet in queued:
+            # Sleep to allow other tasks to run
+            await asyncio.sleep(0)
```
barbibulle, line 1151: I think this may potentially lead to out-of-order packets being delivered. Upon connection, the packets would be re-queued, but there could be newer/fresher packets that are delivered between iterations of this loop.
Maybe this should just requeue everything without sleeping?

bumble/host.py
```diff
@@ -293,6 +296,9 @@ def __init__(
         self.link_key_provider = None
         self.pairing_io_capability_provider = None  # Classic only
         self.snooper: Snooper | None = None
+        self._pending_acl: dict[int, list[tuple[float, hci.HCI_AclDataPacket]]] = {}
+        self._drain_acl_buffer_task: asyncio.Task[None] | None = None
+        self.on("le_connection", self._drain_buffered_acl)
```
barbibulle, line 301: This isn't going to work: the "le_connection" event has parameters, but `_drain_buffered_acl` doesn't take any...

klow68: I fixed this, thx for spotting that :)

bumble/host.py
klow68: I removed it

bumble/host.py
klow68: I removed it in the end, not sure what's the best option.
You where the one asking for a sleep(0) https://github.com/google/bumble/pull/935/changes#r3374682046
But should be very fast to just queue everything without "stopping" the task between each packet

bumble/host.py
```diff
@@ -1120,6 +1133,19 @@ def on_hci_iso_data_packet(self, packet: hci.HCI_IsoDataPacket) -> None:
...
 
+    async def _drain_buffered_acl(
+        self, handle: int, *_connection_parameters: Any
+    ) -> None:
+        """Replay all buffered ACL packet"""
+        queued = self._pending_acl.pop(handle, [])
+        if not queued:
+            return
+        now = time.monotonic()
+        logger.info(f"Replaying buffered ACL packets for handle: {handle}")
```
barbibulle, line 1144: Sorry, one minor change I should have pointed out in my earlier review: this should be `logger.debug`, as `info` is too verbose for this type of detail.

bumble/host.py
```diff
@@ -293,6 +296,9 @@ def __init__(
         self.link_key_provider = None
         self.pairing_io_capability_provider = None  # Classic only
         self.snooper: Snooper | None = None
+        self._pending_acl: dict[int, list[tuple[float, hci.HCI_AclDataPacket]]] = {}
+        self._drain_acl_buffer_task: asyncio.Task[None] | None = None
```
barbibulle, line 300: This is unused, should be removed.

PR #925, opened 2026-05-18
fix: split concatenated HCI events instead of dropping them

YP6729:
# Fix: USB transport silently drops HCI events packed in the same USB transfer

## Summary

When a Bluetooth USB dongle delivers two HCI events in a single USB transfer, Bumble's `UsbTransport.dequeue()` calls `sink.on_packet()` once with the full buffer.  `HCI_Event.from_bytes` then parses only the **first** event, logs a warning about surplus bytes, and **silently discards the rest**.  If the discarded bytes happen to be an `HCI_COMMAND_COMPLETE_EVENT`, the corresponding `send_sync_command` future never resolves, its 10-second timeout fires, and the caller raises `TimeoutError`.

## Observed failure

Some tests fails sporadically (~1 out of 200 test) with a `TimeoutError` during the BLE connection setup.

### Pytest log except

```
2026-05-18 11:59:32.221 [   DEBUG][bumble.host:host.py (send_hci_packet):660] ### HOST -> CONTROLLER: HCI_LE_SET_SCAN_ENABLE_COMMAND:
  le_scan_enable:    1
  filter_duplicates: 0
2026-05-18 11:59:32.225 [   DEBUG][bumble.host:host.py (on_hci_packet):997] ### CONTROLLER -> HOST: HCI_COMMAND_COMPLETE_EVENT:
  num_hci_command_packets: 1
  command_opcode:          HCI_LE_SET_SCAN_ENABLE_COMMAND
  return_parameters:
    status: SUCCESS
2026-05-18 11:59:32.232 [   DEBUG][bumble.host:host.py (on_hci_packet):997] ### CONTROLLER -> HOST: HCI_LE_ADVERTISING_REPORT_EVENT:
  reports[0]:
    event_type:   ADV_IND
    address_type: PUBLIC_DEVICE
    address:      00:04:63:92:0B:2B/P
    data:         020105030202fe170964323163376537652d353262302d313166312d396262
    rssi:         -39
2026-05-18 11:59:32.234 [   DEBUG][bumble.host:host.py (on_hci_packet):997] ### CONTROLLER -> HOST: HCI_LE_ADVERTISING_REPORT_EVENT:
  reports[0]:
    event_type:   SCAN_RSP
    address_type: PUBLIC_DEVICE
    address:      00:04:63:92:0B:2B/P
    data:         09ffa60210eb01000001
    rssi:         -39
[...]

PR #936, opened 2026-06-03
usb transport packet splitter

barbibulle:
This PR adds packet splitters for all incoming HCI data streams. Even though in most cases we observe that each incoming transfer only contains a single packet, in some cases (as reported in #925) there may be more than one packet (or even partial packets) in transfer buffers. The splitters accumulate data until a complete packet has been receive, then emit complete packets only.
Also included in this PR is the use of multiple transfers for isochronous endpoints (currently set at 2, which should be enough for a simple double-buffering mechanism), which helps avoid data drops for SCO packets (I was able to get HFP audio streams without audio glitches now, which were caused by data being dropped when data arrived before a new transfer was submitted, at the end of a completed transfer).

### DIFF
--- a/bumble/transport/usb.py
+++ b/bumble/transport/usb.py
@@ -59,6 +59,8 @@ USB_BT_HCI_CLASS_TUPLE = (
 )
 
 MAX_SCO_PACKET_SIZE = 1024
+MAX_SCO_IN_PACKETS = 128
+NUMBER_OF_SCO_IN_TRANSFERS = 2
 
 
 # -----------------------------------------------------------------------------
@@ -388,20 +390,35 @@ class UsbPacketSink:
 READ_SIZE = 4096
 
 
-class ScoAccumulator:
-    def __init__(self, emit: Callable[[bytes], Any]) -> None:
+class PacketSplitter:
+    """Splitter than can parse a byte stream and extract packets that consist of a
+    header and a body, where the header includes an n-byte 'length' field at a
+    certain offset.
+    Extracted packets are emitted by calling a function passed to the constructor,
+    with the full packet (header + body) as argument.
+    """
+
+    def __init__(
+        self, length_offset: int, length_size: int, emit: Callable[[bytes], Any]
+    ) -> None:
         self.emit = emit
         self.packet = b''
+        self.length_offset = length_offset
+        self.length_size = length_size
+        self.header_size = length_offset + length_size
 
     def feed(self, data: bytes) -> None:
         while data:
-            # Accumulate until we have a complete 3-byte header
-            if (bytes_needed := 3 - len(self.packet)) > 0:
+            # Accumulate until we have a complete header
+            if (bytes_needed := self.header_size - len(self.packet)) > 0:
                 self.packet += data[:bytes_needed]
                 data = data[bytes_needed:]
                 continue
 
-            packet_length = 3 + self.packet[2]
+            packet_length = self.header_size + int.from_bytes(
+                self.packet[self.length_offset : self.length_offset + self.length_size],
+                'little',
+            )
             bytes_needed = packet_length - len(self.packet)
             self.packet += data[:bytes_needed]
             data = data[bytes_needed:]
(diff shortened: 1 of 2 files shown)


zxzxwu (changes requested):
Thanks for the PR! I did a review and found a potential bug in `PacketSplitter.feed`.

### Bug in `PacketSplitter.feed`

If the incoming data chunk ends exactly at the header boundary, and the packet has a 0-length payload, the packet will not be emitted immediately. It will be stuck until the next `feed` call.

#### Scenario:
1. `feed` is called with data that exactly matches the header (e.g. 4 bytes for ACL header).
2. `self.packet` accumulates the header. `data` becomes empty.
3. The code hits `continue` at the end of the header accumulation block:
   ```python
   if (bytes_needed := self.header_size - len(self.packet)) > 0:
       self.packet += data[:bytes_needed]
       data = data[bytes_needed:]
       continue
   ```
4. The loop terminates because `data` is empty.
5. The code below the `if` block (which calculates `packet_length` and emits the packet if complete) is never reached.

If the payload length is 0, the packet is already complete but not emitted.

#### Suggested Fix:
Only `continue` if we don't have a complete header yet:

```diff
@@ -413,7 +413,8 @@
             if (bytes_needed := self.header_size - len(self.packet)) > 0:
                 self.packet += data[:bytes_needed]
                 data = data[bytes_needed:]
-                continue
+                if len(self.packet) < self.header_size:
+                    continue
```

### Unit Tests
I recommend adding unit tests for `PacketSplitter` to verify this. Here are some tests I wrote that catch this issue:

```python
import pytest
from bumble import hci
from bumble.transport import usb

def test_packet_splitter_complete():
    emitted = []
    splitter = usb.AclPacketSplitter(emitted.append)
    packet = bytes([0x01, 0x00, 0x04, 0x00, 0x11, 0x22, 0x33, 0x44])
    splitter.feed(packet)
    assert emitted == [packet]

def test_packet_splitter_chunks():
    emitted = []
    splitter = usb.AclPacketSplitter(emitted.append)
    packet = bytes([0x01, 0x00, 0x04, 0x00, 0x11, 0x22, 0x33, 0x44])
    splitter.feed(packet[:4])
    assert emitted == []
    splitter.feed(packet[4:])
    assert emitted == [packet]

def test_packet_splitter_multiple():
    emitted = []
    splitter = usb.AclPacketSplitter(emitted.append)
    packet1 = bytes([0x01, 0x00, 0x04, 0x00, 0x11, 0x22, 0x33, 0x44])
    packet2 = bytes([0x02, 0x00, 0x02, 0x00, 0x55, 0x66])
    splitter.feed(packet1 + packet2)
    assert emitted == [packet1, packet2]

def test_packet_splitter_partial():
    emitted = []
    splitter = usb.AclPacketSplitter(emitted.append)
    packet1 = bytes([0x01, 0x00, 0x04, 0x00, 0x11, 0x22, 0x33, 0x44])
    packet2 = bytes([0x02, 0x00, 0x02, 0x00, 0x55, 0x66])
    splitter.feed(packet1 + packet2[:4])
    assert emitted == [packet1]
    splitter.feed(packet2[4:])
    assert emitted == [packet1, packet2]

def test_packet_splitter_empty_payload():
    emitted = []
    splitter = usb.AclPacketSplitter(emitted.append)
    packet = bytes([0x01, 0x00, 0x00, 0x00])
    splitter.feed(packet)
    assert emitted == [packet]

def test_sco_packet_splitter():
    emitted = []
    splitter = usb.ScoPacketSplitter(emitted.append)
    packet = bytes([0x01, 0x00, 0x03, 0x11, 0x22, 0x33])
    splitter.feed(packet)
    assert emitted == [packet]

def test_event_packet_splitter():
    emitted = []
    splitter = usb.EventPacketSplitter(emitted.append)
    packet = bytes([0x04, 0x02, 0x11, 0x22])
    splitter.feed(packet)
    assert emitted == [packet]
```

zxzxwu (approved):

PR #943, opened 2026-06-24
CIG: Fix CIG parameters for unidirectional CIS

uier:
When setting up a CIG for a unidirectional CIS (e.g., Central-to-Peripheral only), the unused direction's max SDU size is set to 0. However, the other parameters (retransmission count and PHY) defaulted to non-zero values. This caused controllers to reject the configuration during CIG setup or CIS establishment with error 0x30 (Parameter Out Of Mandatory Range).

This change resets the retransmission count and PHY to 0 for any direction where max SDU is 0. A test_cis_parameters_unidirectional test is included with this change.

This commit also fixes parameterize decorator syntax in several test files to resolve test collection failures in Python 3.13 / pytest 8.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -1576,6 +1576,10 @@ class CigParameters:
         rtn_p_to_c: int = DEVICE_DEFAULT_ISO_CIS_RTN  # Number of P->C retransmissions
 
         def __post_init__(self) -> None:
+            # For unidirectional CIS (e.g., Central-to-Peripheral only), the unused direction's
+            # SDU size is 0. If SDU size is 0, the corresponding retransmission count and PHY
+            # must also be set to 0. Otherwise, some controllers will reject the parameters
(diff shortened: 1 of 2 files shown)


bumble/device.py
```diff
@@ -1575,6 +1575,14 @@ class CisParameters:
         rtn_c_to_p: int = DEVICE_DEFAULT_ISO_CIS_RTN  # Number of C->P retransmissions
         rtn_p_to_c: int = DEVICE_DEFAULT_ISO_CIS_RTN  # Number of P->C retransmissions
 
+        def __post_init__(self) -> None:
+            if self.max_sdu_c_to_p == 0:
```
zxzxwu, line 1583: Need a comment here explaining the reason

uier: Done

zxzxwu (approved):

PR #948, opened 2026-06-29
Fix risk of accepting packets before firmware is loaded

typenoob:
This PR fixes an error that occurs when a remote device repeatedly sends HCI_CONNECTION_REQUEST_COMMAND early while the local device still needs to load its firmware driver:

```
[2026-06-29 09:01:39.812] WARNING in host: HCI_ACCEPT_CONNECTION_REQUEST_COMMAND failed (UNKNOWN_CONNECTION_IDENTIFIER_ERROR)
Task exception was never retrieved
future: <Task finished name='Task-5' coro=<Host.send_async_command() done, defined at C:\Users\user\nxbt\.venv\Lib\site-packages\bumble\host.py:805> exception=HCI_Error()>
  File "C:\Users\user\nxbt\.venv\Lib\site-packages\bumble\host.py", line 829, in send_async_command
    raise hci.HCI_Error(status)
bumble.hci.HCI_Error: HCI_Error(hci/UNKNOWN_CONNECTION_IDENTIFIER_ERROR [0x2])
```

The root cause is that `Host.ready` was incorrectly set to True before the firmware driver had finished loading successfully.

In addition, I believe letting Drivers control `Host.ready` is risky, so this behavior has been consolidated and moved into `Host.reset` instead.

### DIFF
--- a/bumble/drivers/intel.py
+++ b/bumble/drivers/intel.py
@@ -445,7 +445,6 @@ class Driver(common.Driver):
             )
 
     async def load_firmware(self) -> None:
-        self.host.ready = True
         device_info = await self.read_device_info()
         logger.debug(
             "device info: \n%s",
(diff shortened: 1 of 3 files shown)


typenoob:
I believe you can reproduce this issue on Windows by first putting the remote device into pairing mode, and then calling the Intel Bluetooth adapter's `Device.power_on` function.

typenoob:
Testing, linting, and formatting were not performed yet.

PR #950, opened 2026-07-02
fix: exclude channel 76 from default CS channel_map (Core Spec 6.0 compliance)

### LINKED ISSUE #945: Default channel_map in Device.create_cs_config enables channel 76, which the CS spec requires to be excluded
## Summary

`Device.create_cs_config(...)` ([`bumble/device.py:5412`](https://github.com/google/bumble/blob/main/bumble/device.py#L5412)) defaults its `channel_map` parameter to:

```python
channel_map: bytes = b'\x54\x55\x55\x54\x55\x55\x55\x55\x55\x15',
```

The last byte (`0x15` = `0b0001_0101`) has bit 4 set, which corresponds to **LE channel 76**. Per Core Spec 6.0 Vol 4 Part E §7.8.117 (LE CS Create Config Command), channels **0, 1, 23, 24, 25, 76, 77, 78** must be excluded from the CS Channel_Map (they're the LE advertising-channel frequencies) and channel 79 is reserved.

Permissive controllers tolerate the spec-violation and let `create_cs_config` succeed anyway. Strict controllers (verified on **Nordic nRF54L15 + NCS Bluetooth controller firmware**) correctly reject the command with `HCI_COMMAND_STATUS_EVENT { status: UNSUPPORTED_FEATURE_OR_PARAMETER_VALUE_ERROR (0x11) }`.

## Steps to reproduce

```python
from bumble.device import Device
# … connect, pair, encrypt, etc. …
await device.set_default_cs_settings(connection, role_enable=int(hci.CsRoleMask.INITIATOR))
config = await device.create_cs_config(connection)  # uses the buggy default channel_map
```

Against an nRF54L15 controller (or any controller that validates the channel map against spec), the call raises `HCI_Error` with status `0x11` (UNSUPPORTED_FEATURE_OR_PARAMETER_VALUE_ERROR).

## Channel-map byte-by-byte analysis

Bit *k* of byte *⌊k/8⌋* in `channel_map` represents LE channel *k*. The default `b'\x54\x55\x55\x54\x55\x55\x55\x55\x55\x15'` decodes as:

| Byte | Hex | Bits set | Channels enabled | Verdict |
|----|----|----|----|----|
| 0 | `0x54` | 2, 4, 6 | 2, 4, 6 | OK (channels 0, 1 excluded) |
| 1 | `0x55` | 0, 2, 4, 6 | 8, 10, 12, 14 | OK |
| 2 | `0x55` | 0, 2, 4, 6 | 16, 18, 20, 22 | OK (channel 23 excluded) |
| 3 | `0x54` | 2, 4, 6 | 26, 28, 30 | OK (channels 24, 25 excluded) |
| 4 | `0x55` | 0, 2, 4, 6 | 32, 34, 36, 38 | OK |
| 5 | `0x55` | 0, 2, 4, 6 | 40, 42, 44, 46 | OK |
| 6 | `0x55` | 0, 2, 4, 6 | 48, 50, 52, 54 | OK |
| 7 | `0x55` | 0, 2, 4, 6 | 56, 58, 60, 62 | OK |
| 8 | `0x55` | 0, 2, 4, 6 | 64, 66, 68, 70 | OK |
| **9** | **`0x15`** | 0, 2, **4** | 72, 74, **76** | **WRONG — channel 76 must be excluded** |

The intent of the existing default appears to be "even-numbered LE channels in the legal CS range". Byte 9 should be **`0x05`** (bits 0 and 2 only — channels 72 and 74 enabled, 76/77/78/79 cleared).

## Proposed fix

One-byte change at `bumble/device.py:5412`:

```diff
-    channel_map: bytes = b'\x54\x55\x55\x54\x55\x55\x55\x55\x55\x15',
+    channel_map: bytes = b'\x54\x55\x55\x54\x55\x55\x55\x55\x55\x05',
```

Same default channel pattern (even-numbered LE channels in the legal range), now spec-compliant.

A more general fix is to derive the default from a spec-aware helper (akin to Zephyr's `bt_le_cs_set_valid_chmap_bits()`) that takes an "intended pattern" and applies the spec exclusion mask. Reasonable as a follow-up, but the one-byte change above unblocks strict controllers without any API churn.

## Workaround for downstream users

Pass `channel_map` explicitly to `create_cs_config` until the default is fixed:

```python
config = await device.create_cs_config(
    connection,
    channel_map=b'\x54\x55\x55\x54\x55\x55\x55\x55\x55\x05',
)
```

## Notes

- Confirmed against an nRF54L15 dev kit running the NCS Bluetooth controller (HCI uart sample) — the controller rejects only when channel 76 is enabled; with byte 9 corrected to `0x05` the `LE_CS_Create_Config` is accepted.
- Bug surfaces only when the controller validates the channel map strictly. Many controllers don't.

dcaayushd:
Fixes #945

the default channel_map in Device.create_cs_config() had byte 9 set 
to 0x15 which enables channel 76. Core Spec 6.0 Vol 4 Part E §7.8.117 
requires channels 0,1,23,24,25,76,77,78 to be excluded from the CS 
Channel_Map. strict controllers (e.g. Nordic nRF54L15) correctly reject 
the command with UNSUPPORTED_FEATURE_OR_PARAMETER_VALUE_ERROR (0x11).

fix: change byte 9 from 0x15 to 0x05, keeping only channels 72 and 74 
enabled in that byte. same even-numbered channel pattern, now spec-compliant.

added a test verifying the default map has no overlap with the forbidden 
channel set.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -5442,7 +5442,7 @@ class Device(utils.CompositeEventEmitter):
         role: int = hci.CsRole.INITIATOR,
         rtt_type: int = hci.RttType.AA_ONLY,
         cs_sync_phy: int = hci.CsSyncPhy.LE_1M,
-        channel_map: bytes = b'\x54\x55\x55\x54\x55\x55\x55\x55\x55\x15',
+        channel_map: bytes = b'\x54\x55\x55\x54\x55\x55\x55\x55\x55\x05',
         channel_map_repetition: int = 0x01,
         channel_selection_type: int = hci.HCI_LE_CS_Create_Config_Command.ChannelSelectionType.ALGO_3B,
         ch3c_shape: int = hci.HCI_LE_CS_Create_Config_Command.Ch3cShape.HAT,
--- a/tests/device_test.py
+++ b/tests/device_test.py
@@ -17,6 +17,7 @@
 # -----------------------------------------------------------------------------
 import asyncio
 import functools
+import inspect
 import logging
 import os
 from unittest import mock
@@ -691,6 +692,25 @@ async def test_power_on_default_static_address_should_not_be_any():
     await devices[0].power_on()
 
     assert devices[0].static_address != Address.ANY_RANDOM
+
+
+# -----------------------------------------------------------------------------
+def test_cs_channel_map_excludes_forbidden_channels():
+    forbidden = {0, 1, 23, 24, 25, 76, 77, 78, 79}
+    default_map = (
+        inspect.signature(Device.create_cs_config).parameters['channel_map'].default
+    )
+
+    enabled = {
+        byte_idx * 8 + bit
+        for byte_idx, byte in enumerate(default_map)
+        for bit in range(8)
+        if byte & (1 << bit)
+    }
+
+    assert enabled.isdisjoint(
+        forbidden
+    ), f"Default channel_map enables forbidden CS channels: {enabled & forbidden}"
 
 
 # -----------------------------------------------------------------------------


apps/console.py
```diff
@@ -579,7 +579,7 @@ async def rssi_monitor_loop(self):
 
     async def command(self, command):
         try:
-            (keyword, *params) = command.strip().split(' ')
+            keyword, *params = command.strip().split(' ')
```
zxzxwu: Please do not format other files.

dcaayushd: Done

zxzxwu (approved):

PR #953, opened 2026-07-09
Add Bluetooth 6.3 CS Enhancements (Inline PCT) support

JoeBakalor:
## Summary

Adds Bluetooth 6.3 Channel Sounding Enhancements support to Bumble — specifically the **Inline Phase Correction Term transfer (IPT)** feature.

- **HCI plumbing** — opcodes, event codes, dataclasses, `IntFlag` enums, supports-commands mask, host dispatch handler. One breaking rename inside `HCI_LE_CS_Create_Config_Command`: the trailing `reserved` byte becomes `cs_enhancements` (see notes).
- **Device wiring** — `Device.power_on()` prefers the [v2] capabilities read when the controller advertises it and falls back to the 6.0 read otherwise; `create_cs_config()` gains a `cs_enhancements` kwarg, default 0.

Net change: **+225 / −50 across three files** (`hci.py`, `device.py`, `host.py`).

## What's implemented

| Symbol | Value / meaning |
|---|---|
| `HCI_LE_CS_READ_LOCAL_SUPPORTED_CAPABILITIES_V2_COMMAND` | `0x20A5` — new opcode; return parameters append the [v2] tail (6 octets, below). |
| `HCI_LE_CS_WRITE_CACHED_REMOTE_SUPPORTED_CAPABILITIES_V2_COMMAND` | `0x20A6` — opcode constant only (no in-tree caller yet). |
| `HCI_LE_CS_READ_REMOTE_SUPPORTED_CAPABILITIES_COMPLETE_V2_EVENT` | Subevent `0x38` — 6.3 variant of `0x2C` with the same [v2] tail. |
| `HCI_LE_CS_Read_Local_Supported_Capabilities_V2_Command` / `_V2_ReturnParameters` | New dataclasses. |
| `HCI_LE_CS_Read_Remote_Supported_Capabilities_Complete_V2_Event` | New event class. |
| `CsSubfeature.CS_IPT_REFLECTOR` | `1 << 4` — bit in `subfeatures_supported` advertising IPT reflector capability. |
| `CsEnhancements.INLINE_PCT` | `0x01` — bit 0 of the `CS_Enhancements` byte of `HCI_LE_CS_Create_Config`, enabling IPT in the CS reflector. |
| `HCI_LE_CS_Create_Config_Command.reserved` → `cs_enhancements` | Rename of the trailing byte (see below). |
| `Device.create_cs_config(..., cs_enhancements=0)` | New kwarg; default preserves 6.0 behavior. |
| `Device.power_on()` CS bring-up | Issues the [v2] caps read when `supports_command()` reports it, else the 6.0 read. |
| `ChannelSoundingCapabilities` | Gains `t_ip2_ipt_times_supported`, `t_sw_ipt_times_supported`, `rtt_2m_aa_only_n`, `rtt_2m_sounding_n`, `rtt_2m_random_sequence_n`, `cs_ipt_reflector_supported`. |

### The [v2] tail

Per §7.8.130 and §7.7.65.39, the [v2] payloads append 6 octets after `TX_SNR_Capability`, taking the return parameters from 29 to 35 octets including `Status`:

| Parameter | Size |
|---|---|
| `T_IP2_IPT_Times_Supported` | 2 octets |
| `T_SW_IPT_Times_Supported` | 1 octet |
| `RTT_2M_AA_Only_N` | 1 octet |
| `RTT_2M_Sounding_N` | 1 octet |
| `RTT_2M_Random_Sequence_N` | 1 octet |

`RTT_Random_Sequence_N` keeps its 6.0 name in the [v2] structures, so all four payload shapes (local [v1]/[v2] return parameters, remote [v1]/[v2] events) share the 6.0 field names and differ only by this tail.

## The `reserved` → `cs_enhancements` rename

The trailing byte of `HCI_LE_CS_Create_Config` was reserved in 6.0. Bluetooth 6.3 §7.8.137 names it `CS_Enhancements` and assigns bit 0 as **IPT enabled in the CS reflector**.

Impact on downstream callers: the field is now named `cs_enhancements` rather than `reserved`. Callers that were passing `reserved=0x00` explicitly must update to `cs_enhancements=0x00` (or drop the kwarg entirely — the default is `0`). No known in-tree caller does this — `Device.create_cs_config` is updated here.

## Design notes

- **`supports_command()` gate for the [v2] read** — `Device.power_on()` selects between the 6.0 and [v2] capability reads on octet 49 bit 2 of the LE Supported Commands reply, matching the `supports_command()` check used for `HCI_LE_SET_HOST_FEATURE_COMMAND` a few lines above.
- **One event name for both variants** — the [v2] payload is a superset of the 6.0 one, so `Host` emits `cs_remote_supported_capabilities` for both subevents and `Device` keeps a single handler annotated with the union of the two event types. Existing listeners are unaffected.
- **`ChannelSoundingCapabilities.from_hci()`** — one classmethod builds the record from any of the four HCI payload shapes, so the 20-plus-argument constructor isn't repeated per call site. The [v2] tail fields default to `0` / `False`, keeping existing positional constructor calls working.

## Testing

Verified end-to-end against Nordic hardware:

- **Initiator**: nRF54L15 running `hci_uart` from nRF Connect SDK v3.4.0 (with `CONFIG_BT_CTLR_EXTENDED_FEAT_SET=y`), attached as Bumble's HCI transport over USB serial.
- **Reflector**: nRF54L15 running Nordic's `channel_sounding/ipt_reflector` sample from nRF Connect SDK v3.4.0.

Observed:

- [v2] caps read returned `t_ip2_ipt_times_supported = 0x7E` (bits 1–6 set → 10/20/30/40/50/60 μs), `t_sw_ipt_times_supported = 0x0A` (10 μs), `subfeatures_supported = 0x12` (bit 4 → `CS_IPT_REFLECTOR` set).
- `HCI_LE_CS_Create_Config` with the CS Enhancements byte set to `0x01` completed with `status: SUCCESS`.
- `HCI_LE_CS_Config_Complete_Event` echoed the byte back as `1`.
- `HCI_LE_CS_Procedure_Enable_Complete_Event` returned `state: ENABLED`.
- Multiple `HCI_LE_CS_Subevent_Result_Event`s streamed with `procedure_counter` incrementing normally.

### Raw reply lengths, measured

Re-measured on the same nRF54L15 after the field list was corrected, capturing the raw command-complete return parameters for both opcodes:

```
opcode 0x2089 ([v1]): 29 octets  0001000001010101001e001e000000000212007c007e00e00103000000
opcode 0x20A5 ([v2]): 35 octets  0001000001010101001e001e000000000212007c007e00e001030000007e000a1e001e
                                                                                       └── tail: 7e00 0a 1e 00 1e
```

The [v2] reply is byte-identical to the [v1] reply for its first 29 octets and then carries the full 6-octet tail, decoding as `t_ip2_ipt_times_supported = 0x7E`, `t_sw_ipt_times_supported = 0x0A`, `rtt_2m_aa_only_n = 30`, `rtt_2m_sounding_n = 0`, `rtt_2m_random_sequence_n = 30`. The first two match the values this PR originally recorded, so the controller had been sending the full tail all along — the previous field list was discarding the last 6 octets, which Bumble ignores silently as trailing data.

Parsing is additionally covered by decoding a synthesized subevent `0x38` and checking it re-serializes byte for byte, and by confirming the 6.0 payloads still decode with the [v2] tail left at zero.

## Note on Zephyr's struct definitions

Worth flagging for anyone reading these structures against Zephyr rather than the spec: Zephyr's CS capability structs stop after `t_sw_ipt_time_supported` and omit the three `RTT_2M_*` parameters, describing a 32-octet reply.

- [`bt_hci_rp_le_read_local_supported_capabilities_v2`](https://github.com/zephyrproject-rtos/zephyr/blob/62acbd571c7294d899861a507e34533e16da7758/include/zephyr/bluetooth/hci_types.h#L2730-L2777)
- [`bt_hci_evt_le_cs_read_remote_supported_capabilities_complete_v2`](https://github.com/zephyrproject-rtos/zephyr/blob/62acbd571c7294d899861a507e34533e16da7758/include/zephyr/bluetooth/hci_types.h#L4155-L4204)
- `bt_hci_cp_le_write_cached_remote_supported_capabilities_v2` in the same header is short the same three fields.

That is a host-side header, and the measurement above shows the shipping controller replies with the spec's 35 octets, so it doesn't affect this PR. It is only worth knowing that a short reply from some other controller would raise during parsing rather than being rejected cleanly. Happy to add a tolerant tail parse or a fallback to the 6.0 read if you'd rather Bumble degrade there, but I've no controller that needs it.

## Spec references

- **Bluetooth Core Spec 6.3, Vol 4 Part E §7.8.137** — LE CS Create Config: `CS_Enhancements`, bit 0 = IPT enabled in the CS reflector.
- **§7.8.130** — LE CS Read Local Supported Capabilities command [v1] and [v2]: the [v2] opcode and the full return-parameter list, including the IPT timings and the 2M RTT sequence lengths.
- **§7.7.65.39** — LE CS Read Remote Supported Capabilities Complete event [v1] and [v2]: the [v2] subevent code and parameter list.

## Follow-ups (not in this PR)

- Adding a corresponding `HCI_LE_CS_Write_Cached_Remote_Supported_Capabilities_V2_Command` dataclass (opcode constant is defined; body not wired). §7.8.132 gives it the same [v2] tail. Uncommon path in practice.
- Any further CS Enhancements defined past Inline PCT (bits 1..7 of `CS_Enhancements` remain reserved).


bumble/device.py
```diff
@@ -2958,9 +2970,37 @@ async def power_on(self) -> None:
...
+                # IPT timings whenever the controller actually implements V2.
+                # `result` is one of two return-parameter shapes; annotate the
+                # union so the V2 read and the 6.0 fallback both type-check, and
+                # read the renamed rtt_random field in-branch where the concrete
+                # type is known.
+                result: (
+                    hci.HCI_LE_CS_Read_Local_Supported_Capabilities_V2_ReturnParameters
+                    | hci.HCI_LE_CS_Read_Local_Supported_Capabilities_ReturnParameters
                 )
+                try:
```
zxzxwu: shouldn't we check command supported to decide which command to send?

bumble/device.py
```diff
@@ -6831,6 +6873,19 @@ def on_connection_data_length_change(
...
+    def on_cs_remote_supported_capabilities_v2(
+        self,
+        event: hci.HCI_LE_CS_Read_Remote_Supported_Capabilities_Complete_V2_Event,
+    ):
+        # 6.3 variant with T_IP2_IPT_Times and T_SW_IPT_Time appended.
+        # Fed through the same builder so downstream code doesn't care which
+        # HCI event landed.
+        self._on_cs_remote_supported_capabilities_impl(event)
+
+    def _on_cs_remote_supported_capabilities_impl(self, event) -> None:
```
zxzxwu: arg annotation

bumble/device.py
```diff
@@ -6840,6 +6895,13 @@ def on_cs_remote_supported_capabilities(
             )
             return
 
+        # V1 event carries rtt_random_sequence_n; V2 renamed the field to
+        # rtt_random_payload_n (same semantic per spec 6.3).
+        rtt_random = getattr(
```
zxzxwu: if isinstance(event, V2):
  ...
else:
  ...

bumble/hci.py
```diff
@@ -6097,7 +6167,10 @@ class Ch3cShape(SpecableEnum):
     channel_selection_type: int = field(metadata=ChannelSelectionType.type_metadata(1))
     ch3c_shape: int = field(metadata=Ch3cShape.type_metadata(1))
     ch3c_jump: int = field(metadata=metadata(1))
-    reserved: int = field(metadata=metadata(1))
+    # Bluetooth 6.3 CS Enhancements — last byte of the command was called
+    # "reserved" pre-6.3. Bit 0 (`CsEnhancements1.INLINE_PCT`) enables Inline
+    # PCT transfer on this CS configuration. Set to zero for 6.0 behavior.
+    cs_enhancements_1: int = field(default=0, metadata=metadata(1))
```
zxzxwu: should not have default value for HCI events

bumble/device.py
JoeBakalor: You're right, and my premise for the try/except was wrong. On the controller here (nRF54LM20A, NCS v3.4.0) octet 49 is `0x0c` — bit 2 set, `supports_command()` returns `True`. Gated on it in `9413603`; V2 is still selected, and forcing the check false falls back to the 6.0 read.

bumble/device.py
JoeBakalor: Done in `9413603`.

bumble/device.py
JoeBakalor: Done in `9413603` — moved the `isinstance` branch into `ChannelSoundingCapabilities.from_hci()` so `power_on()` shares it.

bumble/device.py
zxzxwu: Oh, I missed the explanation in PR description, but nice to catch that.

bumble/hci.py
JoeBakalor: Done in `729f9d5`.

bumble/device.py
```diff
@@ -6831,6 +6889,25 @@ def on_connection_data_length_change(
...
+    def on_cs_remote_supported_capabilities_v2(
+        self,
+        event: hci.HCI_LE_CS_Read_Remote_Supported_Capabilities_Complete_V2_Event,
+    ):
+        # 6.3 variant with T_IP2_IPT_Times and T_SW_IPT_Time appended.
+        # Fed through the same builder so downstream code doesn't care which
+        # HCI event landed.
+        self._on_cs_remote_supported_capabilities_impl(event)
+
+    def _on_cs_remote_supported_capabilities_impl(
```
zxzxwu: We can just keep a single on_cs_remote_supported_capabilities handler, and emit `'cs_remote_supported_capabilities'` from `Host.on_hci_le_cs_read_remote_supported_capabilities_complete_v2_event`

JoeBakalor: I'll update

zxzxwu (commented):
Thanks for adding Bluetooth 6.3 Inline PCT support and consolidating the capability parsers in `ChannelSoundingCapabilities.from_hci()`!

Comparing the implementation against the official [Bluetooth Core Specification v6.3 (HCI Functional Specification)](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core_v6.3/out/en/host-controller-interface/host-controller-interface-functional-specification.html), there are a few points to address:

1. **Missing 2M RTT Fields in V2 structures**:
   Per Core v6.3 § 7.8.130 (`LE CS Read Local Supported Capabilities command` [v2]) and § 7.7.65.39 (`LE CS Read Remote Supported Capabilities Complete event` [v2]), Bluetooth 6.3 appends 5 fields (7 octets total) after `tx_snr_capability`:
   - `t_ip2_ipt_times_supported`: 2 octets
   - `t_sw_ipt_times_supported`: 1 octet (named `T_SW_IPT_Times_Supported` in spec)
   - `rtt_2m_aa_only_n`: 1 octet
   - `rtt_2m_sounding_n`: 1 octet
   - `rtt_2m_random_sequence_n`: 1 octet

   Currently, `rtt_2m_aa_only_n`, `rtt_2m_sounding_n`, and `rtt_2m_random_sequence_n` are missing in both `HCI_LE_CS_Read_Local_Supported_Capabilities_V2_ReturnParameters` and `HCI_LE_CS_Read_Remote_Supported_Capabilities_Complete_V2_Event`. This results in a 32-octet payload instead of the spec-defined 35 octets (including `Status`).

2. **Field naming `rtt_random_sequence_n` vs `rtt_random_payload_n`**:
   In Core v6.3 § 7.8.130 & § 7.7.65.39, the parameter name is still `RTT_Random_Sequence_N` (and `RTT_2M_Random_Sequence_N` for 2M). Using `rtt_random_sequence_n` uniformly across both V1 and V2 structures matches the official specification and removes the need to branch on the field name in `ChannelSoundingCapabilities.from_hci()`.

3. **`cs_enhancements_1` naming**:
   In Core v6.3 § 7.8.137 (`LE CS Create Config command`), the parameter is named `CS_Enhancements` (1 octet, Bit 0: `IPT enabled in the CS reflector`). Naming the field `cs_enhancements` and the enum `CsEnhancements` (or `CsEnhancement`) would better align with the spec parameter name.

4. **Spec Section References in docstrings**:
   - `HCI_LE_CS_Read_Local_Supported_Capabilities_V2_Command`: The section in Core v6.3 is § 7.8.130 (OCF 0x00A5), rather than § 7.8.161.
   - `HCI_LE_CS_Read_Remote_Supported_Capabilities_Complete_V2_Event`: The section in Core v6.3 is § 7.7.65.39 (Subevent 0x38), rather than § 7.7.65.65.

JoeBakalor:
Thanks — all four addressed, and the branch is rebased onto `main`.

1. Added `rtt_2m_aa_only_n` / `rtt_2m_sounding_n` / `rtt_2m_random_sequence_n` to both [v2] structures and renamed `t_sw_ipt_time_supported` → `t_sw_ipt_times_supported`. The reply is 35 octets now. `9ec85fb`
2. `rtt_random_payload_n` → `rtt_random_sequence_n` on both [v2] shapes; `from_hci()` no longer branches on the field name. `3eb16de`
3. `cs_enhancements_1` → `cs_enhancements`, `CsEnhancements1` → `CsEnhancements`, bit 0 restated as "IPT enabled in the CS reflector". `4a925ea`
4. Sections corrected to 7.8.130 and 7.7.65.39, matching the [v1] classes. `bd1813e`

One thing worth flagging: Zephyr's structures stop after `t_sw_ipt_time_supported`, so a controller built on them advertises octet 49 bit 2 and then replies with 32 octets, which now fails to parse. Links and detail are in the PR description — happy to add either a tolerant tail parse or a fallback to the 6.0 read if you have a preference.

bumble/device.py
JoeBakalor: Done in `3eb16de` — single handler taking the `V1 | V2` union, and `host.py` emits `cs_remote_supported_capabilities` for both subevents.

JoeBakalor:
Followed up on the short-reply concern with a measurement, and it resolves in the spec's favour — worth recording here.

Captured the raw command-complete return parameters off an nRF54L15 (`hci_uart`, NCS v3.4.0) for both opcodes:

```
0x2089 [v1]: 29 octets  0001000001010101001e001e000000000212007c007e00e00103000000
0x20A5 [v2]: 35 octets  0001000001010101001e001e000000000212007c007e00e001030000007e000a1e001e
```

The [v2] reply is byte-identical to [v1] for its first 29 octets, then carries the full tail: `t_ip2_ipt_times_supported=0x7E`, `t_sw_ipt_times_supported=0x0A`, `rtt_2m_aa_only_n=30`, `rtt_2m_sounding_n=0`, `rtt_2m_random_sequence_n=30`. So the controller was sending all 35 octets all along and the old field list was silently dropping the last 6 — Bumble ignores trailing bytes, which is why the original 32-octet parse looked fine.

Zephyr's structs are still 32 octets, but that's host-side only; no controller I have replies short, so I've left the parse strict.

One correction to item 1: the tail is **6** octets, not 7 (2+1+1+1+1). Your 35-octets-including-Status figure is right — 29 + 6. I'd repeated the 7 in the PR description and have fixed it there.

zxzxwu (approved):

PR #958, opened 2026-07-10
Raise InvalidPacketError on a truncated ISO data packet

### LINKED ISSUE #955: HCI_IsoDataPacket.from_bytes(): struct.error on a short ISO packet instead of InvalidPacketError
`HCI_IsoDataPacket.from_bytes()` raises `struct.error` on a short ISO packet instead of `InvalidPacketError`.

## Reproduction

Deterministic, no hardware:

```python
from bumble import hci

# type=0x05 (ISO), pdu_info=0x0001 (handle=1, pb_flag=0b00), data_total_length=0
hci.HCI_IsoDataPacket.from_bytes(bytes([0x05, 0x01, 0x00, 0x00, 0x00]))
```

```
struct.error: unpack_from requires a buffer of at least 9 bytes for unpacking
4 bytes at offset 5 (actual buffer size is 5)
```

## Why

`from_bytes()` (`hci.py`, around lines 8189 and 8193) derives `pb_flag` from the header and then unpacks the timestamp / SDU-info fields without checking that the buffer actually contains them:

```python
if ts_flag:
    time_stamp, *_ = struct.unpack_from('<I', packet, pos)     # hci.py:8189
if should_include_sdu_info:
    packet_sequence_number, sdu_info = struct.unpack_from('<HH', packet, pos)  # hci.py:8193
```

With `pb_flag == 0b00` the SDU info is mandatory per the spec, so `data_total_length == 0` is malformed — but a malformed packet should not produce a `struct.error`.

## How it is reachable

`PacketParser` frames the packet purely from `data_total_length` (`HCI_PACKET_INFO[HCI_ISO_DATA_PACKET] = (2, 2, 'H')`), so a length of `0` yields a 5-byte packet that is happily emitted to the sink, which then calls `from_bytes()`:

```python
from bumble.transport import common
p = common.PacketParser()
p.set_packet_sink(sink)
p.feed_data(bytes([0x05, 0x01, 0x00, 0x00, 0x00]))   # sink.on_packet() called with 5 bytes
```

We hit this in the field on a serial (H:4) LE Audio controller. It is currently non-fatal only because `feed_data()` wraps `sink.on_packet()` in a bare `except Exception` and logs — so the real symptom is a logged traceback rather than a clean `InvalidPacketError`.

## Suggested fix

Validate the buffer length before unpacking and raise `core.InvalidPacketError`, consistent with `PacketReader` (`transport/common.py:195, 201`, which already raises `InvalidPacketError('packet too short')`).

Happy to send a PR.

TheArkadiuszGrzelka:
Fixes #955.

`HCI_IsoDataPacket.from_bytes()` called `struct.unpack_from()` without first checking the buffer length, so a short packet surfaced the low-level `struct.error` instead of the `InvalidPacketError` the sibling `from_bytes()` parsers raise. A 5-byte ISO packet (`data_total_length=0`, `pb_flag=0b00`) reproduces it: the SDU-info read at offset 5 needs 4 bytes that are not present.

This guards each `unpack_from` with a length check and raises `InvalidPacketError`, matching `HCI_AclDataPacket.from_bytes()`.

Test: `tests/hci_test.py::test_iso_data_packet_too_short` (fails with `struct.error` before the change, passes after). Full `hci_test.py` suite green.

### DIFF
--- a/bumble/hci.py
+++ b/bumble/hci.py
@@ -8174,6 +8174,10 @@ class HCI_IsoDataPacket(HCI_Packet):
         packet_status_flag: int | None = None
 
         pos = 1
+        if len(packet) < pos + 4:
+            raise InvalidPacketError(
+                f'ISO data packet too short: {len(packet)} bytes'
+            )
         pdu_info, data_total_length = struct.unpack_from('<HH', packet, pos)
         connection_handle = pdu_info & 0xFFF
         pb_flag = (pdu_info >> 12) & 0b11
@@ -8186,10 +8190,14 @@ class HCI_IsoDataPacket(HCI_Packet):
         if ts_flag:
             if not should_include_sdu_info:
                 logger.warning(f'Timestamp included when pb_flag={bin(pb_flag)}')
+            if len(packet) < pos + 4:
+                raise InvalidPacketError('ISO data packet truncated (timestamp)')
             time_stamp, *_ = struct.unpack_from('<I', packet, pos)
             pos += 4
 
         if should_include_sdu_info:
+            if len(packet) < pos + 4:
+                raise InvalidPacketError('ISO data packet truncated (SDU info)')
             packet_sequence_number, sdu_info = struct.unpack_from('<HH', packet, pos)
             iso_sdu_length = sdu_info & 0xFFF
             packet_status_flag = (sdu_info >> 15) & 1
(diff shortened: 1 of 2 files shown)


TheArkadiuszGrzelka:
@googlebot I signed it!

barbibulle (approved):

barbibulle:
Thanks for the PR. Could you run the formatter to update so we can merge it? (see the GitHub action output for details, or run `invoke project.pre-commit` for a local check)

TheArkadiuszGrzelka:
Done - black is happy now, and the hci_test.py suite is still green.

PR #959, opened 2026-07-10
Don't mask BIG (sync) establishment failure with a KeyError

### LINKED ISSUE #954: create_big_sync(): KeyError instead of HCI_Error hides the BIG sync failure status
`Device.create_big_sync()` raises `KeyError` instead of `HCI_Error`, discarding the BIG sync failure status.

## What happens

When BIG sync establishment fails, the `big_handle` entry is deleted twice, and the resulting `KeyError` replaces the real HCI status.

On `main` (checked at 8292-line `hci.py` / 7026-line `device.py`):

1. `on_big_sync_establishment()` (`device.py:5828-5832`) takes the `status != HCI_SUCCESS` branch:
   ```python
   if status != hci.HCI_SUCCESS:
       del self.big_syncs[big_handle]          # first delete
       ...
       big_sync.emit(BigSync.Event.ESTABLISHMENT_FAILURE, status)
   ```
2. That emit resolves the `established` future with `HCI_Error(status)`.
3. Back in `create_big_sync()` (`device.py:5279-5281`), `await established` raises, and:
   ```python
   except hci.HCI_Error:
       del self.big_syncs[big_handle]          # second delete -> KeyError
       raise
   ```

Because `KeyError` is not an `HCI_Error`, it escapes the handler and propagates to the caller. The `HCI_Error` carrying the status is lost.

## Impact

The caller sees `KeyError: 0` and has no way to learn why the sync failed. In our case the underlying status was `CONNECTION_TERMINATED_DUE_TO_MIC_FAILURE_ERROR (0x3D)` — a wrong broadcast code — but the exception said nothing about it. We lost a full measurement batch to this: twelve runs were scored as crashes while the actual cause was a changed broadcast code on the transmitting device.

Any BIG sync establishment failure is affected, not just MIC failures.

## Same defect in the broadcaster path

`create_big()` has the identical pattern: `device.py:5232` (`except HCI_Error: del self.bigs[big_handle]`) versus `device.py:5777` (`on_big_establishment`, failure branch). Worth fixing together.

## Suggested fix

Make the cleanup idempotent, e.g. `self.big_syncs.pop(big_handle, None)` in both places — `on_big_termination()` at `device.py:5801` already uses `pop(..., None)` for exactly this reason.

## Workaround for others hitting this

The status is still observable on the host event, before the exception is raised:

```python
failure = []
device.host.on('big_sync_establishment', lambda status, *_: status and failure.append(status))
try:
    big_sync = await device.create_big_sync(pa_sync, params)
except Exception:
    if failure:
        print(hci.HCI_Constant.error_name(failure[0]))   # the real reason
    raise
```

Happy to send a PR if the `pop(..., None)` approach looks right.

TheArkadiuszGrzelka:
Fixes #954.

On a BIG or BIG-sync establishment failure, the event handler (`on_big_establishment` / `on_big_sync_establishment`) removes the `bigs` / `big_syncs` entry and emits `ESTABLISHMENT_FAILURE`, which makes the awaited future raise `hci.HCI_Error`. `create_big()` / `create_big_sync()` then ran `del self.bigs[big_handle]` / `del self.big_syncs[big_handle]` in their `except hci.HCI_Error` clause on an entry that was already gone, raising `KeyError: 0` and discarding the real HCI status (e.g. a wrong broadcast code surfaces as `KeyError: 0` instead of `CONNECTION_TERMINATED_DUE_TO_MIC_FAILURE`).

This pops with a default instead of deleting, matching the idempotent pattern already used in `on_big_termination()`. The `HCI_Error` now propagates with its status intact.

Test: `tests/device_test.py::test_create_big_sync_failure_propagates_hci_error` reproduces the failure sequence (handler removes the entry, then the `except` clause runs); it raises `KeyError: 0` before the change and the correct `HCI_Error` after. Full `device_test.py` + `hci_test.py` suites green.

### DIFF
--- a/bumble/device.py
+++ b/bumble/device.py
@@ -5229,7 +5229,11 @@ class Device(utils.CompositeEventEmitter):
                 )
                 await established
             except hci.HCI_Error:
-                del self.bigs[big_handle]
+                # The establishment-failure event handler may have already
+                # removed the entry; pop so this cleanup cannot raise KeyError
+                # and mask the HCI_Error (see on_big_termination for the same
+                # idempotent pattern).
+                self.bigs.pop(big_handle, None)
                 raise
 
         return big
@@ -5278,7 +5282,11 @@ class Device(utils.CompositeEventEmitter):
                 )
                 await established
             except hci.HCI_Error:
-                del self.big_syncs[big_handle]
+                # The establishment-failure event handler may have already
+                # removed the entry; pop so this cleanup cannot raise KeyError
+                # and mask the HCI_Error (see on_big_termination for the same
+                # idempotent pattern).
+                self.big_syncs.pop(big_handle, None)
                 raise
 
         return big_sync
--- a/tests/device_test.py
+++ b/tests/device_test.py
@@ -30,6 +30,7 @@ from bumble.device import (
     Advertisement,
     AdvertisingEventProperties,
     AdvertisingParameters,
+    BigSyncParameters,
     CigParameters,
     CisLink,
     Connection,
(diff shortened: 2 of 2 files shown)


TheArkadiuszGrzelka:
@googlebot I signed it!

barbibulle (approved):

PR #963, opened 2026-07-17
USB HCI transport teardown — `usbi_mutex` abort on close

greateggsgreg:
Closing the USB HCI transport (a Nordic nRF52840 Zephyr dongle, `usb:2fe3:000b`) aborts 

`UsbPacketSink.terminate()` cancels its OUT transfer but reaps it via a shared semaphore
that a prior completion can pre-release, so a cancellation may go unreaped; and the transfer objects are
freed implicitly by python-libusb1's `USBDeviceHandle.close()` rather than explicitly after the join.
`libusb_free_transfer()` on a still-active transfer is illegal and destroys its mutex mid-use

## Reproduce

Run repeated BLE sessions over the nRF USB HCI transport and close each; the abort fires at teardown
(exit 134 + the `usbi_mutex` assertions), including on runs where the BLE work fully succeeded. It is
timing/load dependent in *severity* but reproduces readily under back-to-back cycles.

### DIFF
--- a/bumble/transport/usb.py
+++ b/bumble/transport/usb.py
@@ -665,8 +665,6 @@ class UsbTransport(Transport):
         self.device = device
         self.acl_interface = acl_interface
         self.sco_interface = sco_interface
-        self.loop = asyncio.get_running_loop()
-        self.event_loop_done = self.loop.create_future()
         self.event_loop_should_exit = False
         self.lock = threading.Lock()
         self.closed = False
@@ -719,7 +717,6 @@ class UsbTransport(Transport):
                 logger.warning(f'!!! Exception while handling events: {error}')
 
         logger.debug('ending USB event loop')
-        _safe_call_soon(self.loop, self.event_loop_done.set_result, None)
 
     async def close(self):
         # close() may be reached both explicitly and through Transport.__aexit__.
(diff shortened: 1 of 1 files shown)


bumble/transport/usb.py
```diff
@@ -734,17 +741,22 @@ async def close(self):
...
             logger.warning(f"Failed to interrupt event handler: {error}")
 
+        # handleEvents() owns libusb state until the event thread has returned.
+        # In particular, transfer cancellation callbacks only schedule their
+        # asyncio completion signals; those signals do not prove that the native
+        # callback has returned. Wait and join before freeing transfers, the
+        # device handle, or the context.
+        logger.debug("waiting for USB event loop to be done...")
+        await self.event_loop_done
+        self.event_thread.join()
```
barbibulle: This cannot be done here, because that's a blocking call.
Waiting for the thread to have finished with `join` should be done on a worker thread, with `asyncio.to_thread()`. And in that case, we can get rid of setting  `event_loop_done` prior to returning from `run`. When to_thread returns, the join is known to have completed.

barbibulle (changes requested):

bumble/transport/usb.py
greateggsgreg: Ah you're right - thoughts on current fix?

barbibulle: Looks good. I'll merge it.

barbibulle (approved):

PR #971, opened 2026-08-26
Migrate to flit

zxzxwu:


zxzxwu:
@barbibulle  Do we have any preference here? Recently I saw some [discussion](https://discuss.python.org/t/are-setuptools-abandoned/104390) about deprecation of setuptools. Though there is no official announcement about that, some [documents](https://packaging.python.org/en/latest/guides/tool-recommendations/) hint that setuptools is no longer the top choice, unless with any native extension.

Some direct advantages I saw with flit:
- Simpler pyproject.toml
- egg-info won't be generated

But we will also need to move apps and tools to under `bumble/`. We may also consider hatch or gdm

barbibulle:
I think it's a great idea to move to a newer packaging tool. I don't have any experience with flit or hatchling, but both seem like they would work nicely here.

zxzxwu:
Hatchling looks more flexible. If we want to keep the current layout, it would be the best choice.

zyv:
Did you consider `uv` ? It now has a build backend that is working nicely for me, and I'm a fan.

zxzxwu:
Do you mean uv_build backend? uv itself is the frontend of build system which supports most backends. It's also a choice but not so popular so I didn't list it here.

zyv:
> Do you mean uv_build backend? uv itself is the frontend of build system which supports most backends. It's also a choice but not so popular so I didn't list it here.

Yes, exactly.

PR #973, opened 2026-09-04
Keep the literal type of the AdvertisingData legacy constants

zyv:
Hi @zxzxwu,

I'm not sure how much you care about the legacy constants, but since they are still there, maybe the typing is still worth fixing. I didn't actually know about the new syntax, and that's why I ran into this issue when typechecking my project:

`AdvertisingData` forwards each `AdvertisingData.Type` member as a class-level constant for backward compatibility. Those constants are unannotated, so a type checker widens them to `AdvertisingData.Type`, no `Literal[Type...]` overload of `get()` or `get_all()` matches them, and the return type falls back to the AdvertisingDataObject union:

```python
ad.get(AdvertisingData.COMPLETE_LOCAL_NAME)       ->  the whole union
ad.get(AdvertisingData.Type.COMPLETE_LOCAL_NAME)  ->  str | None
```

The overloads arrived with the constant block in f321143, so they have never been reachable through the forwarded names. Every in-tree call through a forwarded name passes raw=True, which selects on the other parameter, so the mypy run in CI never reached the broken path.

My patch declares the backward-compatibility block `Final`, which keeps the literal type of each assignment and restores overload selection. pyright treats an all-caps class variable as implicitly `Final` and already reports `str | None` for both spellings; mypy, ty and zuban widen.

`invoke project.check-types` runs mypy over `tests/`, so the `assert_type` calls in the new test fail that gate if an overload stops resolving. mypy skips the body of an unannotated function, hence the `-> None`.

### DIFF
--- a/bumble/core.py
+++ b/bumble/core.py
@@ -25,6 +25,7 @@ from collections.abc import Iterable
 from typing import (
     Any,
     ClassVar,
+    Final,
     Literal,
     cast,
     overload,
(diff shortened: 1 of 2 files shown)


zxzxwu:
Interesting. My LSP (pyright, pyrefly) did consider them literals

image

But still this assertion doesn't pass without `Final`

image

zxzxwu (approved):

zyv:
> Interesting. My LSP (pyright, pyrefly) did consider them literals...but still this assertion doesn't pass without `Final`.

Yep, I think that the reason, as I mentioned, is that ALL_CAPS are implicitly considered `Final` in the LSP context, but when assertions are evaluated, it behaves like all the other type checkers - nice footgun under the guise of “user-friendliness," I would say, but that's just me.

PR #977, opened 2026-09-07
Feature - add <Coordinated Set Name> characteristic to csip.py

markusjellitsch:
Recently, the Bluetooth SIG adopted CSIS (Coordinated Set  Identification Service) v1.1, which introduces a
 new optional characteristic: Coordinated Set Name  (UUID 0x2C1A). This PR adds the <Coordindated Set Name> characteristic to `csip,py`. Moreover,  tests have been added in `test_csip.py `
 
 According to Spec.:

>  The Coordinated Set Name characteristic shall be the name of the Coordinated Set. The characteristic value is a UTF-8 string represented by a minimum of 0 octets and a maximum of 128 octets.

### DIFF
--- a/bumble/profiles/csip.py
+++ b/bumble/profiles/csip.py
@@ -27,6 +27,7 @@ from bumble import core, crypto, device, gatt, gatt_client
 # Constants
 # -----------------------------------------------------------------------------
 SET_IDENTITY_RESOLVING_KEY_LENGTH = 16
+COORDINATED_SET_NAME_MAX_LENGTH = 128
 
 
 class SirkType(enum.IntEnum):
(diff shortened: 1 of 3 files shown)


bumble/profiles/csip.py
```diff
@@ -200,6 +212,7 @@ class CoordinatedSetIdentificationProxy(gatt_client.ProfileServiceProxy):
     coordinated_set_size: gatt_client.CharacteristicProxy[bytes] | None = None
     set_member_lock: gatt_client.CharacteristicProxy[bytes] | None = None
     set_member_rank: gatt_client.CharacteristicProxy[bytes] | None = None
+    coordinated_set_name: gatt_client.CharacteristicProxy[bytes] | None = None
```
barbibulle: It would be even more convenient to use an adapter here, so that string decoding happens automatically. Use a `UTF8CharacteristicProxyAdapter`.
(see the `audio_input_description` charactertistic from `aics.py` for an example).
The adapter isn't really needed on the server side, since that value is static, so the conversion to bytes might as well be done inline in the constructor, like you have here, but on the client side, it's dynamic.

markusjellitsch: agree...this makes sense.
I have updated the client (and test) to now use the `UTF8CharacteristicProxyAdapter`.

barbibulle (approved):

